PR #12, opened 2020-06-19
Use interceptor for axios instance

rawagner:
Add possibility to pass custom axios instance

UHC portal can call `setClient(customClient)`. The `customClient` instance will have a proper interceptors set (adding auth/altering url). Standalone Facet app, does not have to do anything, just use the default client.

### DIFF
--- a/src/api/axiosClient.ts
+++ b/src/api/axiosClient.ts
@@ -1,6 +1,17 @@
-import axios from 'axios';
+import axios, { AxiosInstance } from 'axios';
 import applyCaseMiddleware from 'axios-case-converter';
 
-const client = applyCaseMiddleware(axios.create());
+const getDefaultClient = () => {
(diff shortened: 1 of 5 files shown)


src/api/axiosClient.ts
```diff
@@ -1,6 +1,17 @@
-import axios from 'axios';
+import axios, { AxiosInstance } from 'axios';
 import applyCaseMiddleware from 'axios-case-converter';
 
-const client = applyCaseMiddleware(axios.create());
+const getDefaultClient = () => {
+  const client = axios.create();
+  client.interceptors.request.use((cfg) => ({
+    ...cfg,
+    url: `${process.env.REACT_APP_API_ROOT}${cfg.url}`,
```
mareklibra, line 8: Can we take `API_ROOT` from `config.ts`?

PR #13, opened 2020-06-19
Wrap re-exports into namespaces

mareklibra:
Components can be imported directly. The rest of code is wrapped into namespaces.

In addition, potential circular dependencies cause build failure.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,8 +2,8 @@ import './hacks';
 
 export * from './components';
 
-export * from './api';
-export * from './config';
-export * from './selectors';
-export * from './store';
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -2,8 +2,8 @@ import './hacks';
 
 export * from './components';
 
-export * from './api';
-export * from './config';
-export * from './selectors';
-export * from './store';
-export * from './types';
+export * as Api from './api';
```
mareklibra, line 5: @jtomasek, @rawagner WDYT about this granularity?

Issue #6, opened 2020-06-15
Fix navigation - history is not available


mareklibra:
Fix use of `history` in the  `newClusterModal.tsx`.

PR #27, opened 2020-06-22
Fix new cluster modal redirect

jtomasek:
- fix axios client export
- fix getClusters url
- hide alerts section when empty
Screenshot 2020-06-22 at 20 23 15
Delete button is now clickable (not hidden behind empty AlertsSection

### DIFF
--- a/src/api/axiosClient.ts
+++ b/src/api/axiosClient.ts
@@ -10,8 +10,10 @@ const getDefaultClient = () => {
   return applyCaseMiddleware(client);
 };
 
-export let client: AxiosInstance = getDefaultClient();
+let client: AxiosInstance = getDefaultClient();
 
 export const setClient = (axiosInstance: AxiosInstance) => {
   client = applyCaseMiddleware(axiosInstance);
 };
+
+export { client };
--- a/src/components/clusters/Clusters.tsx
+++ b/src/components/clusters/Clusters.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
+import { RouteComponentProps } from 'react-router-dom';
 import {
   PageSectionVariants,
   TextContent,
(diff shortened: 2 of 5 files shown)


src/api/axiosClient.ts
```diff
@@ -10,8 +10,10 @@ const getDefaultClient = () => {
...
 };
 
-export let client: AxiosInstance = getDefaultClient();
+let client: AxiosInstance = getDefaultClient();
 
 export const setClient = (axiosInstance: AxiosInstance) => {
   client = applyCaseMiddleware(axiosInstance);
 };
+
+export { client };
```
mareklibra, line 19: @rawagner , please note this change

src/components/clusters/Clusters.tsx
```diff
@@ -124,7 +127,7 @@ const Clusters: React.FC = () => {
   return (
     <>
       {body}
-      {isModalOpen && <NewClusterModal closeModal={closeModal} />}
+      {isModalOpen && <NewClusterModal closeModal={closeModal} history={history} />}
```
mareklibra, line 130: @rawagner, do we have this in UHC?

mareklibra:
Fixes: #6

src/api/axiosClient.ts
rawagner: should be fine

src/components/clusters/newClusterModal.tsx
```diff
@@ -47,9 +47,10 @@ export const NewClusterModalButton: React.FC<NewClusterModalButtonProps> = ({
 
 type NewClusterModalProps = {
   closeModal: () => void;
+  history: History;
 };
 
-export const NewClusterModal: React.FC<NewClusterModalProps> = ({ closeModal }) => {
+export const NewClusterModal: React.FC<NewClusterModalProps> = ({ closeModal, history }) => {
```
rawagner, line 53: why dont we use withHistory HOC ?
Please also note that UHC uses older router (v4) which does not have hooks so we should avoid using them. Maybe we should downgrade the router version to make sure we wont use any new features of v5. v5 and v4 is backward compatibile so we shouldnt have any problems doing so.

src/components/clusters/Clusters.tsx
rawagner: commented here https://github.com/mareklibra/facet-lib/pull/27/files#r444028129

src/components/clusters/newClusterModal.tsx
jtomasek: I didn't see it necessary to use the HOC when the props passing is just one level.

jtomasek: We can downgrade the router in a separate PR

PR #18, opened 2020-06-22
Downgrade PF

rawagner:
add basepath for all routes
create lib router component

PR #28, opened 2020-06-22
Fix the cluster table header styling

andybraren:
Given that I didn't test this locally and don't know React, I would be shocked if this is correct. Feel free to close and fix elsewhere. 🙂

This PR tries to remove two button-related classes on `.pf-c-table__button-content` that aren't needed, fixing the styling to go from this:

image

To this:

image

### DIFF
--- a/src/components/ui/table/sortable.tsx
+++ b/src/components/ui/table/sortable.tsx
@@ -45,7 +45,6 @@ const sortable: ITransform = (label, extra: IExtra | undefined) => {
         isSortedBy={isSortedBy}
         sortDirection={isSortedBy ? sortBy?.direction : ''}
         onSort={sortClicked}
-        className="pf-c-button pf-m-plain"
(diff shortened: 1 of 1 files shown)


jtomasek:
This is a leftover from old patternfly. With the new patternfly we can switch back to using patternfly's `sortable` which fixes this issue as well.

However it seems that we'll need to switch back to old PF version. So I'd suggest changing this PR to use patternfly's `sortable` but keep our file around so it can be used again as part of https://github.com/mareklibra/facet-lib/pull/18

mareklibra:
> However it seems that we'll need to switch back to old PF version. So I'd suggest changing this PR to use patternfly's sortable but keep our file around so it can be used again as part of #18

Just to double-check my understanding: we will fix for new PF now but will "immediately" revert in#18. So we can piggy-back on #18 in the future when upgrading again.

PR #29, opened 2020-06-23
Remove footer alert padding

andybraren:
Attempts to resolve [MGMT-1234](https://issues.redhat.com/browse/MGMT-1234) by removing the top padding (which doesn't seem to do anything?) defined in the custom CSS for `.pf-c-alert-group.alerts-section`. This allows the `ssh-keygen` command to be selected.

Feel free to close if this isn't the right approach. It seems like the entire div and this CSS could be removed with no effect on the alerts that appear in the footer.

Removes the green region here:

image


jtomasek:
This is fixed as part of https://github.com/mareklibra/facet-lib/pull/27 by not rendering the alerts section at all when alerts are empty

mareklibra:
Closing per ^^^ comment

PR #30, opened 2020-06-23
Relabels Created At column heading to Discovered At

andybraren:
This PR attempts to rename the `Created At` column heading to `Discovered At`. The word "created" may match what's happening in the backend, but from the user's perspective these hosts are "Discovered" at a certain time, and "Discovering" is [the first status](https://docs.google.com/document/d/1jxNMTlotmJ0GFZ1GUEQ3hfOOVLzMT6mvY8Ufo5RdErY/edit#heading=h.gptlbp92g8lj) a user sees when the host appears in the table. WDYT @romfreiman @avishayt?

I see some references to `createdAt` in the code but don't want to break anything by renaming those too. 🙂 Feel free to close this PR and tweak it elsewhere if we generally agree with this proposed change.

image


avishayt:
Makes sense to me

PR #41, opened 2020-06-23
MGMT-1146 Enforce exactly 3 master hosts

mareklibra:
In addition, the validation button changes it's text from
- `Validate` if there are no changes (form is not dirty)
- `Validate & Save Changes` if there are unsaved changes.

The button is newly disabled if only a submit procedure is in progress or fronted validations failed.
Otherwise it is enable (i.e. even if there are no changes in the form).

In other words, the user can `Validate` even an untouched form or once a role is changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "facet-lib",
-  "version": "1.0.11",
+  "version": "1.0.12",
   "description": "React component library for the Bare Metal Installer",
   "author": "mareklibra",
   "license": "Apache-2.0",


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -267,10 +267,10 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
                 type="submit"
                 name="save"
                 variant={ButtonVariant.secondary}
-                isDisabled={isSubmitting || !isValid || !dirty}
+                isDisabled={isSubmitting || !isValid}
                 onClick={submitForm}
               >
-                Validate & Save Changes
+                Validate {dirty ? '& Save Changes' : ''}
```
jtomasek, line 273: IIUC there should be no case when user needs to revalidate without also having any changes. The cluster either gets updated from the server which automatically updates the state and validations or user has done some changes to the form and in that case `Validate & Save Changes` button is enabled.

I quite like how enabling that button lets user know there are unsaved changes (note that this is in addition reflected in the global errors in the toolbar)

This change also makes the button label dynamic which will be confusing for the user ("Sometimes I see 'Validate' sometimes I see 'Validate & Save', what does it mean?")

@andybraren your opinion?

mareklibra: Sounds like consensus, so I removed the commit.

PR #43, opened 2020-06-24
MGMT-1109 Add filters to the ClustersPage

mareklibra:


mareklibra:
Default, after page load, no filter is selected but everything is shown:


---

Filter by cluster name:


---

If no status-filter is selected then everything matches. Otherwise per selection:

mareklibra:
Cc @andybraren

andybraren:
Sweet! 👍

Some small suggestions:

image

1. Add a filter icon to the Status dropdown button (aligns with OCP, OCM, and [PF](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#filter-chips))
2. Add placeholder text to the input field, `Filter by name...` (aligns with OCM and [PF guidance](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#filters), OCP will likely move to this wording soon)
3. Switch the order so that the Status filter is first and the search field is second (aligns with OCP and [PF's examples](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#filter-chips), and will look a bit better so the two icons aren't next to each other after the filter one is added)

mareklibra:
@andybraren , flipped:

andybraren:
Perfect, thank you!

src/components/clusters/ClustersFilterToolbar.tsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react';
+import {
+  Toolbar,
+  ToolbarItem,
+  ToolbarContent,
+  ToolbarFilter,
+  ToolbarToggleGroup,
+  ToolbarGroup,
```
jtomasek, line 8: Unused imports

src/components/clusters/ClustersFilterToolbar.tsx
```diff
@@ -0,0 +1,141 @@
...
+          <Select
+            variant="checkbox"
+            aria-label="status"
+            onToggle={onStatusToggle}
+            onSelect={onStatusSelect}
+            selections={filters.status}
+            isOpen={isStatusExpanded}
+            placeholderText={statusPlaceholder}
+          >
+            {Object.getOwnPropertyNames(CLUSTER_STATUS_LABELS).map((status) => (
```
jtomasek, line 114:
```suggestion
            {Object.keys(CLUSTER_STATUS_LABELS).map((status) => (
```
seems to work as well

src/components/clusters/ClustersFilterToolbar.tsx
mareklibra: removed

src/components/clusters/ClustersFilterToolbar.tsx
mareklibra: there is no difference in our case, but changed

PR #50, opened 2020-06-26
Add ISO link to ISO download dialog

jtomasek:
This changes the discovery modal behavior - after the download is ready
the modal does not close and provides an ISO URL to be coppied and a button
to trigger a download via browser.

iso-download-final


andybraren:
Nice! If the user sets a proxy URL, would that be useful to be able to see in the "results" screen too? FYI @nirfarkas

romfreiman:
Boom!!!!

jtomasek:
Added proxy URL in the ISO summary if provided.
Screenshot 2020-06-29 at 10 53 27

PR #51, opened 2020-06-29
Move DNS base domain filed to Networking section

mareklibra:


src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -11,6 +11,12 @@ const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets })
   const { validateField } = useFormikContext<ClusterConfigurationValues>();
   return (
     <>
+      <InputField
+        label="Base DNS Domain"
+        name="baseDnsDomain"
+        helperText="The base domain of the cluster. All DNS records must be sub-domains of this base and include the cluster name."
```
ohadlevy, line 17: didnt want to add an example? e.g. ```basedomain.example.com``` ?

mareklibra: good idea since we provide examples for other fields

mareklibra: https://github.com/mareklibra/facet-lib/pull/53

PR #53, opened 2020-06-29
Add example of base DNS domain

mareklibra:


mareklibra:
I find placeholder useful, but so far they are not used in this form. @andybraren, wdyt ?

andybraren:
I think placeholder text would make sense here. And to keep the scope creep train going 😎 could we get even fancier and dynamically show the full address?  (Someone proposed this in Slack, and it seems to make a lot of sense):

base-dns-service-iteration-1

mareklibra:
@andybraren , done

>  And to keep the scope creep train going ...

The most valuable thing is a bright idea.

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -43,7 +60,7 @@ const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets })
       <InputField
         label="API Virtual IP"
         name="apiVip"
-        helperText="Virtual IP used to reach the OpenShift cluster API. Make sure that the VIP's are unique and not used by any other device on your network."
+        helperText={`Virtual IP used to reach the OpenShift cluster API. Make sure that the VIP's are unique and not used by any other device on your network${subnet}.`}
```
jtomasek, line 63: miss
```suggestion
        helperText={`Virtual IP used to reach the OpenShift cluster API. Make sure that the VIP's are unique and not used by any other device on your network ${subnet}.`}
```
missing space ^

mareklibra: No, that was intentional :-) Otherwise the `dot` would be after a space, see how `subnet` is defined above.

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -1,20 +1,37 @@
...
 
 type BasicNetworkFieldsProps = {
   hostSubnets: HostSubnets;
 };
 
 const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets }) => {
   const { validateField } = useFormikContext<ClusterConfigurationValues>();
+  const [clusterName] = useField({ name: 'name' });
+  const [baseDnsField] = useField({ name: 'baseDnsDomain' });
+  const [hostSubnet] = useField({ name: 'hostSubnet' });
```
jtomasek, line 14: useField is used to construct custom input fields. the values are available through useFormikContext

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -1,20 +1,37 @@
...
 import { InputField, SelectField } from '../ui/formik';
-import { useFormikContext } from 'formik';
+import { useFormikContext, useField } from 'formik';
 
 type BasicNetworkFieldsProps = {
   hostSubnets: HostSubnets;
 };
 
 const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets }) => {
   const { validateField } = useFormikContext<ClusterConfigurationValues>();
```
jtomasek, line 11:
```suggestion
  const { validateField, values } = useFormikContext<ClusterConfigurationValues>();
  const { clusterName, baseDnsDomain, hostSubnet } = values;
```

src/components/clusterConfiguration/BasicNetworkFields.tsx
jtomasek: Ah, I see, thanks :)

src/components/clusterConfiguration/BasicNetworkFields.tsx
mareklibra: done

PR #57, opened 2020-06-30
Add Red Hat's DNS service capability

jtomasek:
assisted-dns

### DIFF
--- a/src/components/clusterConfiguration/ClusterConfiguration.tsx
+++ b/src/components/clusterConfiguration/ClusterConfiguration.tsx
@@ -71,8 +71,8 @@ const validationSchema = (hostSubnets: HostSubnets) =>
 
 const sshPublicKeyHelperText = (
   <>
-    SSH public key for debugging OpenShift nodes, value of <em>~/.ssh/id_rsa.pub</em> can be
-    copy&amp;pasted here. To generate new pair, use <em>ssh-keygen -o</em>.
+    SSH public key for debugging OpenShift nodes, value of <em>~/.ssh/id_rsa.pub</em> can be copy
+    &amp; pasted here. To generate new pair, use <em>ssh-keygen -o</em>.
   </>
 );
 
@@ -104,12 +104,14 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster }) 
 
     // update the cluster configuration
     try {
-      let params = _.omit(values, ['hostSubnet', 'isPullSecretEdit']);
+      let params = _.omit(values, ['hostSubnet', 'isPullSecretEdit', 'useRedHatDnsService']);
       if (!values.isPullSecretEdit) {
         params = _.omit(params, ['pullSecret']);
       }
       const { data } = await patchCluster(cluster.id, params);
-      formikActions.resetForm({ values: getInitialValues(data) });
+      formikActions.resetForm({
+        values: getInitialValues(data),
+      });
       dispatch(updateCluster(data));
     } catch (e) {
       handleApiError<ClusterUpdateParams>(e, () =>
@@ -191,7 +193,9 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster }) 
                     </GridGap>
                   </GridItem>
                   <GridItem span={12}>
-                    <BaremetalInventory cluster={cluster} />
+                    <GridGap>
+                      <BaremetalInventory cluster={cluster} />
+                    </GridGap>
                   </GridItem>
                   <GridItem span={12} lg={10} xl={6}>
                     <GridGap>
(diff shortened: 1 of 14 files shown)


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -1,332 +1,38 @@
...
+  const [domains, setDomains] = React.useState<ManagedDomain[] | undefined>();
+
+  React.useEffect(() => {
+    const fetchManagedDomains = async () => {
+      try {
+        const { data } = await getManagedDomains();
+        setDomains(data);
+      } catch (e) {
+        setDomains([]);
+        handleApiError(e);
```
rawagner, line 23: Shouldnt we also notify user that we could not fetch the domains via alert `dispatchAlertsAction` ?

jtomasek: The list of domains is not something that user can influence, so I opted to fail silently (by logging the error in the console) and fall back to empty list of domains. If the list of domains is empty user won't see the option to select to use the RH provided domains.

src/components/clusterConfiguration/utils.ts
```diff
@@ -63,4 +66,7 @@ export const getInitialValues = (cluster: Cluster): ClusterConfigurationValues =
   sshPublicKey: cluster.sshPublicKey || '',
   isPullSecretEdit: !cluster.pullSecretSet, // toggles edit mode and drives validation
   hostSubnet: findMatchingSubnet(cluster.ingressVip, cluster.apiVip, getHostSubnets(cluster)),
+  useRedHatDnsService:
+    !!cluster.baseDnsDomain &&
+    managedDomains.map((d: ManagedDomain) => d.domain).includes(cluster.baseDnsDomain),
```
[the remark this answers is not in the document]
jtomasek: Done

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -22,11 +29,58 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({ hostSubnets
...
+          helperText="A base DNS domain will be provided by Red Hat's DNS service. Because the cluster's DNS can't be changed later, this should only be used for temporary, non-production clusters."
+          onChange={toggleRedHatDnsService}
+        />
+      )}
+      {values.useRedHatDnsService ? (
+        <SelectField
+          label="Base DNS Domain"
+          name="baseDnsDomain"
+          helperText={baseDnsHelperText}
+          options={managedDomains.map((d: ManagedDomain) => ({
```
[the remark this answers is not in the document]
jtomasek: Done

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -22,11 +29,58 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({ hostSubnets
...
+    <>
+      The base domain of the cluster. All DNS records must be sub-domains of this base and include
+      the cluster name. This cannot be changed later. The full cluster address will be:{' '}
+      <strong>
+        {clusterName || '[Cluster Name]'}.{baseDnsDomain || '[example.com]'}
+      </strong>
+    </>
+  );
+
+  const toggleRedHatDnsService = (checked: boolean) => {
```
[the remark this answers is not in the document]
jtomasek: Done

src/components/clusterConfiguration/utils.ts
```diff
@@ -63,4 +66,6 @@ export const getInitialValues = (cluster: Cluster): ClusterConfigurationValues =
   sshPublicKey: cluster.sshPublicKey || '',
   isPullSecretEdit: !cluster.pullSecretSet, // toggles edit mode and drives validation
   hostSubnet: findMatchingSubnet(cluster.ingressVip, cluster.apiVip, getHostSubnets(cluster)),
+  useRedHatDnsService:
+    !!cluster.baseDnsDomain && managedDomains.map((d) => d.domain).includes(cluster.baseDnsDomain),
```
rawagner, line 70: What happens if I 
1. create a cluster
2. set my custom dns domain
3. save and close the form
4. open the form again
 ? 
Seems like `useRedHatDnsService` will be true, which is not correct ?

src/components/clusterConfiguration/ClusterConfiguration.tsx
rawagner: Its true that user cannot influence it, but he still may be left wondering what happened to managed domains ? I still prefer transparency over failing silently. I think any request that fails and shouldn't  (meaning its an actual error) should be visualized to user. But thats just my preference, if you think the current solution is better then lets go with it.

src/components/clusterConfiguration/utils.ts
rawagner: ok, somehow I missed the includes :)

src/components/clusterConfiguration/ClusterConfiguration.tsx
jtomasek: OK, I've enhanced the error handling to include errors for fetching the domains.

PR #58, opened 2020-07-01
Add checkbox to enable proxy configuration on discovery ISO dialog

jtomasek:
assisted-proxy-checkbox


andybraren:
Nice, I like “Configure” better than “Use,” lgtm!

PR #63, opened 2020-07-08
Alerts context

jtomasek:
Add AlertsContext
    
    - Convert hosts table actions to use alerts in AlertsSection
    - Provide alerts list and actions via context
    - Update host actions to dispatch updateHost action instead of
      forcing the cluster refresh since the API now correctly returns
      a resource

Depends on https://github.com/mareklibra/facet-lib/pull/47/files


src/features/clusters/currentClusterSlice.ts
```diff
@@ -33,6 +34,13 @@ export const currentClusterSlice = createSlice({
   name: 'currentCluster',
   reducers: {
     updateCluster: (state, action: PayloadAction<Cluster>) => ({ ...state, data: action.payload }),
+    updateHost: (state, action: PayloadAction<Host>) => {
+      const hostIndex = _.findIndex(state.data?.hosts, (host) => host.id === action.payload.id);
+      if (hostIndex >= 0) {
+        _.set(state, `data.hosts[${hostIndex}]`, action.payload);
```
rawagner, line 40: you should not modify state directly

src/components/ui/AlertsSection/index.tsx
```diff
@@ -1,27 +1,33 @@
...
-        ))}
-      </AlertGroup>
-    </PageSection>
-  ) : null;
+  if (alerts.length) {
+    return (
+      <PageSection padding={{ default: 'noPadding' }}>
+        <AlertGroup className="alerts-section">
+          {alerts.map((alert) => (
+            // eslint-disable-next-line react/jsx-key
```
rawagner, line 18: use `alert.key` field as key ?

src/features/clusters/currentClusterSlice.ts
jtomasek: redux toolkit uses Immer behind the scenes which allows using direct mutations in reducers. https://redux-toolkit.js.org/api/createReducer#direct-state-mutation

I agree that it is not ideal to overuse this type of magic but in this case it makes the code quite simpler.

src/components/ui/AlertsSection/index.tsx
jtomasek: `{...alert}` below adds the key correctly
image

PR #64, opened 2020-07-09
MGMT-1736 Add CreatedAt column to clusters list

jtomasek:
~~Depends on https://github.com/mareklibra/facet-lib/pull/63~~

### DIFF
--- a/src/components/clusters/ClustersTable.tsx
+++ b/src/components/clusters/ClustersTable.tsx
@@ -44,6 +44,7 @@ const columns = [
   { title: 'Version', ...columnConfig },
   { title: 'Status', ...columnConfig },
   { title: 'Hosts', ...columnConfig },
+  { title: 'Created at', ...columnConfig },
 ];
 
 const getStatusCell = (row: IRow) => row.cells?.[3] as HumanizedSortable | undefined;
--- a/src/selectors/clusters.tsx
+++ b/src/selectors/clusters.tsx
@@ -7,7 +7,7 @@ import { ResourceUIState } from '../types';
 import { Link } from 'react-router-dom';
 import { IRow, IRowData } from '@patternfly/react-table';
 import { RootState } from '../store/rootReducer';
-import { HumanizedSortable } from '../components/ui/table/utils';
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -100,6 +99,8 @@
     "library"
   ],
   "dependencies": {
+    "@types/human-date": "^1.4.0",
```
rawagner, line 102: this should be `devDependency`

src/components/clusters/ClustersTable.tsx
```diff
@@ -40,6 +40,8 @@ const columns = [
   { title: 'Version', ...columnConfig },
   { title: 'Status', ...columnConfig },
   { title: 'Hosts', ...columnConfig },
+  { title: 'Status Updated At', ...columnConfig },
+  { title: 'Created At', ...columnConfig },
```
andybraren, line 44: If the date format we use is "2 days ago" and eventually a date, `Created` without the `at` may be better here, which is what OCM uses. Their `Created` is a bit different than ours, but I'm struggling to find out a better word.

Maybe our `database_created_date` could be shown as the `Created` date while the cluster is a Draft, and after it succeeds the `installation_finished_time` could be used as the `Created` date instead? WDYT?

package.json
jtomasek: done

src/selectors/clusters.tsx
```diff
@@ -48,6 +48,7 @@ const clusterToClusterTableRow = (cluster: Cluster): IRow => {
         title: hostsCount.toString(),
         sortableValue: hostsCount,
       } as HumanizedSortable,
+      getRelativeTimeCell(createdAt),
```
mareklibra, line 51: I think the user should be able to see even non-relative time. In a popover? @andybraren 

So far all attempts to show relative time throughout the application were reverted back to rendering real time. How is this different?

jtomasek:
Reverted back to not using relative time
Screenshot 2020-09-30 at 10 26 39

mareklibra (approved):

mareklibra (approved):

PR #69, opened 2020-07-13
Progress api updates

jtomasek:
- Adjust progress info utils to API changes
- Accommodate 'installing-pending-user-action' host status
- Calculate progress based on progress info provided from backend

### DIFF
--- a/src/components/hosts/HostStatus.tsx
+++ b/src/components/hosts/HostStatus.tsx
@@ -16,12 +16,13 @@ import {
   UnknownIcon,
 } from '@patternfly/react-icons';
 import { Host } from '../../api/types';
-import HostProgress, { getHostInstallationStepNumber } from './HostProgress';
+import HostProgress from './HostProgress';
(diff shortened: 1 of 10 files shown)


src/components/hosts/HostStatus.tsx
```diff
@@ -31,82 +32,55 @@ const getStatusIcon = (status: Host['status']) => {
...
-    progressInfo.steps,
-    progressInfo.currentStep,
-  );
-  const bodyContent = React.useMemo(
-    () => (
-      <div>
-        {['installing', 'installing-in-progress'].includes(status) ? (
-          <>
-            <HostProgress status={status} progressInfo={progressInfo} />
+  const bodyContent = React.useMemo(() => {
```
rawagner, line 53: I dont think there's any performance boost from using `useMemo` sice we have `host` in dependency array. And `host` is the only param that can trigger a rerender.

jtomasek: I've extracted that into a function

PR #73, opened 2020-07-15
Downgrade react-router to uhc-portal version.

rawagner:
Use react-redux as dependency to not depend on uhc-portal version


rawagner:
uhc portal has a pending PR to update react-redux and react-router

rawagner:
closing, uhc-portal upgraded their deps

PR #78, opened 2020-07-15
Update "close" to "Back to all clusters"

nirfarkas:


nirfarkas:
I (think) I've updated also #131 (:

nirfarkas:
Yay!

PR #84, opened 2020-07-17
Fix lint tests

jtomasek:


package.json
```diff
@@ -18,11 +18,11 @@
     "sync-to-facet": "FACET_ROOT=../facet ./scripts/sync-dist.sh",
     "prepare": "install-peers",
     "test": "run-s test:circular test:lint",
-    "test:lint": "eslint .",
+    "test:lint": "eslint --ext .js,.jsx,.ts,.tsx --color ./src",
```
mareklibra, line 21: I think we can remove this target now and update the `test` target accordingly.

PR #86, opened 2020-07-20
Add severity labels to EventsList

jtomasek:
- Update API types
- include new 'preparing-for-installation' host status

Screenshot from 2020-07-20 13-07-06

### DIFF
--- a/src/api/swagger.yaml
+++ b/src/api/swagger.yaml
@@ -941,6 +941,7 @@ definitions:
     type: object
     required:
       - entity_id
+      - severity
       - message
       - event_time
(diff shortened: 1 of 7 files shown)


jtomasek:
@andybraren maybe not using icon would be cleaner?

mareklibra:
What do you think about accompanying similar changes with immediate updates of the e2e tests within `facet` project?

jtomasek:
> What do you think about accompanying similar changes with immediate updates of the e2e tests within `facet` project?

+1 yes, should have done it!

PR #88, opened 2020-07-21
Installation vs cluster ssh keys

mareklibra:


mareklibra:
---

mareklibra:
Implementation slightly deviates from the design - the `switch` is part of the form-item, means it is not added as an extra field.

Behaviour:
- If the image key is not set, the switch does not appear and the cluster-ssh text area works as before.
- Otherwise (if it appears):
  - The switch is `off` when the two keys are different (means the user already provided different cluster key).
  - The switch is `on` if the two keys are equal or by default.

The `textarea` always shows the "effective" ssh cluster key (means even a copy of the installation key). If the switch is `on`, the text area is disabled for editation.

mareklibra:
Cc @andybraren

src/components/clusterConfiguration/DiscoveryImageForm.tsx
```diff
@@ -133,7 +133,7 @@ const DiscoveryImageForm: React.FC<DiscoveryImageFormProps> = ({
                 <TextAreaField
                   label="SSH public key"
```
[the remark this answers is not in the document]
jtomasek: I think 'installation' is misleading as that term is used for cluster installation rather than host registration across the application. 'Installation SSH Key' I'd expect to be a key provided for the cluster installation rather than host registration. How about 'Discovery Image SSH Public Key'?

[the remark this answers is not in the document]
jtomasek: Alternatively we could keep this one intact 'SSH public key' and in the switch refer to this as 'Host discovery SSH key'

mareklibra:
@andybraren ,

> IIRC the form currently says something like "this key is being hidden for security reasons" which might be true

This is probably relevant to the pull-secret. Since these SSH keys are public, there is probably no issue in revealing them.

Otherwise I got your point in hiding the textarea and moving the switch up. Will change this PR to that state.
I would postpone the kind of "file upload" for a later PR. So far the user can "copy to clipboard" the secret via a button right on the webpage and (s)he is instructed to do so. But I agree it makes sense in other cases.

ohadlevy:
perhaps troubleshooting / debug key? (vs install)

src/components/clusterConfiguration/DiscoveryImageForm.tsx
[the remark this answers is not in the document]
mareklibra: The `Discovery Image SSH Public Key` is clear from user perspective since it points where it is entered.

Anyway I don't think it is clear what are these two keys used for and especially why the user might want to make them different.

What about `Host SSH Public Key while instlling` and `Host SSH Public Key for troubleshooting after installation`?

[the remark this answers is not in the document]
ohadlevy: > The `Discovery Image SSH Public Key` is clear from user perspective since it points where it is entered.
> 

I would challenge that assumption, you are assuming the user understand what is about to happen...


perhaps we can have the setting of reusing the same key where the user first set it (e.g. in the iso creation phase)? and then have a checkbox for using different ssh keys for the cluster and troubleshooting installation.

IMHO 99/100 folks would always use the same ssh key.

src/components/clusterConfiguration/utils.ts
```diff
@@ -64,6 +64,7 @@ export const getInitialValues = (
   ingressVip: cluster.ingressVip || '',
   pullSecret: '',
   sshPublicKey: cluster.sshPublicKey || '',
+  imageSshPublicKey: cluster.imageInfo.sshPublicKey || '',
```
jtomasek, line 67: I think this should not be set in initialValues as initial values are set only when the form is mounted. Instead we should pass the cluster.imageInfo.sshPublicKey as prop to the `SshClusterKey` component

src/components/clusterConfiguration/ClusterSshKey.tsx
```diff
@@ -0,0 +1,61 @@
...
+    Provide a SSH public key to debug OpenShift nodes after installation. To generate a new key, use
+    the <em>ssh-keygen -o</em> command and paste the value of <em>~/.ssh/id_rsa.pub</em> here.
+  </>
+);
+
+const ClusterSshKey: React.FC<{}> = () => {
+  const { setFieldValue, initialValues } = useFormikContext<ClusterConfigurationValues>();
+  const isChecboxVisible = !!initialValues.imageSshPublicKey;
+  const [isChecked, setChecked] = React.useState(
+    (isChecboxVisible && initialValues.sshPublicKey === initialValues.imageSshPublicKey) ||
```
jtomasek, line 18:
```suggestion
    (isChecboxVisible && initialValues.sshPublicKey === props.imageSshPublicKey) ||
```

src/components/clusterConfiguration/ClusterSshKey.tsx
```diff
@@ -0,0 +1,61 @@
...
+      !initialValues.sshPublicKey,
+  );
+
+  const onChange = (newVal: boolean) => {
+    setChecked(newVal);
+    if (newVal) {
+      setFieldValue('sshPublicKey', initialValues.imageSshPublicKey);
```
jtomasek, line 25:
```suggestion
      setFieldValue('sshPublicKey', props.imageSshPublicKey);
```

src/components/clusterConfiguration/ClusterSshKey.tsx
mareklibra: done

mareklibra:
@andybraren , the `textarea` is hiding now.

Best-effort rewording:
- in discovery modal: `Host SSH public key for troubleshooting during discovery`
- in the cluster form: `Host SSH Public Key for troubleshooting after installation`

---



---

andybraren:
After re-reading all of the [Jira comments](https://issues.redhat.com/browse/MGMT-1051) and discussions in this PR, I think it really comes down to what @ohadlevy wrote:

> IMHO 99/100 folks would always use the same ssh key.

I know @romfreiman thinks the same.

If that's true, then separate keys is an "Advanced" use case that we probably don't need to support for now. Maybe someday a user will have a strong use case and we can support it in an "Advanced" area/step of a future UI, but until then having two fields doesn't seem worth the UX impact. Wording changes only get us so far, and require more reading for everyone in the end.

What do we think about just keeping the `SSH Public Key` in the Discovery ISO modal and removing the second field entirely for now?

FYI @mgoldboi @jtomasek @nirfarkas

src/components/clusterConfiguration/ClusterSshKey.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (newVal) {
+      setFieldValue('sshPublicKey', imageSshPublicKey);
+    } else {
+      setFieldValue('sshPublicKey', initialValues.sshPublicKey);
+    }
+  };
+
+  React.useEffect(
+    () => {
+      if (isChecked || !values.sshPublicKey) {
```
jtomasek, line 44: I think this condition is wrong. The effect should happen only if the imageSshPublicKey is set and values.sshPublicKey is not.

jtomasek: Currently the value.sshPublicKey is updated even if sshPublicKey is not set. Which makes the user prompt to save the form even when there are no changes to it because the effect sets the values.sshPublicKey to `''`.

jtomasek: If the user re-configures the discovery image with a new key, what do we do? Should we switch the switch off and keep the old ssh key value (which is what is currently set in cluster) or should we keep the switch on and prompt the user to 'save and validate' to update the value in cluster configuration as well?

mareklibra: I think the second option is right - the value should be updated and user should save it. I PR changed to this behaviour.

mareklibra:
Not sure about behaviour for "erasing cluster ssh key". If the user wants the key for discovery image but does not want it further, (s)he can "switch off" syncing, remove content and save. Next-time (s)he enters the page, the switch is back to "on" (since the cluster.sshKey is empty string) and the page is marked as dirty. 
So the user needs to switch syncing off again manually. But probably without additional info stored/provided by backend we can not handle this corner case right.

src/components/clusterConfiguration/ClusterSshKey.tsx
```diff
@@ -0,0 +1,83 @@
...
+      helperText={<HelperText fieldId={fieldId}>{sshPublicKeyHelperText}</HelperText>}
+    >
+      {checkbox}
+    </FormGroup>
+  ) : (
+    <TextAreaField
+      name="sshPublicKey"
+      label={sshPublicKeyLabel}
+      helperText={sshPublicKeyHelperText}
+      className="ssh-public-key__textarea"
```
jtomasek, line 76: Seems like the custom css is no longer needed

mareklibra: This is still needed otherwise spacing between switch and text-area does not work

mareklibra:
I rewrote the component into a Formik's field, so Formik handles state of the Switch and so the Discard button works properly now.

There is still an issue with the "removing the cluster ssh key" case. It can be done but after page refresh the switch is automatically back on "on" and so the form is marked as `dirty` (to be saved), since this is Switch's default and we have no way how to detect whether "empty cluster ssh key" is for the first-time or intentional. In such case, the user needs to manually turn the switch off and so the form turns to clean (not-dirty state). Support from backend would be needed if this appears to be an issue.

Such behaviour might be annoying if the "no-cluster-key" case is common. I doubt but let's see.

mareklibra:
Rebased. Bunch of conflicts with #93.

src/components/clusterConfiguration/ClusterSshKeyField.tsx
```diff
@@ -0,0 +1,67 @@
...
+const sshPublicKeyHelperText = (
+  <>
+    Provide a SSH public key to debug OpenShift nodes after installation. To generate a new key, use
+    the <em>ssh-keygen -o</em> command and paste the value of <em>~/.ssh/id_rsa.pub</em> here.
+  </>
+);
+
+const label = 'Host SSH Public Key for troubleshooting after installation';
+
+const ClusterSshKeyField: React.FC<ClusterSshKeyFieldProps> = ({
```
mareklibra, line 26: This component handles so specific use-case that there is probably no reason to make it generic

PR #89, opened 2020-07-22
Handle pending-for-input host status

jtomasek:
Enhance host status popover to show validation groups and results

Screenshot 2020-07-22 at 13 23 25
Screenshot 2020-07-22 at 13 23 37
Screenshot 2020-07-22 at 13 29 39

### DIFF
--- a/src/components/hosts/HostStatus.tsx
+++ b/src/components/hosts/HostStatus.tsx
@@ -14,18 +14,23 @@ import {
   ConnectedIcon,
   OffIcon,
   UnknownIcon,
+  PendingIcon,
 } from '@patternfly/react-icons';
 import { Host } from '../../api/types';
+import { ValidationsInfo } from '../../types/hosts';
 import HostProgress from './HostProgress';
 import { HOST_STATUS_LABELS, HOST_STATUS_DETAILS } from '../../config/constants';
 import { getHumanizedDateTime } from '../ui/utils';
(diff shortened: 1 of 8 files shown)


src/components/hosts/HostValidationGroups.tsx
```diff
@@ -0,0 +1,104 @@
...
+              />
+            </AlertGroup>
+          </Fragment>
+        );
+      })}
+    </>
+  );
+};
+
+export default HostValidationGroups;
```
mareklibra, line 77: nit: can we move this to the end of the file?

src/components/hosts/HostStatus.tsx
```diff
@@ -56,22 +64,32 @@ const getPopoverContent = (host: Host) => {
...
-    <TextContent>
-      <Text>
-        {HOST_STATUS_DETAILS[status] || ''}
-        <br />
-        {toSentence(statusInfo)}
-      </Text>
-    </TextContent>
+    <>
+      <TextContent>
+        <Text>
```
mareklibra, line 81: Can we use `<Text component={TextVariants.p}>` to avoid `<br/>`?

jtomasek:
The messages need work as a follow up. We need to find a way to identify the problem from the backend data and provide relevant message on the frontend. This will resolve the problem of combining  backend and frontend messages and prepare us for eventual i18n implementation since the messages will be defined on the frontend.

PR #91, opened 2020-07-23
Add hostname-change feature

mareklibra:
---

---


src/components/hosts/Hostname.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const hostname = computeHostname(host, inventory);
+  const isHostnameChangeRequested = host.requestedHostname !== inventory.hostname;
+
+  return (
+    <>
+      <Button variant={ButtonVariant.link} isInline onClick={() => setOpen(true)}>
+        {hostname}
+        {isHostnameChangeRequested && ' *'}
+      </Button>
+      <EditHostModal
```
jtomasek, line 27: We're rendering the modal in 2 different places, we could just call `setShowEditHostModal` on the button click if we pass it as a prop (by adding it as a parameter to `hostToHostTableRow` function)

src/components/hosts/EditHostForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+      initialValues={initialValues}
+      initialStatus={{ error: null }}
+      validationSchema={validationSchema(initialValues, cluster.hosts)}
+      onSubmit={handleSubmit}
+    >
+      {({ handleSubmit, status, setStatus }) => (
+        <Form onSubmit={handleSubmit}>
+          <ModalBoxBody>
+            <GridGap>
+              {status.error && (
```
jtomasek, line 106: The `status.error` is never set since you're using global cluster alerts to handle the submit error. So the Alert below also never shows.

In other modals we've been using modal's own alert to report errors so I'd be in favor of that here as well.

src/components/hosts/EditHostForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const { requestedHostname } = host;
+  const { hostname } = inventory;
+
+  const initialValues: HostUpdateParams = {
+    hostId: host.id,
+    hostname: requestedHostname || '',
+  };
+
+  const handleSubmit = async (values: HostUpdateParams) => {
```
jtomasek, line 68:
```suggestion
  const handleSubmit = async (values: HostUpdateParams, formikActions: FormikHelpers<DiscoveryImageFormValues>,) => {
```

src/components/hosts/EditHostForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+        hostname: values.hostname,
+      },
+    ];
+
+    try {
+      const { data } = await patchCluster(cluster.id, params);
+      dispatch(updateCluster(data));
+    } catch (e) {
+      handleApiError(e, () =>
+        addAlert({ title: 'Failed to update host', message: getErrorMessage(e) }),
```
jtomasek, line 88:
```suggestion
        formikActions.setStatus({
            error: {
              title: 'Failed to update host',
              message: getErrorMessage(error),
            },
          });
```

src/components/hosts/EditHostForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+                isRequired
+                isDisabled={!canHostnameBeChanged(host.status)}
+              />
+            </GridGap>
+          </ModalBoxBody>
+          <ModalBoxFooter>
+            <Button
+              key="submit"
+              type={ButtonType.submit}
+              isDisabled={!!status.error /* TODO: use gray */}
```
jtomasek, line 136: the button should be disabled when `isSubmitting` or `isInvalid`. not status.error

src/components/hosts/EditHostForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+    try {
+      const { data } = await patchCluster(cluster.id, params);
+      dispatch(updateCluster(data));
+    } catch (e) {
+      handleApiError(e, () =>
+        addAlert({ title: 'Failed to update host', message: getErrorMessage(e) }),
+      );
+    }
+
+    onSuccess();
```
jtomasek, line 92: Call this only when API call is successful (line 85) on api error we'll show the alert above the form and not close it.

src/components/hosts/Hostname.tsx
mareklibra: I understand your point, but it's a single row which we replace by another single row call but with callback. If it is not a blocker, I prefer to keep it as it is.

src/components/hosts/EditHostForm.tsx
mareklibra: Thanks, fixed.

PR #92, opened 2020-07-23
Export New cluster modal

rawagner:


mareklibra:
This component is about to be reworked, most probably can be considered as deprecated.

rawagner:
Thats fine, but until then It would be nice to export it so we can use it for UHC integration

mareklibra:
As you wish, you have been warned :-)

PR #97, opened 2020-07-25
Adjust DNS checkbox label

andybraren:
Based on some feedback we received:

> nit: I would change from Red Hat DNS service to "Temporary OpenShift POC domain name"?

I tweaked this suggestion a bit to avoid saying POC (which we would have to spell out) while still emphasizing even further that this option should only be used for temporary clusters.

## Before

image

## After

image

### DIFF
--- a/src/components/clusterConfiguration/NetworkConfiguration.tsx
+++ b/src/components/clusterConfiguration/NetworkConfiguration.tsx
@@ -51,7 +51,7 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
       {!!managedDomains.length && (
         <CheckboxField
           name="useRedHatDnsService"
-          label="Use Red Hat's DNS service"
+          label="Use a temporary DNS domain from Red Hat"
           helperText="A base DNS domain will be provided by Red Hat's DNS service. Because the cluster's DNS can't be changed after cluster installation, this should only be used for temporary, non-production clusters."
           onChange={toggleRedHatDnsService}
         />


nirfarkas:
1. I think it's a great idea to emphasize the "temporary" nature of the service, though, doesn't the abbreviation 'DNS' already include 'Domain' inside (Domain Name System)?
2. It might be a good idea to place the help text in a '?' popover to avoid text "overload"?

andybraren:
> 1. I think it's a great idea to emphasize the "temporary" nature of the service, though, doesn't the abbreviation 'DNS' already include 'Domain' inside (Domain Name System)?

@nirfarkas I agree, it's weird. We also have a `Base DNS domain` field partly because OCP's documentation uses that wording, like [this example](https://docs.openshift.com/container-platform/4.5/networking/dns-operator.html). I'd like for us to work with a technical writer at some point to clarify/improve both our docs and the UI at the same time.

> 2. It might be a good idea to place the help text in a '?' popover to avoid text "overload"?

Good point. Interestingly, PatternFly's [guidance on helper text](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#helper-text) focuses mostly on form inputs, but the same idea probably applies to checkboxes too. I suspect there are parts of the OCP Web Console we should probably change too.

And this is just one of many places we (still) need to reduce text overload 😞. We should probably have a dedicated story for that. I started but never quite finished an ACM/OCM/Installer help text alignment doc [here](https://docs.google.com/document/d/1rhWOOUbwobxiHNPy9HyflSUpPVml6PUQCYNKlg4KLHo/edit#).

For this field, maybe "Only enable for temporary, non-production clusters" would be the need-to-know helper text (based on PatternFly's guidance), and the text as we have it now would be placed in a `?`?

nirfarkas:
image
Something like this?

andybraren:
@nirfarkas I guess even this tweak wasn't enough (via kni-devel):

> Same comment from an earlier review on "temporary DNS" - do not imply it is from "Red Hat DNS"

Maybe mentioning "Red Hat" is the issue. How about a slight tweak of your proposal:

> Use a temporary DNS service (?)
> A base domain will be provided for temporary, non-production clusters

🤷‍♂️

ohadlevy:
> @nirfarkas I guess even this tweak wasn't enough (via kni-devel):
> 
> > Same comment from an earlier review on "temporary DNS" - do not imply it is from "Red Hat DNS"
> 
> Maybe mentioning "Red Hat" is the issue. How about a slight tweak of your proposal:
> 
> > Use a temporary DNS service (?)
> > A base domain will be provided for temporary, non-production clusters
> 
> man_shrugging

I believe we wanted to add a time limit on it as well (e.g. 60 days like our eval subscriptions) so its extremely clear this is not a production instance and we may delete your dns record after that time period.

PR #100, opened 2020-07-28
Enable kubeconfig download for selected cluster statuses only

mareklibra:


src/components/hosts/utils.ts
```diff
@@ -30,6 +30,9 @@ export const canEditRole = (clusterStatus: Cluster['status'], status: Host['stat
     'pending-for-input',
   ].includes(status);
 
+export const canDownloadKubeconfig = (clusterStatus: Cluster['status']) =>
```
tsorya, line 33: we have 2 different kubeconfigs, first one is kubeconfig-noingress and second is kubeconfig (with it's own api)
kubeconfig-noingress will be available in ['ready', 'installing', 'finalizing'].
Though kubeconfig only in installed state

src/components/hosts/utils.ts
```diff
@@ -30,6 +30,9 @@ export const canEditRole = (clusterStatus: Cluster['status'], status: Host['stat
...
+  ['ready', 'installing', 'finalizing', 'installed'].includes(clusterStatus);
```
filanov, line 34: I don't think that we have kubeconfig when cluster is ready.

src/components/hosts/utils.ts
mareklibra: Thanks. This is ok since https://github.com/mareklibra/facet-lib/pull/100/files#diff-3fac78db77e5a14dea183ab0cd17fcebL12 . The button should be enabled in any of these cases.

src/components/hosts/utils.ts
mareklibra: thanks!

src/components/hosts/utils.ts
filanov: Kubeconfig in  ready? we generate it only on install no?

tsorya: do we have files in ready state at all?

PR #101, opened 2020-07-28
Explicitely warn about failed work nodes in the "installed" stated

mareklibra:


andybraren:
Nice! I believe when the install fails or is canceled an inline alert appears directly below the progress bar, so I positioned this inline alert in the same place in [the design](https://issues.redhat.com/browse/MGMT-1276). Do you think that position would make sense? FYI @nirfarkas

image

mareklibra:
> Do you think that position would make sense? 

Sure, moving.

PR #117, opened 2020-07-31
No Cluster install abort when a host is in installing-pending-user-action

mareklibra:


mareklibra:
There's recently review-in-progress of the states enabling/disabling the installation abort. Potential outcome will be implemented in follow-ups.

PR #129, opened 2020-08-06
Fix VIP re-validation after cluster install abort

mareklibra:


mareklibra:
The lazy validation schema is causing the use of stale validations when hostSubnets are changed (`vipRangeValidationSchema` receives old (empty) `hostSubnets` no matter the Formik is already provided with updated validation schema of non-empty hostSubnets).
Fixed by passing the validation in `vipRangeValidationSchema` when the subnet is unknown.
Another approach would be forcing revalidation but this is probably not feasible without a hacky workaround.

PR #131, opened 2020-08-07
Enhance Host installation progress for info stageStartedAt and stageUpdatedAt

mareklibra:


mareklibra:
Considering the data recently received from backend, I am not sure if this change brings much value to the user.
It seems that the `stageUpdatedAt` works recently just for the `writing image` stage, since the sub-progress is updated during the process.

Otherwise even long-running stages (like the `waiting for control plane`) stay untouched long time till reaching next stage.

---


---


---


---

mareklibra:
@andybraren Time passed, thoughts evolved, do we want something like this?

andybraren:
Hm, it's been a while so I've lost some context, but displaying `Step started at` even if the `updated 3 seconds ago` part only appears for the `Writing image to disk` step still seems like it could be somewhat useful to me. WDYT @nirfarkas? Worth the additional text?

nirfarkas:
LGTM

jtomasek (approved):

mareklibra:
Rebased

jtomasek (approved):

PR #137, opened 2020-08-10
CIDR fields are required

mareklibra:
Backend provides defaults for them.

In addition, new `.requiredOnceSet()` validation schema is introduced and used for the `baseDNS` and both VIPs fields.


mareklibra:
required

PR #138, opened 2020-08-10
Automatically format the host prefix when non numeric value is provided

jtomasek:


src/components/clusterConfiguration/AdvancedNetworkFields.tsx
```diff
@@ -1,31 +1,45 @@
...
-    />
-    <InputField
-      name="serviceNetworkCidr"
-      label="Service Network CIDR"
-      helperText="The IP address pool to use for service IP addresses. You can enter only one IP address pool. If you need to access the services from an external network, configure load balancers and routers to manage the traffic."
-      isRequired
-    />
-  </>
-);
+  return (
```
mareklibra, line 17: Why do we need this `return` here?

jtomasek: Because I've added some variable definitions in the component so it was necessary to add return statement to the function component.

mareklibra: Oh, I was looking incorrectly at the diff.

PR #143, opened 2020-08-12
Add HTML IDs to cluster progress status info

mareklibra:
With this patch the cluster status HTML element on the cluster detail page looks like:
```
<dd id="cluster-progress-status-value" data-pf-content="true" class="">Installing</dd>
````

### DIFF
--- a/src/components/clusterDetail/ClusterProgress.tsx
+++ b/src/components/clusterDetail/ClusterProgress.tsx
@@ -67,8 +67,16 @@ const ClusterProgress: React.FC<ClusterProgressProps> = ({ cluster }) => {
   return (
     <>
       <DetailList>
-        <DetailItem title="Started on" value={getHumanizedDateTime(cluster.installStartedAt)} />
-        <DetailItem title="Status" value={getInstallationStatus(cluster)} />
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mareklibra** [comment]
cc @ukalifon

File: src/components/clusterDetail/ClusterProgress.tsx [hunk 1]
```diff
@@ -67,8 +67,16 @@ const ClusterProgress: React.FC<ClusterProgressProps> = ({ cluster }) => {
   return (
     <>
       <DetailList>
-        <DetailItem title="Started on" value={getHumanizedDateTime(cluster.installStartedAt)} />
-        <DetailItem title="Status" value={getInstallationStatus(cluster)} />
+        <DetailItem
+          title="Started on"
+          value={getHumanizedDateTime(cluster.installStartedAt)}
+          idPrefix="cluster-progress-started-on"
```
jtomasek, line 73: Nit: we could simply call it `id`

mareklibra: But we are further appending it so I thought this is clear that such ID will not appear in the DOM

PR #146, opened 2020-08-17
Trim ssh public key before validation

mareklibra:


mareklibra:
I don't like changing the value on the fly, the user might be confused. I would trim it again right before sending. wdyt?

jtomasek:
`onBlur` happens when user leaves the field so it does not collide with the user input. Hmm, trimming it after submit would still require to trim in validation.

ohadlevy:
> I don't like changing the value on the fly, the user might be confused. I would trim it again right before sending. wdyt?

I think thats the last thing I would expect, e.g. whats in the field is what I expect you to validate and send, so for me as a user I would go and manually change it before sending it... my vote is to trim it once you paste it in.

worth thinking of allowing multiple ssh keys, but I guess thats outside the scope of this pr..

mareklibra:
Ok, the field is altered once the user changes focus. I keep the trimming in validation anyway for the case that the field already contains a value and the user adds extra whitespaces. Without it, the validation would be failing until the user clicks somewhere else.

PR #147, opened 2020-08-17
Edit host is disabled when cluster is in installing or preparing-for-installation

mareklibra:


src/components/hosts/utils.ts
```diff
@@ -31,6 +31,9 @@ export const canEditRole = (clusterStatus: Cluster['status'], status: Host['stat
     'pending-for-input',
   ].includes(status);
 
+export const canEditHost = (clusterStatus: Cluster['status']) =>
+  !['preparing-for-installation', 'installing'].includes(clusterStatus);
```
jtomasek, line 35: I think the condition should be pretty much the same as `canEditRole`.

PR #150, opened 2020-08-17
Add ClusterRowMapper

mareklibra:
To inject HTML ID for `<tr>` in the clusters list to simplify tests.

Example of the row id:
```
  <tr id="cluster-row-test-infra-cluster"></tr>
```


src/components/clusters/ClustersTable.tsx
```diff
@@ -44,6 +46,12 @@ const columns = [
 
 const getStatusCell = (row: IRow) => row.cells?.[3] as HumanizedSortable | undefined;
 
+const ClusterRowWrapper = (props: RowWrapperProps) => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  return <RowWrapper {...props} id={`cluster-row-${props.row.props?.name}`} />;
```
jtomasek, line 52: Nit: this yields the same result but aligns to the props type. The missing `id` can't be worked out though...
```suggestion
  return <RowWrapper {...props} id={`cluster-row-${props?.rowProps?.rowKey}`} />;
```

mareklibra: The `props.row` is provided in this context, but adding the extra `?` just for the case.
The use of `cluster.id` might be cumbersome for the tests, the name is easier to handle.

PR #157, opened 2020-08-20
Add backend-based cluster validations

jtomasek:
image

The validation messages are taken directly from the backend and need polishing. Ideally we define validation messages on the client side and map them to backend validation id+status combination. This will allow easy i18n.

### DIFF
--- a/src/components/clusterConfiguration/ClusterValidationSection.tsx
+++ b/src/components/clusterConfiguration/ClusterValidationSection.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash';
 import { FormikErrors } from 'formik';
 import {
   AlertGroup,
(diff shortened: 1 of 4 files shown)


src/components/clusterConfiguration/ClusterValidationSection.tsx
```diff
@@ -41,6 +44,27 @@ const ClusterValidationSection: React.FC<ClusterValidationSectionProps> = ({
   const errorFields = Object.keys(formErrors);
   const ready = cluster.status === 'ready' && !errors.length && !errorFields.length && !dirty;
 
+  const validationsInfo = React.useMemo<ValidationsInfo>(
```
rawagner, line 47: you can merge all these `useMemo`s into one

rawagner (approved):

PR #160, opened 2020-08-21
Add auto-assign role

rawagner:


src/config/constants.ts
```diff
@@ -29,7 +29,11 @@ export const getProductBrandingCode = () => 'redhat';
 export const POLLING_INTERVAL = 10 * 1000;
 export const EVENTS_POLLING_INTERVAL = 10 * 1000;
 
-export const HOST_ROLES = ['worker', 'master'];
+export const HOST_ROLES: { [key in HostRoleUpdateParams]: string } = {
+  'auto-assign': 'Automatic',
```
rawagner, line 33: auto-assign is shown as Automatic.
I've also changed other roles to be upper case

src/components/ui/SimpleDropdown.tsx
```diff
@@ -4,50 +4,52 @@ import { CaretDownIcon } from '@patternfly/react-icons';
...
-  const items = values.map((value) => (
-    <DropdownItem key={value} id={value}>
-      {value}
+  const dropdownItems = Object.keys(items).map((key) => (
+    <DropdownItem key={key} id={key}>
+      {items[key]}
     </DropdownItem>
   ));
 
-  const onRoleSelect = React.useCallback(
```
rawagner, line 25: Renamed a few things as is was specific to Role, but the dropdown is supposed to be generic

jtomasek (approved):

PR #161, opened 2020-08-21
Add VIP DHCP configuration

rawagner:
Just a simple switch to enable VIP DHCP allocation

DHCP off - user has to specify api and ingress VIP
Screenshot from 2020-08-21 12-53-55

DHCP on - api and ingress VIP fields are hidden
Screenshot from 2020-08-21 12-53-44

JIRA https://issues.redhat.com/browse/MGMT-1751

@andybraren I guess we will want to add some help text and possibly rename the switch too.

### DIFF
--- a/src/components/clusterConfiguration/BasicNetworkFields.tsx
+++ b/src/components/clusterConfiguration/BasicNetworkFields.tsx
@@ -1,13 +1,14 @@
 import React from 'react';
 import { HostSubnets, ClusterConfigurationValues } from '../../types/clusters';
-import { InputField, SelectField } from '../ui/formik';
+import { InputField, SelectField, SwitchField } from '../ui/formik';
 import { useFormikContext } from 'formik';
 
 type BasicNetworkFieldsProps = {
   hostSubnets: HostSubnets;
+  isAdvanced: boolean;
 };
 
-const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets }) => {
+const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets, isAdvanced }) => {
   const { validateField, values } = useFormikContext<ClusterConfigurationValues>();
   const { hostSubnet } = values;
 
@@ -38,20 +39,25 @@ const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets }) 
         }}
         isRequired
       />
-      <InputField
-        label="API Virtual IP"
-        name="apiVip"
-        helperText={`Virtual IP used to reach the OpenShift cluster API. Make sure that the VIP's are unique and not used by any other device on your network${subnet}.`}
-        isRequired
-        isDisabled={!hostSubnets.length}
-      />
-      <InputField
-        name="ingressVip"
-        label="Ingress Virtual IP"
-        helperText="Virtual IP used for cluster ingress traffic."
-        isRequired
-        isDisabled={!hostSubnets.length}
-      />
+      {isAdvanced && <SwitchField label="Use VIP DHCP allocation" name="vipDhcpAllocation" />}
+      {(!values.vipDhcpAllocation || !isAdvanced) && (
+        <>
+          <InputField
+            label="API Virtual IP"
+            name="apiVip"
+            helperText={`Virtual IP used to reach the OpenShift cluster API. Make sure that the VIP's are unique and not used by any other device on your network${subnet}.`}
+            isRequired
+            isDisabled={!hostSubnets.length}
+          />
+          <InputField
+            name="ingressVip"
+            label="Ingress Virtual IP"
+            helperText="Virtual IP used for cluster ingress traffic."
+            isRequired
+            isDisabled={!hostSubnets.length}
+          />
+        </>
+      )}
     </>
   );
 };
--- a/src/components/clusterConfiguration/NetworkConfiguration.tsx
+++ b/src/components/clusterConfiguration/NetworkConfiguration.tsx
@@ -16,8 +16,8 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
   hostSubnets,
   managedDomains,
 }) => {
-  const [type, setType] = React.useState<'basic' | 'advanced'>('basic');
   const { setFieldValue, initialValues, values } = useFormikContext<ClusterConfigurationValues>();
+  const [isAdvanced, setAdvanced] = React.useState<boolean>(!!values.vipDhcpAllocation);
   const { name: clusterName, baseDnsDomain, useRedHatDnsService } = values;
 
   const backToBasic = () => {
@@ -26,7 +26,7 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
     setFieldValue('clusterNetworkCidr', clusterNetworkCidr);
     setFieldValue('clusterNetworkHostPrefix', clusterNetworkHostPrefix);
     setFieldValue('serviceNetworkCidr', serviceNetworkCidr);
-    setType('basic');
+    setAdvanced(false);
   };
 
   const baseDnsHelperText = (
@@ -81,7 +81,7 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
         <Radio
           id="networkConfigurationTypeBasic"
           name="networkConfigurationType"
-          isChecked={type === 'basic'}
+          isChecked={!isAdvanced}
           value="basic"
           onChange={backToBasic}
           label="Basic"
@@ -92,15 +92,15 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
           id="networkConfigurationTypeAdvanced"
           name="networkConfigurationType"
           value="advanced"
-          isChecked={type === 'advanced'}
-          onChange={() => setType('advanced')}
+          isChecked={isAdvanced}
+          onChange={() => setAdvanced(true)}
           label="Advanced"
           description="Configure a custom networking type and CIDR ranges."
           isLabelWrapped
         />
       </FormGroup>
-      <BasicNetworkFields hostSubnets={hostSubnets} />
-      {type === 'advanced' && <AdvancedNetworkFields />}
+      <BasicNetworkFields hostSubnets={hostSubnets} isAdvanced={isAdvanced} />
+      {isAdvanced && <AdvancedNetworkFields />}
     </>
   );
 };
--- a/src/components/clusterConfiguration/ClusterConfigurationForm.tsx
+++ b/src/components/clusterConfiguration/ClusterConfigurationForm.tsx
@@ -114,6 +114,15 @@ const ClusterConfigurationForm: React.FC<ClusterConfigurationFormProps> = ({
         params.sshPublicKey = cluster.imageInfo.sshPublicKey;
       }
 
+      if (values.vipDhcpAllocation) {
+        delete params.apiVip;
+        delete params.ingressVip;
+        const cidr = hostSubnets
+          .find((hn) => hn.humanized === values.hostSubnet)
+          ?.subnet.toString();
+        params.machineNetworkCidr = cidr;
+      }
+
       const { data } = await patchCluster(cluster.id, params);
       formikActions.resetForm({
         values: getInitialValues(data, managedDomains),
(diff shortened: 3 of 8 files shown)


nirfarkas:
Does enabling the DHCP allocation provides instant feedback to the user? Does the user see these action implications immediately (the vips)? If the answer is "Yes" it's a good use case for using a toggle, but if not, we should use a checkbox instead. This is because the switches (toggles) are used when the change is immediate and checkboxes are for setting a preference for another action that the user needs to do.

rawagner:
> Does enabling the DHCP allocation provides instant feedback to the user? Does the user see these action implications immediately (the vips)? If the answer is "Yes" it's a good use case for using a toggle, but if not, we should use a checkbox instead. This is because the switches (toggles) are used when the change is immediate and checkboxes are for setting a preference for another action that the user needs to do.

toggling DHCP will show/hide VIP fields. Its similar to what we do just below with `Use the same host discovery SSH key` which when turned off will show a text area field.

andybraren:
We very recently got some consensus within OpenShift UXD around when to use checkboxes vs switches ([see summary](https://github.com/openshift/openshift-origin-design/issues/392#issuecomment-672835290)) since the idea is so similar.

PatternFly doesn't provide exact guidance, but we're thinking that if the change is saved and applied immediately (like settings toggles in an Operator Details page) then a switch should be used, but in forms where a second confirmation button click is still needed, [use a checkbox](https://uxplanet.org/checkbox-vs-toggle-switch-7fc6e83f10b8). Either one could potentially reveal additional parameters when enabled.

That would probably mean changing both this new DHCP switch and the SSH key switches to checkboxes instead. Does that sound right to you @nirfarkas?

andybraren:
> @andybraren I guess we will want to add some help text and possibly rename the switch too.

I admit that I don't have a good enough understanding of the mechanism to be able to describe it myself. [The doc](https://docs.google.com/document/d/1OBn0scklFQZ6HnkYgtZdNGWaEfhz81L30I_Ezb7BDRQ/edit#) linked in Jira helped a little, but I could use a "for dummies" version. 🙂

Since having DHCP allocation reduces the number of fields that the user needs to fill out, I assume (?) that we would want it to be the default state and have users opt into a "Custom" option to reveal the API and Ingress VIP fields. Having DHCP allocation on by default is suggested [here](https://docs.google.com/document/d/1OBn0scklFQZ6HnkYgtZdNGWaEfhz81L30I_Ezb7BDRQ/edit?disco=AAAAJ2PV1GE).

So maybe we flip the behavior and text a little to be:

> [ ] Configure custom DHCP allocation (unchecked by default)
> Manually provide API and Ingress virtual IP ranges.

(Note: "Configure" aligns with the [new wording](https://github.com/mareklibra/facet-lib/pull/159) in the Discovery ISO modal)

Does something like this make sense @rawagner @nirfarkas? Maybe we can ask the team for more thoughts.

rawagner:
> > @andybraren I guess we will want to add some help text and possibly rename the switch too.
> 
> I admit that I don't have a good enough understanding of the mechanism to be able to describe it myself. [The doc](https://docs.google.com/document/d/1OBn0scklFQZ6HnkYgtZdNGWaEfhz81L30I_Ezb7BDRQ/edit#) linked in Jira helped a little, but I could use a "for dummies" version. 
> 
> Since having DHCP allocation reduces the number of fields that the user needs to fill out, I assume (?) that we would want it to be the default state and have users opt into a "Custom" option to reveal the API and Ingress VIP fields. Having DHCP allocation on by default is suggested [here](https://docs.google.com/document/d/1OBn0scklFQZ6HnkYgtZdNGWaEfhz81L30I_Ezb7BDRQ/edit?disco=AAAAJ2PV1GE).
> 

How about basic network config would contain only Available Subnet configuration and advanced would reveal option to disable DHCP and configure VIPs among other things ? The basic configuration would be really simple.

> So maybe we flip the behavior and text a little to be:
> 
> > [ ] Configure custom DHCP allocation (unchecked by default)

I think there's no DHCP involved if user chooses to specify VIPs

> > Manually provide API and Ingress virtual IP ranges.
> 
> (Note: "Configure" aligns with the [new wording](https://github.com/mareklibra/facet-lib/pull/159) in the Discovery ISO modal)


> 
> Does something like this make sense @rawagner @nirfarkas? Maybe we can ask the team for more thoughts.

andybraren:
> How about basic network config would contain only Available Subnet configuration and advanced would reveal option to disable DHCP and configure VIPs among other things ? The basic configuration would be really simple.

I like that idea, and [originally hoped](https://issues.redhat.com/browse/MGMT-1639?focusedCommentId=14400879&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14400879) that would be the case. Sounds good.

Will keep thinking about the text.

nirfarkas:
I've read the thread (even twice (: ), but still having trouble to understand:
1. What should be the default? Use DHCP allocation \ Custom vips?
2. I think that if we choose to use an unchecked checkbox with the label "Disable DHCP allocation" it will be very confusing. The label should be positive (to avoid double negation).
3. I think that if we want the user to skip the advanced section we should incline to "Use DHCP allocation" checked by default, otherwise, the user has to fill in the vips manually.
4. I totally agree that we need to state the risk of choosing the DHCP allocation (popover or under the field?)

rawagner:
As per my understanding :+1: 
> 1. What should be the default? Use DHCP allocation \ Custom vips?

DHCP will be default, use can turn it off in Advanced networking section. Basic section will contain only Available Subnets dropdown.

> 2. I think that if we choose to use an unchecked checkbox with the label "Disable DHCP allocation" it will be very confusing. The label should be positive (to avoid double negation).

I agree, when a user switches to Advanced section he will see `DHCP allocation` checkox checked

> 3. I think that if we want the user to skip the advanced section we should incline to "Use DHCP allocation" checked by default, otherwise, the user has to fill in the vips manually.

+1 DHCP will be on by default

> 4. I totally agree that we need to state the risk of choosing the DHCP allocation (popover or under the field?)

+1. But where do we put the warning in basic section if we are not going to have DHCP checkbox there ?

rawagner:
I just talked to @ori-amizur and he suggested that we should also show CIDR value in `Available subnets` dropdown. Something like `192.168.126.0/24 (192.168.126.1 - 192.168.126.254)`

Also the DHCP is not going to be enabled by default on backend for now, so maybe we should align too and put the enablement of the feature into advanced config. Once backend will enable it by default we can update.

rawagner:
For now, the DHCP configuration lives in Advanced section. We will change that once backend decides to turn the feature on by default.

Basic config
Screenshot from 2020-08-26 13-45-01

Advanced config (DHCP off by default)
Screenshot from 2020-08-26 13-45-25

DHCP on
Screenshot from 2020-08-26 13-45-15


I've also updated `Available subnets` values to include CIDR value

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { InputField, SelectField, SwitchField } from '../ui/formik';
 import { useFormikContext } from 'formik';
 
 type BasicNetworkFieldsProps = {
   hostSubnets: HostSubnets;
+  isAdvanced: boolean;
 };
 
-const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets }) => {
+const BasicNetworkFields: React.FC<BasicNetworkFieldsProps> = ({ hostSubnets, isAdvanced }) => {
```
rawagner, line 11: yes, i know this is ugly. Passing `isAdvanced` to `BasicNetworkFields` - but we need this since we want to place the DHCP switch between basic fields. I could refactor the code but once the DHCP will be turned off we will want to go back to Basic and AdvancedNetworkingFields as we have now

rawagner:
We will also want to move from Switch component to Checkbox per @nirfarkas recommendation. I will create a ticket to move all Switches not just the DHCP one

src/components/clusterConfiguration/BasicNetworkFields.tsx
ori-amizur: I think DHCP should be considered basic, because at effects the behavior of basic networking fields

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -92,15 +92,15 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
           id="networkConfigurationTypeAdvanced"
           name="networkConfigurationType"
           value="advanced"
-          isChecked={type === 'advanced'}
-          onChange={() => setType('advanced')}
+          isChecked={isAdvanced}
```
ErezAlster, line 95: is it me or this parameter appears twice?

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -92,15 +92,15 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
...
+          onChange={() => setAdvanced(true)}
```
ErezAlster, line 96: don't we need to use the the value of the radio button instead of using true?

eranco74 (approved):
/lgtm

src/components/clusterConfiguration/NetworkConfiguration.tsx
rawagner: You actually cannot uncheck the radio button.
See screenshot

Screenshot from 2020-08-26 13-45-15

If Advanced Radio is checked, you cannot click on it to uncheck it, you have to click on Basic Radio. 
I guess that also answers your other question. We have two radio buttons which are mutually exclusive. You can select only one.

ErezAlster: So lets say that I'm trying to add a cluster and clicking on Advanced and then want to switch back to basic mode. what are my options to do it?

src/components/clusterConfiguration/BasicNetworkFields.tsx
eranco74: @ori-amizur what do you mean by it effects the behavior...
DHCP is not the default hence it's under advanced options..

src/components/clusterConfiguration/NetworkConfiguration.tsx
rawagner: The only option is to click on Basic radio

src/components/clusterConfiguration/BasicNetworkFields.tsx
rawagner: We will add it to Basic once its on by default in backend. This is just a temporary measure

rawagner: So this https://github.com/mareklibra/facet-lib/pull/161/files#diff-f936227366e52be8b7bf47d374fbc45cR42-R59 will move to Advanced.

rawagner:
So just to summarize - DHCP is still experimental and turned off by default in backend. User can turn it on in Advanced section in UI. Once the DHCP feature gets stable it will be turned on by default in backend and we will adjust the UI accordingly.

src/components/clusterConfiguration/BasicNetworkFields.tsx
ori-amizur: @eranco74 If you set DHCP, then vips filling should be disabled.  Vips filling is basic.
In addition, validations will behave differently.

avishayt: 1. It is an experimental feature, not advanced.  If it's too complicated now to make a new section, then advanced is the best option.
2. Can we somehow indicate to the user that this option exists?  For example near where the user enters the VIPs write something like "If you don't have available IP addresses you may try the experimental option of automatically allocating them".

ori-amizur: Also, machine-cidr selection should have different functionality, since only in DHCP mode, the UI should include it in the UpdateCluster request

ronniel1: @avishayt why do you say it is experimental?

@ori-amizur Once DHCP is default, VIPs setting is not basic anymore, its just an output field.
@rawagner Notice that when DHCP is the default, the user must still see the values for the VIPs, so it cannot be hidden in the advanced section. I think that when DHCP is enabled, the VIPs field should turn into output fields

avishayt: It's not production-grade from what I understand because there are known cases where the lease will expire and then the cluster is hosed.

jtomasek (approved):

src/components/clusterConfiguration/BasicNetworkFields.tsx
ori-amizur: @avishayt Can you indicate what cases?

rawagner:
Im merging now so we can build. Lets create followup issues for remaining changes.
https://issues.redhat.com/browse/MGMT-1933 - dont hide VIPs when DHCP is on, just disable

PR #162, opened 2020-08-21
Add Sentry exception logging

rawagner:


src/api/utils.ts
```diff
@@ -1,28 +1,30 @@
...
   } else {
+    let message = `Error config: ${error.config}\n`;
     if (error.response) {
       // The request was made and the server responded with a status code
       // that falls out of the range of 2xx
-      console.error('Response data:', error.response.data);
-      console.error('Response status:', error.response.status);
-      console.error('Response headers:', error.response.headers);
+      message += `Response data: ${error.response.data}\n`;
+      message += `Response status: ${error.response.status}\n`;
```
ErezAlster, line 16: Can we add these parameters as more structured data instead of a string?
Just thinking about the ability to search easily.

I'm not familiar to sentry and what it offers

src/components/clusterConfiguration/ClusterConfigurationForm.tsx
```diff
@@ -103,7 +104,7 @@ const ClusterConfigurationForm: React.FC<ClusterConfigurationFormProps> = ({
         return formikActions.setFieldError('name', `Name "${values.name}" is already taken.`);
       }
     } catch (e) {
-      console.error('Failed to perform unique cluster name validation.', e);
+      captureException('Failed to perform unique cluster name validation.', e);
```
ErezAlster, line 107: what is the right parameters order? you are missing string and error in different api calls to capture exceptions.

src/sentry.ts
```diff
@@ -0,0 +1,18 @@
...
+export enum SEVERITY {
+  ERROR = 'error',
+  WARN = 'warning',
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+export const captureException = (error, message?: string, severity: SEVERITY = SEVERITY.ERROR) => {
+  if (ocmClient) {
+    message && Sentry.captureMessage(message, SEVERITY[severity]);
```
ErezAlster, line 13: Same comment as the structured parameters to Sentry.
maybe we can send additional parameters instead of string only.

src/components/clusterConfiguration/ClusterConfigurationForm.tsx
rawagner: good catch. Order was reversed. Fixed

src/api/utils.ts
rawagner: Yeah, Im not sure how it is going to look like. Im still waiting for access to Sentry. Im sure we will tweak stuff once we can see how it looks in Sentry UI

src/api/utils.ts
```diff
@@ -1,28 +1,30 @@
...
+      message += `Response headers: ${error.response.headers}`;
```
jtomasek, line 17: missing `\n` at the end? Same for the 2 below.

jtomasek (changes requested):

src/api/utils.ts
rawagner: there's no more lines after, so no need for `\n`

mareklibra:
Closed in favour of #188

PR #170, opened 2020-08-31
Fix typescript error for innerRef prop

mareklibra:


mareklibra:
No more needed, fixed by other PR.

PR #174, opened 2020-09-01
Fix Host Events title for hostname

mareklibra:
Use `requestedHostname`

### DIFF
--- a/src/components/hosts/HostsTable.tsx
+++ b/src/components/hosts/HostsTable.tsx
@@ -134,7 +134,7 @@ const isHostShown = (skipDisabled: boolean) => (host: Host) =>
 const HostsTable: React.FC<HostsTableProps> = ({ cluster, skipDisabled = false }) => {
   const { addAlert } = React.useContext(AlertsContext);
   const [showEventsModal, setShowEventsModal] = React.useState<
-    { cluster: Cluster['id']; host: Host['id'] } | undefined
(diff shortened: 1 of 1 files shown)


src/components/hosts/HostsTable.tsx
```diff
@@ -299,8 +299,13 @@ const HostsTable: React.FC<HostsTableProps> = ({ cluster, skipDisabled = false }
...
 
-  const getHostInventory = (hostId: Host['id']): Inventory =>
-    hostRows.find((rowData) => rowData.host?.id === hostId)?.inventory;
+  let eventsModalTitle = 'Host Events';
+  if (showEventsModal) {
+    const hostRow = hostRows.find((rowData) => rowData.host?.id === showEventsModal.host);
+    const hostname =
+      hostRow?.host.requestedHostname || hostRow?.inventory?.hostname || showEventsModal.host;
+    eventsModalTitle += `: ${hostname}`;
+  }
```
jtomasek, line 308: I think this would be much better to calculate when calling `setShowEventsModal` in `onViewHostEvents` as that is exactly the moment when we need the hostname. Ideally the object passed to `setShowEventsModal` would be
```
{ clusterId: Cluster['id'], hostId: Host['id'], hostname: string }
```

jtomasek (changes requested):

src/components/hosts/HostsTable.tsx
mareklibra: done

PR #179, opened 2020-09-03
Downgrade from Yarn to NPM

mareklibra:
The `package-lock.json` is generated from scratch instead of converting from `yarn.lock`.

TODO:
- [x] "inferred type" issue, probably node_modules resolving issue


mareklibra:
It is slower compared to `yarn` but works

mareklibra:
No more needed

PR #185, opened 2020-09-08
Change Draft-cluster icon to file-alt

mareklibra:


mareklibra:
@andybraren Will we stay with the default colour? It was yellow formerly.

andybraren:
The [color guideline](https://www.patternfly.org/v4/design-guidelines/styles/colors) isn't super clear about when to use Color-100 vs link--Color, but I think using the blue link color works here. LGTM, thank you!

jtomasek (approved):

PR #196, opened 2020-09-11
Fix seting of machine_cidr

mareklibra:
Prior this fix, the host subnet was not selected even if the dropdown box
showed a first available subnet by default.
Consequently, the machine_cidr was not passed during cluster patch and
so DHCP allocation failed.


romfreiman:
ETA?

jtomasek (approved):

PR #198, opened 2020-09-13
MGMT-1015 Remove time zone from dates following API fix

empovit:
This PR is required to make the fix in the assisted service API work https://github.com/openshift/assisted-service/pull/350/.

Until the API fix is merged, this PR may cause incorrect dates to be shown in the UI (wrong timezone offset), but at least they will be meaningful. If the API PR is merged first, invalid date format errors will appear in the date/time fields instead of dates.


src/components/ui/utils.ts
```diff
@@ -2,13 +2,12 @@ import { DASH } from '../constants';
 
 export const getHumanizedDateTime = (dateTime?: string) => {
   if (!dateTime) return DASH;
-  // NOTE(jtomasek): server does not return ISO-8601 formatted date, therefore we need to adjust it
-  const date = new Date(`${dateTime}Z`);
+  const date = new Date(`${dateTime}`);
```
jtomasek, line 5:
```suggestion
  const date = new Date(dateTime);
```

src/components/ui/utils.ts
```diff
@@ -2,13 +2,12 @@ import { DASH } from '../constants';
...
   return date.toLocaleString();
 };
 
 export const getHumanizedTime = (dateTime?: string) => {
   if (!dateTime) return DASH;
-  const date = new Date(`${dateTime}Z`);
+  const date = new Date(`${dateTime}`);
```
jtomasek, line 11:
```suggestion
  const date = new Date(dateTime);
```

jtomasek (changes requested):

jtomasek (approved):

PR #200, opened 2020-09-13
Miscellaneous wording fixes

andybraren:
- Standardizes on a capital `D` in `Discovery ISO`
- Removes the mention of DNS records needing to resolve now that we have the temporary option, and adds a mention of being on the same L2 network (based on some recent discussion, FYI @ohadlevy)
- Removes redundant text in the table's empty state
- Small wording tweaks and fixes


src/components/clusterConfiguration/discoveryImageModal.tsx
```diff
@@ -25,7 +25,7 @@ export const DiscoveryImageModalButton: React.FC<DiscoveryImageModalButtonProps>
         onClick={() => setIsModalOpen(true)}
         id="button-download-discovery-iso"
       >
-        Download discovery ISO
+        Download Discovery ISO
```
ohadlevy, line 28: perhaps Generate?

jtomasek (approved):

PR #201, opened 2020-09-14
Update pull secret link and description text

andybraren:
OCM has a dedicated page that lets users download/copy their pull secret, so using that URL instead of the UPI instructions probably makes sense.

This PR also tweaks the description text's wording slightly to not mention OCM while the user is already within OCM. There are more updates being discussed in [MGMT-1917](https://issues.redhat.com/browse/MGMT-1917).

image


src/components/clusters/NewClusterPage.tsx
```diff
@@ -36,13 +36,13 @@ import { captureException } from '../../sentry';
 
 const pullSecretHelperText = (
   <>
-    The pull secret can be obtained from the Pull Secret page on the{' '}
+    Your Red Hat account's pull secret will be used by default. The pull secret of an account can
```
jtomasek, line 39:
```suggestion
    Your Red Hat account's pull secret will be used by default. The pull secret of an account can be
```

jtomasek (changes requested):
Prettier check is failing, run `yarn run prettier` to fix it automatically.

mareklibra (approved):

jtomasek (approved):

PR #221, opened 2020-09-21
Use bullets in list of validation errors

mareklibra:
Follow-up for: https://github.com/openshift/assisted-service/pull/351

### DIFF
--- a/src/components/clusterConfiguration/ClusterValidationSection.tsx
+++ b/src/components/clusterConfiguration/ClusterValidationSection.tsx
@@ -9,11 +9,10 @@ import {
   ButtonVariant,
   Split,
   SplitItem,
-  TextList,
-  TextListItem,
-  TextContent,
-  Text,
-  TextVariants,
+  Flex,
+  FlexItem,
+  List,
+  ListItem,
 } from '@patternfly/react-core';
 import { TimesIcon } from '@patternfly/react-icons';
 import { Cluster } from '../../api/types';
@@ -89,14 +88,16 @@ const ClusterValidationSection: React.FC<ClusterValidationSectionProps> = ({
               title="Cluster is not ready to be installed yet"
               isInline
             >
-              <TextContent>
-                <Text component={TextVariants.p}>Following validations need to pass:</Text>
-                <TextList>
-                  {failedValidations.map((validation) => (
-                    <TextListItem key={validation.id}>{validation.message}</TextListItem>
-                  ))}
-                </TextList>
-              </TextContent>
+              <Flex spaceItems={{ default: 'spaceItemsSm' }} direction={{ default: 'column' }}>
+                <FlexItem>The following requirements must be met:</FlexItem>
+                <FlexItem>
+                  <List>
+                    {failedValidations.map((validation) => (
+                      <ListItem key={validation.id}>{validation.message}</ListItem>
+                    ))}
+                  </List>
+                </FlexItem>
+              </Flex>
             </Alert>
           )}
         </AlertGroup>
--- a/src/components/clusterConfiguration/ClusterValidationSection.tsx
+++ b/src/components/clusterConfiguration/ClusterValidationSection.tsx
@@ -11,6 +11,9 @@ import {
   SplitItem,
   TextList,
   TextListItem,
+  TextContent,
+  Text,
+  TextVariants,
 } from '@patternfly/react-core';
 import { TimesIcon } from '@patternfly/react-icons';
 import { Cluster } from '../../api/types';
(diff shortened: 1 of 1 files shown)


mareklibra:
@andybraren , I do not like standard Patternfly's formatting without additional text.


The indentation works better with extra text on top of the list:


wdyt?

andybraren (commented):
Slight wording tweak for now then LGTM. I'm asking in the PatternFly Slack about whether there's a better way to add a bulleted list that looks more natural.

andybraren:
I was reminded about the magical [`Flex` layout](https://www.patternfly.org/v4/layouts/flex/react) and [`List` component](https://www.patternfly.org/v4/components/list/react). 🙂

This should work for us:

```
<div class="pf-c-alert__description">
  <div class="pf-l-flex pf-m-column pf-m-space-items-sm">
    <div class="pf-l-flex__item">
      <p>Success alert description. This should tell the user more information about the alert.</p>
    </div>
    <div class="pf-l-flex__item">
      <ul class="pf-c-list">
        <li>In fermentum leo eu lectus mollis, quis dictum mi aliquet.</li>
        <li>Morbi eu nulla lobortis, lobortis est in, fringilla felis.</li>
        <li>Aliquam nec felis in sapien venenatis viverra fermentum nec lectus.</li>
      </ul>
    </div>
  </div>
</div>
```

image

mareklibra:
@andybraren Thank you for passing the reminder of Flex :-)
This is the result:

nirfarkas:
Nits
1. @mareklibra I think the first bullet should also begin with uppercase ("No"). 
2. @andybraren Are you sure the first bullet is in proper English? I'm missing some punctuation...
3. @andybraren @mareklibra Didn't we decided to change "vips" to "virtual IPs" to be more coherent? 
4. A general comment: AFAIK, the best practice for error or warning messages is: <Error\warning message><Reason for the error\warning><How to solve> I'm missing this structure in bullets 1,3 & 4. 
5. We have two places where I can see that the cluster is not ready to be installed (the notification & the warning in the action bar. Do we need both?

mareklibra:
>@mareklibra I think the first bullet should also begin with uppercase ("No").
> @andybraren Are you sure the first bullet is in proper English? I'm missing some punctuation...
> @andybraren @mareklibra Didn't we decided to change "vips" to "virtual IPs" to be more coherent?

@nirfarkas I believe (most of) these are addressed here: https://github.com/openshift/assisted-service/pull/351
In addition, I can automatically convert the errors coming from backed to sentences (means to capitalize the first letter and add a dot at the end if missing) but this would require deeper analysis what text is coming to avoid issues in corner cases.

> A general comment: AFAIK, the best practice for error or warning messages is: <Error\warning message><Reason for the error\warning> I'm missing this structure in bullets 1,3 & 4.

These are validation errors. I think it makes no sense to add word `error` at every line. Please feel free to suggest rephrasing of the text above the bullets. Otherwise, the messages themselves should be tweaked within https://github.com/openshift/assisted-service 

> We have two places where I can see that the cluster is not ready to be installed (the notification & the warning in the action bar. Do we need both?

I am not sure I understand the issue. The alert on the screenshot is rendered in a response to clicking on the `The cluster is not ready to be installed` link in the toolbar.

nirfarkas:
> I think it makes no sense to add word error at every line.
 I didn't mean to have the word "Error" or "Warning" in every bullet. I meant that we need to provide some hint for the user where he can solve the issues he's having. For example "API virtual IP is undefined, you can provide it in the networking section".

mareklibra:
> I meant that we need to provide some hint for the user where he can solve the issues he's having. For example "API virtual IP is undefined, you can provide it in the networking section".

It makes totally sense. Please file an issue when you find it is missing somewhere. In the particular case of VIPs, the fields are already red with additional information.

andybraren:
I think/hope I fixed `1`, `2`, and `3` in that PR. Feel free to leave comments there @nirfarkas! 🙂

`4.` The tricky thing about adding "you can provide it in the networking section" to some of these messages is that they're also visible to CLI users interacting directly with the API. At the moment within OCM we won't have many/any of those, but when the installer is available on-prem for geekier users we likely will. As Marek said, the typical [red validation styling](https://www.patternfly.org/v4/components/form#invalid) for empty fields should be pretty prominent on the page; especially once we move to a step-based wizard and only a few fields are shown at a time.

If we think we need to enhance certain messages with GUI-specific help text, would we have a maintainable way of doing that @mareklibra?

`5.` The current behavior where that link acts like a button is indeed a little odd. I think it'd be worth exploring other ways of showing the current overall cluster status once we move to a wizard-based flow.

That reminds me... in the very early days we had a mockup that showed some kind of status or "achieved capability level" in the sidebar of the wizard, seen [here](https://docs.google.com/presentation/d/1UCHZ6XyoHZLKRWkKx0vw1Ga4FVGMsbOS0awLurPAe8s/edit#slide=id.g7f04c6a990_6_45). Maybe something worth thinking about again!

mareklibra:
> If we think we need to enhance certain messages with GUI-specific help text, would we have a maintainable way of doing that @mareklibra?

Majority of such messages can be found in https://github.com/mareklibra/facet-lib/blob/master/src/components/ui/formik/validationSchemas.ts , or just `git grep` :-)

I can extract the text to constants stored in a single file but so far this was not needed and would not bring much value. Anyway, it is an option.

jtomasek (approved):
Nice

PR #222, opened 2020-09-22
MGMT-2232 Retrieve host HW requirements from backend

mareklibra:


mareklibra:
hwReqs

src/api/utils.ts
```diff
@@ -51,3 +52,34 @@ export const stringToJSON = <T>(string: string | undefined): T | undefined => {
...
 export const removeProtocolFromURL = (url = '') => url.replace(/^(http|https):\/\//, '');
+
+type FetchApiEndpointParams<T> = {
+  apiCallback: () => AxiosPromise<T>;
+  onSuccess: (data: T) => void;
+  errorTitle: string;
+  onError?: (e: AxiosError<T>) => void;
+};
+
+export const fetchApiEndpoint = <T>({
```
mareklibra, line 63: I plan to refactor similar code at other places to use this function.

src/components/clusterConfiguration/BaremetalInventory.tsx
```diff
@@ -1,19 +1,30 @@
...
 
 interface BareMetalInventoryProps {
   cluster: Cluster;
 }
 
 const BaremetalInventory: React.FC<BareMetalInventoryProps> = ({ cluster }) => {
   const [isDiscoveryHintModalOpen, setDiscoveryHintModalOpen] = React.useState(false);
+  const [hostRequirements, setHostRequirements] = React.useState<HostRequirements | null>(null);
+
+  React.useEffect(() => {
```
mareklibra, line 21: I was thinking about placing this fetch at higher level in the hierarchy but so far there is no other use for it.

jtomasek (changes requested):
Can we maybe postpone introducing `fetchApiEndpoint` helper into a separate PR which would introduce it's usage on multiple places to see how it works out?

src/api/utils.ts
mareklibra: It's not a must to have it. The pattern repeats, so the idea was to abstract it. I understand your point, I am not big fan of over-abstracting either. So let's postpone such refactoring to a follow-up.

mareklibra: removed from here

src/components/clusterConfiguration/BaremetalInventory.tsx
```diff
@@ -1,19 +1,30 @@
...
+    fetchApiEndpoint<HostRequirements>({
+      apiCallback: getHostRequirements,
+      onSuccess: setHostRequirements,
+      errorTitle: 'Failed to retrieve minimal host requierements',
```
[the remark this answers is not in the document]
mareklibra: Just out of consistency :-), but both are valid: https://english.stackexchange.com/questions/277236/what-is-the-difference-between-minimum-and-minimal

jtomasek (approved):

PR #225, opened 2020-09-23
MGMT-1917 Make pull secret editable in OCM

mareklibra:
With this patch, the OCM user can edit pull-secret. Since this is expected to be a rare case, the field is not visible by default. Prior this patch, this field was hidden in OCM.

The pull-secret is not visible by default (is collapsed) but the user canexpand an area to edit it.
If pull-secret automatic fetch fails, the area is expanded by default (plus error message is shown).

Outside of the OCM, the pull-secret filed is always rendered.

### DIFF
--- a/src/components/clusters/PullSecret.tsx
+++ b/src/components/clusters/PullSecret.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { Checkbox, Popover } from '@patternfly/react-core';
 import { ExternalLinkAltIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
-import { ocmClient } from '../../api';
+import { ClusterCreateParams, ocmClient } from '../../api';
 import { CLUSTER_MANAGER_SITE_LINK, PULL_SECRET_INFO_LINK } from '../../config';
 import { TextAreaField } from '../ui';
+import { useFormikContext } from 'formik';
 
 export type PullSecretProps = {
   pullSecret?: string;
@@ -47,6 +48,8 @@ const PullSecretInfo = () => (
 const PullSecret: React.FC<PullSecretProps> = ({ pullSecret }) => {
   // Fetched pull secret will never change - see LoadingState in NewCluster
   const [isExpanded, setExpanded] = React.useState(!pullSecret);
+  const { setFieldValue } = useFormikContext<ClusterCreateParams>();
+
   const textArea = (
     <TextAreaField
       name="pullSecret"
@@ -63,6 +66,14 @@ const PullSecret: React.FC<PullSecretProps> = ({ pullSecret }) => {
   );
 
   if (ocmClient) {
+    const onCheckboxChange = () => {
+      if (isExpanded) {
+        // about to collapse, reset to original value
+        setFieldValue('pullSecret', pullSecret);
+      }
+      setExpanded(!isExpanded);
+    };
+
     return (
       <>
         <Checkbox
@@ -74,7 +85,7 @@ const PullSecret: React.FC<PullSecretProps> = ({ pullSecret }) => {
           }
           aria-label="customize pull secret"
           isChecked={isExpanded}
-          onChange={() => setExpanded(!isExpanded)}
+          onChange={onCheckboxChange}
         />
         {isExpanded && textArea}
       </>
--- a/src/components/clusters/PullSecret.tsx
+++ b/src/components/clusters/PullSecret.tsx
@@ -1,0 +1,87 @@
+import React from 'react';
+import { Checkbox, Popover } from '@patternfly/react-core';
+import { ExternalLinkAltIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import { ocmClient } from '../../api';
+import { CLUSTER_MANAGER_SITE_LINK, PULL_SECRET_INFO_LINK } from '../../config';
+import { TextAreaField } from '../ui';
+
+export type PullSecretProps = {
+  pullSecret?: string;
+};
+
+const ClusterManagerSiteLink = ({ text = 'OpenShiftCluster Manager' }: { text?: string }) => (
+  <a href={CLUSTER_MANAGER_SITE_LINK} target="_blank" rel="noopener noreferrer">
+    {text} <ExternalLinkAltIcon />
+  </a>
+);
+
+const pullSecretHelperText = ocmClient ? (
+  <>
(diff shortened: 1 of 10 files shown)


mareklibra:
---

mareklibra:
Cc @andybraren

andybraren:
@mareklibra I've added a new design iteration to [MGMT-1917](https://issues.redhat.com/browse/MGMT-1917), let me know what you think. Regardless, this may be a good time to add the `?` help text to describe what pull secrets are/do as well. The "Learn more..." link would direct to `https://cloud.redhat.com/openshift/install/pull-secret` and we can enhance the description on that page via [PD-645](https://issues.redhat.com/browse/PD-645).

mareklibra:
**Stand-alone:**


---

**In OCM:**


---



---

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <Checkbox
+          id="checkbox-pull-secret"
+          label={
+            <span onClick={(event) => event.preventDefault()}>
+              Customize pull secret <PullSecretInfo />
+            </span>
+          }
+          aria-label="customize pull secret"
+          isChecked={isExpanded}
+          onChange={() => setExpanded(!isExpanded)}
```
jtomasek, line 77: I think we need to reset the `pullSecret` value to initial value when user disables the checkbox.

jtomasek (changes requested):
`src/components/clusters/PullSecret.tsx` probably better fits in `src/components/clusterConfiguration` dir.

src/components/clusters/PullSecret.tsx
mareklibra: good catch, it's not just hiding but "Customize" now

nirfarkas:
Wording nit:
I think "Edit pull secret" is better here. @andybraren ?

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { ClusterCreateParams, ocmClient } from '../../api';
+import { CLUSTER_MANAGER_SITE_LINK, PULL_SECRET_INFO_LINK } from '../../config';
+import { TextAreaField } from '../ui';
+import { useFormikContext } from 'formik';
+
+export type PullSecretProps = {
+  pullSecret?: string;
+};
+
+const ClusterManagerSiteLink = ({ text = 'OpenShiftCluster Manager' }: { text?: string }) => (
```
andybraren, line 13:
```suggestion
const ClusterManagerSiteLink = ({ text = 'OpenShift Cluster Manager' }: { text?: string }) => (
```

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,98 @@
...
+      setExpanded(!isExpanded);
+    };
+
+    return (
+      <>
+        <Checkbox
+          id="checkbox-pull-secret"
+          label={
+            <span onClick={(event) => event.preventDefault()}>
+              Customize pull secret <PullSecretInfo />
```
andybraren, line 83:
```suggestion
              Edit pull secret <PullSecretInfo />
```

I think Nir is right - `Edit` is simpler and more common.

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,98 @@
...
+            </span>
+          }
+          aria-label="customize pull secret"
```
andybraren, line 86:
```suggestion
          aria-label="edit pull secret"
```

andybraren (commented):
Nit: I'm not sure how to edit the code for this, but the `?` popover for c.rh.c should have a link for "Learn more about pull secrets" instead of "in OpenShift Cluster Manager" since the user is already in OCM.

image

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,111 @@
...
+  </>
+);
+
+const PullSecretInfo = () => (
+  <Popover
+    position="top"
+    bodyContent={
+      <>
+        Pull secrets are used to download OpenShift Container Platform components and connect
+        clusters to a Red Hat account.
```
jtomasek, line 43:
```suggestion
        clusters to a Red Hat account.{' '}
```

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,111 @@
...
+        ) : (
+          <>
+            Pull secrets can be found in&nbsp;
+            <ClusterManagerSiteLink />
+          </>
+        )}
+      </>
+    }
+  >
+    <OutlinedQuestionCircleIcon />
```
jtomasek, line 55:
```suggestion
    <button onClick={(e) => e.preventDefault()} className="pf-c-form__group-label-help">
      <OutlinedQuestionCircleIcon noVerticalAlign />
    </button>
```
(according to https://www.patternfly.org/v4/components/form#basic)

We could eventually extract this into a specific `<FormFieldHelpPopover>` component.

src/components/clusters/PullSecret.tsx
```diff
@@ -0,0 +1,111 @@
...
+            aria-label="edit pull secret"
+            isChecked={isExpanded}
+            onChange={onCheckboxChange}
+            label="Edit pull secret"
+          />
+        }
+        labelIcon={<PullSecretInfo />}
+      >
+        {isExpanded && textArea}
+      </FormGroup>
```
jtomasek, line 104:
```suggestion
      <>
        <Checkbox
          id="checkbox-pull-secret"
          className="pf-u-display-inline-flex"
          aria-label="edit pull secret"
          isChecked={isExpanded}
          onChange={onCheckboxChange}
          label={
            <>
              Edit pull secret <PullSecretInfo />
            </>
          }
        />
        {isExpanded && textArea}
      </>
```

Standalone checkbox seems to work fine unless I am missing something.

jtomasek (changes requested):

src/components/clusters/PullSecret.tsx
jtomasek: This change ensures correct icon size and alignment.

src/components/clusters/PullSecret.tsx
mareklibra: The `FormGroup` was added in attempt to fix alignment and event propagation. It's a good idea to wrap the icon at lower level, so I am reverting this change here.

jtomasek (approved):

PR #226, opened 2020-09-23
MGMT-2258 Add error handling for failed DHCP VIP allocation

jtomasek:
image

### DIFF
--- a/src/components/clusterConfiguration/BasicNetworkFields.tsx
+++ b/src/components/clusterConfiguration/BasicNetworkFields.tsx
@@ -1,21 +1,39 @@
 import React from 'react';
 import { useFormikContext } from 'formik';
-import { Spinner } from '@patternfly/react-core';
-import { HostSubnets, ClusterConfigurationValues } from '../../types/clusters';
+import { Spinner, Alert, AlertVariant } from '@patternfly/react-core';
+import { HostSubnets, ClusterConfigurationValues, ValidationsInfo } from '../../types/clusters';
 import { InputField, SelectField, SwitchField } from '../ui/formik';
 import { Cluster } from '../../api/types';
-import StaticTextField from '../ui/StaticTextField';
+import { StaticField } from '../ui/StaticTextField';
 import { NO_SUBNET_SET } from './utils';
+import { stringToJSON } from '../../api/utils';
 
 type VipStaticValueProps = {
   vipName: string;
   cluster: Cluster;
+  validationErrorMessage?: string;
 };
 
-const VipStaticValue: React.FC<VipStaticValueProps> = ({ vipName, cluster }) => {
+const VipStaticValue: React.FC<VipStaticValueProps> = ({
+  vipName,
+  cluster,
+  validationErrorMessage,
+}) => {
   const { vipDhcpAllocation, machineNetworkCidr } = cluster;
+
   if (vipDhcpAllocation && cluster[vipName]) {
     return cluster[vipName];
+  }
+  if (vipDhcpAllocation && validationErrorMessage) {
+    return (
+      <Alert
+        variant={AlertVariant.danger}
+        title="The DHCP server failed to allocate the IP"
+        isInline
+      >
+        {validationErrorMessage}
+      </Alert>
+    );
   }
   if (vipDhcpAllocation && machineNetworkCidr) {
     return (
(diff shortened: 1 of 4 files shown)


src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -1,22 +1,39 @@
...
   const { vipDhcpAllocation, machineNetworkCidr } = cluster;
+
   if (vipDhcpAllocation && cluster[vipName]) {
     return cluster[vipName];
   }
+  if (vipDhcpAllocation && isValidationFailed) {
+    return (
+      <>
+        <ExclamationCircleIcon color={globalDangerColor200.value} />{' '}
+        <i>The DHCP server failed to allocate the IP</i>
```
mareklibra, line 33: nit: shouldn't this be a sentence?

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -42,6 +59,26 @@ const getVipHelperSuffix = (
...
+  const validationsInfo = stringToJSON<ValidationsInfo>(validationsInfoString) || {
+    hostsData: [],
+    network: [],
+    configuration: [],
+  };
+  const failedDhcpAllocationMessageStub = 'VIP IP allocation from DHCP server has been timed out';
+  return validationsInfo.network.reduce((lookup, validation) => {
+    if (['api-vip-defined', 'ingress-vip-defined'].includes(validation.id)) {
+      lookup[validation.id] =
+        validation.status === 'failure' && validation.message.match(failedDhcpAllocationMessageStub)
```
mareklibra, line 74: Having this, the only error we pass is the one matching `failedDhcpAllocationMessageStub`.
What if new errors are added in the future?

I think we should reverse the condition and black-list (read: use a list of skipped) messages which will not be shown in the UI. And report the rest.

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -42,6 +59,26 @@ const getVipHelperSuffix = (
...
+          ? validation.message
+          : undefined;
```
mareklibra, line 76: Considering previous comment, this will probably need to keep `lookup[validation.id]` which is either `undefined` or an already found message.

mareklibra (changes requested):

mareklibra:
Looking at the screenshots, the messages coming from backend will need to be tweaked. wdyt @nirfarkas  @andybraren ?

andybraren:
@mareklibra https://github.com/openshift/assisted-service/pull/351 should tweak these messages to:

> The API virtual IP is undefined; IP allocation from the DHCP server timed out.
> 
> The Ingress virtual IP is undefined; IP allocation from the DHCP server timed out.

This also seems like an opportunity to use a PatternFly inline alert, mocked up below.

But maybe we can do better with something more custom: what would the user's next step be to resolve this issue if it happens? Maybe we could provide some additional guidance or troubleshooting help either within the alert or with another documentation modal? Or maybe a simple retry button if that could sometimes work?

mgmt-2258-errors

src/components/clusterConfiguration/BasicNetworkFields.tsx
jtomasek: I tried doing this and it lead to displaying an error message when there was no dhcp allocation failure. In both cases we're matching the string which is error prone. I've decided to handle that specific message as Ori mentioned that the only way the allocation fails is the timeout (IOW the backend always reports timeout regardless of how the dhcp allocation fails because the allocation job timeouts and causes the failure)

jtomasek:
@andybraren @mareklibra I'll adjust the message stub to match the updated one.

Expected user interaction after the failure is probably to disable the DHCP allocation and enter the IPs manually. I am inclined to iterate on the message in a subsequent PRs and to get the general error handling in.

jtomasek:
Screenshot 2020-09-25 at 16 48 28

Used `Alert` to display the error, added action to disable DHCP in the alert, enhanced code to work with both old and new error messages.

I am not sure if the action in the alert is necessary since the toggle to disable DHCP is right above the alert.

andybraren:
I'm fine with removing the actions for now and merging to get this error handling in.

jtomasek:
Removed the alert action for now

src/components/clusterConfiguration/BasicNetworkFields.tsx
mareklibra: I still think we will reverse this search. At latest when fixing bug which will be opened right after adding another error message on backend. :-) Can be changed in a follow-up.

mareklibra (approved):

PR #228, opened 2020-09-23
MGMT-1538 Add support of the single-cluster flow

mareklibra:
TODO:
- [x] remove `~new` route by moving it under `SingleCluster`

### DIFF
--- a/src/components/Router.tsx
+++ b/src/components/Router.tsx
@@ -12,21 +12,16 @@ export const FacetRouter: React.FC = () => (
   </Provider>
 );
 
-export const Router: React.FC = ({ children }) =>
-  isSingleClusterMode() ? (
-    <Switch>
-      <Route path={`${routeBasePath}/clusters/~new`} component={NewClusterPage} />
-      <Route path={`${routeBasePath}/clusters/:clusterId`} component={ClusterPage} />
+export const Router: React.FC = ({ children }) => (
+  <Switch>
+    <Route path={`${routeBasePath}/clusters/~new`} component={NewClusterPage} />
+    <Route path={`${routeBasePath}/clusters/:clusterId`} component={ClusterPage} />
+    {isSingleClusterMode() ? (
       <Route path={`${routeBasePath}/clusters`} component={SingleCluster} />
-      {children}
-      <Redirect to={`${routeBasePath}/clusters/~new`} />
-    </Switch>
-  ) : (
-    <Switch>
-      <Route path={`${routeBasePath}/clusters/~new`} component={NewClusterPage} />
-      <Route path={`${routeBasePath}/clusters/:clusterId`} component={ClusterPage} />
+    ) : (
       <Route path={`${routeBasePath}/clusters`} component={Clusters} />
-      {children}
-      <Redirect to={`${routeBasePath}/clusters`} />
-    </Switch>
-  );
+    )}
+    {children}
+    <Redirect to={`${routeBasePath}/clusters`} />
+  </Switch>
+);
--- a/src/components/Router.tsx
+++ b/src/components/Router.tsx
@@ -12,16 +12,20 @@ export const FacetRouter: React.FC = () => (
   </Provider>
 );
 
-export const Router: React.FC = ({ children }) => (
-  <Switch>
-    <Route path={`${routeBasePath}/clusters/~new`} component={NewClusterPage} />
-    <Route path={`${routeBasePath}/clusters/:clusterId`} component={ClusterPage} />
-    {isSingleClusterMode() ? (
+export const Router: React.FC = ({ children }) =>
+  isSingleClusterMode() ? (
+    <Switch>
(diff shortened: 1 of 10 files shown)


src/components/Router.tsx
```diff
@@ -3,20 +3,30 @@ import { Provider } from 'react-redux';
...
-  </Switch>
-);
+export const Router: React.FC = ({ children }) =>
+  isSingleClusterMode() ? (
+    <Switch>
+      <Route path={`${routeBasePath}/clusters/~new`} component={NewClusterPage} />
+      <Route path={`${routeBasePath}/clusters/:clusterId`} component={ClusterPage} />
+      <Route path={`${routeBasePath}/clusters`} component={SingleCluster} />
+      {children}
+      <Redirect to={`${routeBasePath}/clusters/~new`} />
```
jtomasek, line 22: This should be the same as in the default mode to ensure that we run through the `SingleCluster` component to ensure that we don't create another new cluster in case when user accesses unknown path.
 
```suggestion
      <Redirect to={`${routeBasePath}/clusters`} />
```

jtomasek (changes requested):
There are buttons which point to `/clusters` path which should not be shown in case of single-cluster mode. We should be able to extract them into a component and handle the mode there as we're doing with breadcrumbs.

Comments inline...

src/components/Router.tsx
mareklibra: Right, good point. Since difference between these two branches in a single line only, I will collapse ...

src/components/SingleCluster.tsx
```diff
@@ -0,0 +1,61 @@
...
+      return handleApiError(e, () => setError('Failed to fetch cluster.'));
+    }
+  }, [setClusters]);
+
+  React.useEffect(() => {
+    fetchClusters();
+  }, [fetchClusters]);
+
+  React.useEffect(() => {
+    if (clusters) {
```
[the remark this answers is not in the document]
mareklibra: > clusters are always true because it is an array. 

Not in recent implementation. The `useState()` is called without param, so it can be `undefined`.
The `clusters === undefined` is farther used to show loading spinner.

> We don't need to use effect here, instead we could just render Redirect component (either to ~new or clusterId)

The effect fetches cluster list (with a single item) to get the `clusterId`. Otherwise redirects to `~new`. Mind the component is rendered for the `/clusters` path.
I can redirect using `<Redirect>` component, but this will make the logic more complicated compared to the recent use of `history`. Anyway, the fetching needs to be handled in an effect.

mareklibra:
> There are buttons which point to /clusters path which should not be shown in case of single-cluster mode.

So far I found `Back to all clusters` link. Any idea what else, please?

jtomasek (changes requested):
There is one more button at `ClusterPage.tsx` (cancel button displayed only in error state).

Since we have more of these buttons how about introducing `BackToClusters` button component which would be hidden in case of `isSingleClusterMode` so we don't need to explicitly set isHidden for all buttons and accessing isSingleClusterMode would be also reduced to less places in the code.

To handle rendering of different buttons (Button vs ToolbarButton) we could add `component` prop.

mareklibra:
It is a good idea with wrapping the feature into `BackToClusters` button. Anyway, when writing it to cover all recent uses, this abstraction leads to much harder-to-read code than just 3-4 single liners with `isHidden`. I vote for keeping the code simple.

src/components/Router.tsx
```diff
@@ -15,7 +16,11 @@ export const Router: React.FC = ({ children }) => (
   <Switch>
     <Route path={`${routeBasePath}/clusters/~new`} component={NewClusterPage} />
```
mareklibra, line 17: Let's move this inside `SingleCluster` to reduce chance of 2nd cluster creation.

src/components/clusters/NewClusterPage.tsx
```diff
@@ -135,7 +138,7 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ history, pullSecret = '
   );
 };
 
-const NewCluster: React.FC<RouteComponentProps> = (props) => {
+const NewCluster: React.FC<{}> = (props) => {
```
jtomasek, line 141:
```suggestion
const NewCluster: React.FC = (props) => {
```

src/components/clusters/NewClusterPage.tsx
```diff
@@ -168,7 +171,7 @@ const NewCluster: React.FC<RouteComponentProps> = (props) => {
   return <NewClusterForm pullSecret={pullSecret} {...props} />;
 };
 
-const NewClusterPage: React.FC<RouteComponentProps> = (props) => (
+const NewClusterPage: React.FC<{}> = (props) => (
```
jtomasek, line 174:
```suggestion
const NewClusterPage: React.FC = (props) => (
```

jtomasek (approved):
Few nits just in case you rebase.

PR #229, opened 2020-09-24
MGMT-2089 Show hyper-threading if available

mareklibra:
The user is provided with information whether hyper-threading is available on a host.


mareklibra:
I do not think we need a new field for it or to occupy the summary within a collapsed row.

So added the HT indication to the `CPU clock speed` section. Text `(hyper-threaded)` is rendered if HT is available only, otherwise skipped.
The idea is to list similar "selected details" here (a CPU can have multiple tens of flags but just a very few might be important for the user in our UI). So far just the HT is considered to be important.

andybraren:
For accessibility and clarity reasons, we may want to consider adjusting the combined field to something like:

> CPU cores and clock speed
> 4 cores (hyper-threaded) at 2,095 MHz

WDYT?

jtomasek (approved):

PR #231, opened 2020-09-25
MGMT-793 Detect host to be a virtual machine

mareklibra:

### DIFF
--- a/src/components/hosts/hardwareInfo.ts
+++ b/src/components/hosts/hardwareInfo.ts
@@ -64,7 +64,7 @@ export const getHostRowHardwareInfo = (inventory: Inventory): HostRowHardwareInf
   }
 
   return {
-    serialNumber: inventory.systemVendor?.serialNumber || DASH,
+    serialNumber: inventory.systemVendor?.serialNumber || 'Virtual machine',
     cores,
     cpuSpeed,
     memory,


src/components/hosts/hardwareInfo.ts
```diff
@@ -64,7 +64,7 @@ export const getHostRowHardwareInfo = (inventory: Inventory): HostRowHardwareInf
   }
 
   return {
-    serialNumber: inventory.systemVendor?.serialNumber || DASH,
+    serialNumber: inventory.systemVendor?.serialNumber || 'Virtual machine',
```
andybraren, line 67:
```suggestion
    serialNumber: inventory.systemVendor?.serialNumber || 'None (virtual machine)',
```

This might flow a little better: "What's the serial number? Oh, it's `None`, because it's a virtual machine"

Also, is not having a serial number a reliable way of detecting a virtual machine? Could there be hosts that we just can't detect a serial number for, but they're actually bare metal and not VM-based? 🤷‍♂️

mareklibra: @ori-amizur Can you please comment ^^^ ?

mareklibra:
Based on discussion, this flag will be provided by backend

mareklibra:
Closing without merge. Will be replaced by another PR.

PR #233, opened 2020-09-25
MGMT-1822 Make Cluster list filters persistent

mareklibra:
The state of filters on the Cluster list page stays persistent
unless the application is opened in a new tab.

The state survives page refresh (F5) or navigatino within the application.

### DIFF
--- a/src/components/clusters/ClustersTable.tsx
+++ b/src/components/clusters/ClustersTable.tsx
@@ -69,8 +69,12 @@ const ClustersTable: React.FC<ClustersTableProps> = ({ rows, deleteCluster }) =>
   React.useEffect(() => {
     const marshalled = window.sessionStorage.getItem(STORAGE_KEY_CLUSTERS_FILTER);
     if (marshalled) {
-      const parsed = JSON.parse(marshalled);
-      setFilters(parsed);
+      try {
+        const parsed = JSON.parse(marshalled);
+        setFilters(parsed);
+      } catch (e) {
+        console.info('Failed to restore clusters filter: ', e);
+      }
     }
   }, []);
 
--- a/src/components/clusters/ClustersTable.tsx
+++ b/src/components/clusters/ClustersTable.tsx
@@ -23,6 +23,8 @@ import ClustersFilterToolbar, { ClusterFiltersType } from './ClustersFilterToolb
 
 const rowKey = ({ rowData }: ExtraParamsType) => rowData?.props.id;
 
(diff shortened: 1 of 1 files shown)


src/components/clusters/ClustersTable.tsx
```diff
@@ -64,6 +66,18 @@ const ClustersTable: React.FC<ClustersTableProps> = ({ rows, deleteCluster }) =>
     status: [],
   });
 
+  React.useEffect(() => {
+    const marshalled = window.sessionStorage.getItem(STORAGE_KEY_CLUSTERS_FILTER);
```
jtomasek, line 70: why not using localStorage? It would allow persisting across tabs. There are no sensitive data stored in the filter so it should be fine, right?

src/components/clusters/ClustersTable.tsx
```diff
@@ -64,6 +66,18 @@ const ClustersTable: React.FC<ClustersTableProps> = ({ rows, deleteCluster }) =>
...
+    if (marshalled) {
+      const parsed = JSON.parse(marshalled);
+      setFilters(parsed);
```
jtomasek, line 73: Wrap this in try/catch to handle JSON.parse errors since the string in the storage can be manipulated so we're not sure we always get the string we expect.

jtomasek (changes requested):

src/components/clusters/ClustersTable.tsx
mareklibra: The decision was not driven by sensitivity of data but by reseting the state when reentering the application which is expected (I believe). In addition, I do not think we want to persist/share filters among tabs since the user can theoretically use different filters there.

If we want to make the filters persistent between reentering the application, I would leverage the GET URL params for that, so the user can "save" the search. But considering complexity of the recent application, I do not think this would bring significant value to the user.

src/components/clusters/ClustersTable.tsx
mareklibra: Plenty of other things can be hacked and cause subsequently application failure. But no problem with adding it.

mareklibra: Without explicit handling, we would learn about it via sentry

romfreiman:
Does it deal with sorting as well?

mareklibra:
@romfreiman , good point, sorting is recently not stored. This will require additional change in the flow. I will add it.

jtomasek (approved):

PR #235, opened 2020-09-29
MGMT-2280 Allow ssh public key comments

mareklibra:


mareklibra:
@oshercc Can you please ssh public key confirm format for https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L294 ?

jtomasek (approved):

PR #239, opened 2020-10-01
MGMT-2300 Show cluster.statusUpdatedAt on Cluster detail page

mareklibra:


mareklibra:
@andybraren The status-related details grow. Wdyt about organizing them using flex layout?
Please note, the bottom "Cluster details" part uses Grid layout recently.

andybraren:
Ah, `DescriptionList` is neat, but for now I think we can keep the current layout and not surface a separate `Status reached on` field. It seems to duplicate the date we integrate into the `Status` field when the install finishes/fails.

WRT the screenshot in MGMT-2300 that shows an error state without a timestamp, I would expect it to look similar to this:

MGMT-2300-failed-at

WDYT of the "View cluster events" link btw? 🙂

mareklibra:
> WDYT of the "View cluster events" link

We have already this link on the page in the right bottom corner throughout whole installation process, not just at the final state.

Recently just the `Installed` state comes with time information. What about showing time at every state? Sort of `Installing since XYZ`, `Finalizing since XYZ`, `Error occurred at XYZ`

andybraren:
True - we'll think more about the placement as part of [PD-649](https://issues.redhat.com/browse/PD-649).

@mareklibra Is that the full list of states? Time started and time installed/failed seem like the most important two to me, but if there are others and we think that information might be useful for debugging (without needing to dig into the full events list) maybe it's something we could consider as part of [PD-630](https://issues.redhat.com/browse/PD-630) @nirfarkas?

mareklibra:
All recent cluster statuses:
```
 'error'
 'preparing-for-installation'
 'installing'
 'finalizing'
 'installed'
```
Irrelevant for our discussion:
```
 'insufficient'
 'ready'
 'pending-for-input'
 'adding-hosts';
```

mareklibra:
Will be handled within redesign.

PR #245, opened 2020-10-05
Review new host-adding states

mareklibra:
Follow-up for added host's `added-to-existing-cluster` and cluster's `adding-hosts` statuses.

If the host is in `added-to-existing-cluster` status:
- status icon is `InProgressIcon`
- status detail text: `Host is added to existing cluster`

If the cluster is in `adding-hosts` status:
- icon: `InProgressIcon`
- Status label: `Adding hosts`

I do not see effect of these new states on our `canDo...()` UI checks atm.

Depends on:
- [x] #244

### DIFF
--- a/src/components/clusters/ClusterStatus.tsx
+++ b/src/components/clusters/ClusterStatus.tsx
@@ -25,18 +25,14 @@ const getStatusIcon = (status: Cluster['status']): React.ReactElement => {
       return <FileAltIcon />;
     case 'pending-for-input':
       return <PendingIcon />;
-    case 'ready':
-      return <CheckCircleIcon color={okColor.value} />;
-    case 'preparing-for-installation':
-      return <InProgressIcon />;
-    case 'installing':
-      return <InProgressIcon />;
     case 'error':
       return <ExclamationCircleIcon color={dangerColor.value} />;
+    case 'ready':
     case 'installed':
       return <CheckCircleIcon color={okColor.value} />;
+    case 'preparing-for-installation':
+    case 'installing':
     case 'finalizing':
-      return <InProgressIcon />;
     case 'adding-hosts':
       return <InProgressIcon />;
   }
--- a/src/components/hosts/HostStatus.tsx
+++ b/src/components/hosts/HostStatus.tsx
@@ -56,7 +56,7 @@ const getStatusIcon = (status: Host['status']): React.ReactElement => {
     case 'resetting-pending-user-action':
       return <WarningTriangleIcon color={warningColor.value} />;
     case 'added-to-existing-cluster':
-      return <InProgressIcon />; // TODO(mlibra): hotfix only, review in a follow-up
+      return <InProgressIcon />;
   }
 };
 
--- a/src/components/hosts/HostStatus.tsx
+++ b/src/components/hosts/HostStatus.tsx
@@ -31,30 +31,24 @@ const getStatusIcon = (status: Host['status']): React.ReactElement => {
       return <ConnectedIcon />;
     case 'pending-for-input':
       return <PendingIcon />;
-    case 'known':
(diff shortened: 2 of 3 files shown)


src/config/constants.ts
```diff
@@ -103,6 +105,7 @@ export const HOST_STATUS_DETAILS: { [key in Host['status']]: string } = {
   resetting: 'This host is resetting the installation.',
   'resetting-pending-user-action':
     'Host already booted from disk during previous installation. To finish resetting the installation please boot the host into Discovery ISO.',
+  'added-to-existing-cluster': 'Host is added to existing cluster',
```
jtomasek, line 108:
```suggestion
  'added-to-existing-cluster': 'This host is being added to existing cluster.',
```

- consistent with other status messages
- quoting Yevgeny: "it means that we have installed RCHOS with the ignition on the node and rebooted it. Once node is rebooted we have no way of monitoring its progress ( kubelet connection to OCP cluster etc'). So from our point of view we have finished with the node, but user must see a meessage that he must no monitor its progress till it is actually added to the OCP cluster"

IIUC we inform the user that the host is still in process of adding it to the cluster.

jtomasek (changes requested):

src/components/clusters/ClusterStatus.tsx
```diff
@@ -37,6 +37,8 @@ const getStatusIcon = (status: Cluster['status']): React.ReactElement => {
       return <CheckCircleIcon color={okColor.value} />;
     case 'finalizing':
       return <InProgressIcon />;
+    case 'adding-hosts':
```
jtomasek, line 40: All states with the same icon can be collapsed:
```
case 'preparing-for-installation':
case 'installing':
case 'finalizing':
case 'adding-hosts':
      return <InProgressIcon />;
```

src/components/hosts/HostStatus.tsx
```diff
@@ -55,6 +55,8 @@ const getStatusIcon = (status: Host['status']): React.ReactElement => {
       return <InProgressIcon />;
     case 'resetting-pending-user-action':
       return <WarningTriangleIcon color={warningColor.value} />;
+    case 'added-to-existing-cluster':
```
jtomasek, line 58: similar as with ClusterStatus

jtomasek (changes requested):

src/components/clusters/ClusterStatus.tsx
mareklibra: done

src/config/constants.ts
```diff
@@ -105,7 +105,7 @@ export const HOST_STATUS_DETAILS: { [key in Host['status']]: string } = {
   resetting: 'This host is resetting the installation.',
   'resetting-pending-user-action':
     'Host already booted from disk during previous installation. To finish resetting the installation please boot the host into Discovery ISO.',
-  'added-to-existing-cluster': 'Host is added to existing cluster', // TODO(mlibra): add this new state plus cluster's adding-hosts properly (in a follow up)
+  'added-to-existing-cluster': 'This host is being added to existing cluster.',
```
andybraren, line 108:
```suggestion
  'added-to-existing-cluster': 'This host is being added to an existing cluster.',
```

andybraren (commented):
Will any of these changes appear in the UI we have today once it merges? Or is this preparing for future stuff?

mareklibra:
> Will any of these changes appear in the UI we have today once it merges? Or is this preparing for future stuff?

I guess the question is related to c.rh.c . We even skipped last release there since backend is outdated atm. SO hard to say. We need to get in sync asap.
Regarding PSI or our dev envs, it can be done immediately.

jtomasek (approved):

PR #247, opened 2020-10-06
OCPBUGSM-18591 Host status alerts can be extended for user actions

mareklibra:
The first user action is the hostname change in case of 'hostname-unique' or 'hostname-valid' host state.

### DIFF
--- a/src/components/hosts/HostStatus.tsx
+++ b/src/components/hosts/HostStatus.tsx
@@ -15,7 +15,7 @@ import {
   BanIcon,
   PendingIcon,
 } from '@patternfly/react-icons';
-import { Host } from '../../api/types';
+import { Cluster, Host } from '../../api/types';
 import { ValidationsInfo } from '../../types/hosts';
 import HostProgress from './HostProgress';
 import { HOST_STATUS_LABELS, HOST_STATUS_DETAILS } from '../../config/constants';
@@ -23,7 +23,7 @@ import { getHumanizedDateTime } from '../ui/utils';
 import { toSentence } from '../ui/table/utils';
 import { getHostProgressStageNumber, getHostProgressStages } from './utils';
 import { stringToJSON } from '../../api/utils';
-import HostValidationGroups from './HostValidationGroups';
+import HostValidationGroups, { ValidationInfoActionProps } from './HostValidationGroups';
 
 const getStatusIcon = (status: Host['status']): React.ReactElement => {
   switch (status) {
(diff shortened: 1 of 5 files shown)


src/components/hosts/HostStatus.tsx
```diff
@@ -113,10 +116,11 @@ const HostStatus: React.FC<HostStatusProps> = ({ host }) => {
     <>
       <Popover
         headerContent={<div>{title}</div>}
-        bodyContent={getPopoverContent(host)}
+        bodyContent={getPopoverContent({ host, cluster, onValidationActionToggle })}
         footerContent={<small>Status updated at {getHumanizedDateTime(statusUpdatedAt)}</small>}
         minWidth="30rem"
         maxWidth="50rem"
+        hideOnOutsideClick={!keepOnOutsideClick}
```
jtomasek, line 123: I'd expect the popover to hide when the host editing modal opens. Keeping the popover open causes the popover to overlay the modal and having 2 overlays open is quite confusing

src/components/hosts/HostValidationGroups.tsx
```diff
@@ -7,13 +7,76 @@ import {
...
+  if (!validations.length) {
+    return null;
+  }
+  return (
+    <Alert title={title} variant={variant} isInline>
+      <ul>
+        {validations.map((v) => (
+          <li key={v.id}>
+            <strong>{HOST_VALIDATION_LABELS[v.id]}:</strong>&nbsp;{v.message}&nbsp;
+            <ValidationInfoAction validation={v} {...props} />
```
jtomasek, line 68: I think it would be better to use standard `Alert`s `actionLinks` prop to provide the actions for failing validations in the group. (https://www.patternfly.org/v4/components/alert#props)

jtomasek (changes requested):

src/components/hosts/HostStatus.tsx
mareklibra: In that case I will need to move the modal component up in the hierarchy which is breaking the structure, imo.

If we keep the modal component down in the hierarchy (as it is now), an user's click inside of the modal closes the popover and subsequently destroys the modal component.

Or we might do bigger refactoring of (all) modals we have, shift all of them to a single higher place and drive their state via functions from context or redux. If we decide to go this way, we should implement it in a follow-up.

jtomasek (approved):

PR #248, opened 2020-10-06
MGMT-1750 Make list of hosts not matching subnet readable

mareklibra:


mareklibra:
---

---

mareklibra:
The later commit moves the list into a popover:


---

andybraren (commented):
Small word tweak suggestion then lgtm, thanks Marek!

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -84,6 +96,56 @@ const getVipValidationsById = (
...
+        }
+        minWidth="30rem"
+        maxWidth="50rem"
+      >
+        <Button
+          variant={ButtonVariant.link}
+          id="form-input-hostSubnet-field-helper-view-excluded"
+          isInline
+        >
+          {`View ${excludedHosts.length} not matching host${excludedHosts.length > 1 ? 's' : ''}`}
```
andybraren, line 142:
```suggestion
          {`View ${excludedHosts.length} affected host${excludedHosts.length > 1 ? 's' : ''}`}
```

`affected` further emphasizes that these hosts that will be impacted in some way, which is probably good (and is a little less backend-ey).

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -84,6 +96,56 @@ const getVipValidationsById = (
   }, {});
 };
 
+const SubnetHelperText: React.FC<{ matchingSubnet: HostSubnet; cluster: Cluster }> = ({
+  matchingSubnet,
+  cluster,
+}) => {
+  const excludedHosts =
+    cluster?.hosts?.filter(
```
jtomasek, line 104:
```suggestion
    cluster.hosts?.filter(
```

src/components/clusterConfiguration/BasicNetworkFields.tsx
```diff
@@ -84,6 +96,56 @@ const getVipValidationsById = (
...
+              )
+              .map((host) => (
+                <li key={host.id}>{host.requestedHostname}</li>
+              ))}
+          </ul>
+        }
+        minWidth="30rem"
+        maxWidth="50rem"
+      >
+        <Button
```
jtomasek, line 137: It would be better to use `actionLinks` prop + `AlertActionLink` instead of Button on the `Alert` but I am not sure if it works well with `Popover`.

jtomasek (changes requested):

src/components/clusterConfiguration/BasicNetworkFields.tsx
mareklibra: works fine, done

jtomasek (approved):

PR #257, opened 2020-10-13
MGMT-2423 Display cluster/host UUIDs

jtomasek:
Display Cluster UUID in cluster properties
Display Host UUID in host expanded detail

Screenshot from 2020-10-13 11-04-25
Screenshot from 2020-10-13 11-05-24

### DIFF
--- a/src/components/clusterDetail/ClusterProperties.tsx
+++ b/src/components/clusterDetail/ClusterProperties.tsx
@@ -14,7 +14,7 @@ const ClusterProperties: React.FC<ClusterPropertiesProps> = ({ cluster }) => (
         <Text component="h2">Cluster Details</Text>
       </TextContent>
     </GridItem>
-    <GridItem md={6} lg={6} xl={6}>
+    <GridItem md={6}>
       <DetailList>
         <DetailItem title="OpenShift version" value={cluster.openshiftVersion} />
         <DetailItem title="Base DNS domain" value={cluster.baseDnsDomain} />
@@ -22,8 +22,9 @@ const ClusterProperties: React.FC<ClusterPropertiesProps> = ({ cluster }) => (
         <DetailItem title="Ingress virtual IP" value={cluster.ingressVip} />
       </DetailList>
     </GridItem>
-    <GridItem md={6} lg={6} xl={6}>
+    <GridItem md={6}>
       <DetailList>
+        <DetailItem title="UUID" value={cluster.id} />
         <DetailItem title="Cluster network CIDR" value={cluster.clusterNetworkCidr} />
         <DetailItem title="Cluster network host prefix" value={cluster.clusterNetworkHostPrefix} />
         <DetailItem title="Service network CIDR" value={cluster.serviceNetworkCidr} />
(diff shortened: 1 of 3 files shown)


nirfarkas:
Nit:
Reviewing both screenshots I wonder what is the convention we should use in the “Details” section:
In the cluster details, we show "Cluster network CIDR" & "Cluster network Host prefix" but in the host details, we're not adding "host" to any of the headers...
So, should it be "Cluster UUID" just "UUID"?

jtomasek:
"Cluster network CIDR" and "Cluster network Host prefix" are specific OpenShift terms that's why it includes words 'Cluster and Host'. (for example compare 'Cluster network CIDR' vs 'Service network CIDR')

src/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -14,21 +14,26 @@ const ClusterProperties: React.FC<ClusterPropertiesProps> = ({ cluster }) => (
...
+    <GridItem md={6} lg={4} xl={4}>
       <DetailList>
         <DetailItem title="Cluster network CIDR" value={cluster.clusterNetworkCidr} />
         <DetailItem title="Cluster network host prefix" value={cluster.clusterNetworkHostPrefix} />
         <DetailItem title="Service network CIDR" value={cluster.serviceNetworkCidr} />
       </DetailList>
     </GridItem>
+    <GridItem md={6} lg={4} xl={4}>
+      <DetailList>
+        <DetailItem title="UUID" value={cluster.id} />
```
mareklibra, line 34: We need to make sure, the user understands that this is just the `Assisted Installed UUID` and not the OCP ID which the user will use any farther.
There is a plan/wish to expose this second (probably more important for the user) ID too, but it's so far missing in the API.

@nirfarkas @andybraren

mareklibra (changes requested):

src/components/clusterDetail/ClusterProperties.tsx
jtomasek: I've discussed this briefly with @andybraren yesterday. I am calling the ID just UUID to imply that it is the cluster resource ID and when it is available, we could call the OCP ID 'OpenShift Cluster ID'. This should provide enough differentiation, what do you think?

src/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -14,21 +14,26 @@ const ClusterProperties: React.FC<ClusterPropertiesProps> = ({ cluster }) => (
...
     </GridItem>
-    <GridItem md={6} lg={6} xl={6}>
+    <GridItem md={6} lg={4} xl={4}>
       <DetailList>
         <DetailItem title="Cluster network CIDR" value={cluster.clusterNetworkCidr} />
         <DetailItem title="Cluster network host prefix" value={cluster.clusterNetworkHostPrefix} />
         <DetailItem title="Service network CIDR" value={cluster.serviceNetworkCidr} />
       </DetailList>
     </GridItem>
+    <GridItem md={6} lg={4} xl={4}>
```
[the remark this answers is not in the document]
jtomasek: done

src/components/clusterDetail/ClusterProperties.tsx
mareklibra: We can call those who understand the difference "Advanced Assisted Installer" users.

mareklibra (approved):

PR #258, opened 2020-10-13
Fix Developer Preview banner alignment

jtomasek:
image


nirfarkas:
We want to move the "Developer Preview" to the left, right? (:

jtomasek:
Why? Putting it into breadcrumbs makes it always visible as it never scrolls off the screen. Aligning it to the right also saves the vertical space.

mareklibra (approved):

nirfarkas:
I meant something like this:
image
I think that visually it's not balanced in the far right.
Also here it's off:
image

I presume that it's not that important that it should be always visible (it's not an interactive element that needs to be available at all times like "Help" or similar.  WDYT? Makes sense to you?

PR #265, opened 2020-10-15
MGMT-1576, MGMT-2228 Update the Discovery troubleshooting modal

jtomasek:
- Align to docs team recommendations
- Update code sections to resemble latest state

### DIFF
--- a/src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
+++ b/src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
@@ -38,7 +38,7 @@ export const DiscoveryTroubleshootingModal: React.FC<DiscoveryTroubleshootingMod
   ]);
   return (
     <Modal
-      title="Host Discovery Troubleshooting"
+      title="Troubleshooting Host Discovery Issues"
       isOpen={isOpen}
       actions={[
         <Button key="close" variant={ButtonVariant.primary} onClick={onClose}>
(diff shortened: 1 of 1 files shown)


src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
```diff
@@ -50,78 +50,66 @@ export const DiscoveryTroubleshootingModal: React.FC<DiscoveryTroubleshootingMod
...
-        <Text component={TextVariants.p}>Execute</Text>
         <PrismCode code={`ps -ef | grep agent`} />
-        <Text component={TextVariants.p}>
-          and see whether the agent runs with correct parameters. The result should look similar to
-          this:
-        </Text>
+        <Text component={TextVariants.p}>The output displays the following:</Text>
         <PrismCode
-          code={`root        1342       1  0 09:56 ?        00:00:00 /usr/local/bin/agent --host 192.168.39.162 --port 30956 --cluster-id a8142e14-8bfe-46bd-bfbd-4bd1a126892b core        2063    2003  0 10:41 pts/0    00:00:00 grep --color=auto agent`}
+          code={`root        1775       1  0 11:19 ?        00:00:00 /usr/local/bin/agent --url http://10.19.130.1:6000 --cluster-id f1063d85-3099-4d7b-ade3-269142184044 --agent-version quay.io/ocpmetal/assisted-installer-agent:latest --insecure=true
```
ohadlevy, line 77: should be https:// and probably api.openshift.com or something?

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
```diff
@@ -50,78 +50,66 @@ export const DiscoveryTroubleshootingModal: React.FC<DiscoveryTroubleshootingMod
...
         </Text>
-        <Text component={TextVariants.h2}>Check agent logs</Text>
-        <PrismCode code={`[core@vm-11-47 ~]$ less /var/log/agent.log`} />
         <Text component={TextVariants.h2}>Check assisted-installer logs</Text>
         <PrismCode
-          code={`[core@vm-11-47 ~]$ sudo su
-[core@vm-11-47 ~]$ podman ps -a | grep assisted-installer
-[core@vm-11-47 ~]$ podman logs <container id>`}
+          code={`sudo su
+podman ps -a | grep assisted-installer
```
ohadlevy, line 108: just wondering if we should be consistent (above we use ps -ef :)

ohadlevy (commented):
really minor nitpicking :)

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
jtomasek: Hmm, it is `ps -ef` vs `podman ps -a` which are different commands.

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
jtomasek: Fixed, thanks

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
```diff
@@ -50,78 +50,66 @@ export const DiscoveryTroubleshootingModal: React.FC<DiscoveryTroubleshootingMod
...
-        <PrismCode code={`ssh core@<machine-ip>`} />
         <Text component={TextVariants.p}>
-          The SSH key provided when generating the discovery ISO is used to authenticate the user.
-          It should not be required to provide the password.
+          Verify that you can access your host machine using SSH, or a console such as BMC or
+          virtual machine console. In the CLI, enter the following command:
+        </Text>
+        <PrismCode code="ssh core@<machine-ip>" />
+        <Text component={TextVariants.p}>
+          NOTE: Authentication is provided by the discovery ISO, therefore when you access your host
```
mareklibra, line 68: This works just in case that default private key on the machine/session of current user matches the public key provided when generating the ISO.
I would mention the `-i` param here. Or at least note about keys (which one is actually used).

olivelakra: What do you suggest?

mareklibra: To extend the hint to cover even the case when the default key is not matching the key provided in the ISO discovery modal.

If the existing "happy-path" fails, the user needs to use `ssh -i [MATCHING_PRIVATE_KEY]` to access the host.

Reasons for an issue include:
- different machine
- different user
- multiple keys used by the user
- a key previously generated just for the cluster installation

mareklibra: @olivelakra Can you please have a look ^^^ ?

jtomasek:
This needs to be also updated with information captured here https://issues.redhat.com/browse/MGMT-2784

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
olivelakra: So we can replace `ssh core@<machine-ip>` by ssh -i [MATCHING_PRIVATE_KEY] which is more generic.

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
```diff
@@ -50,78 +50,66 @@ export const DiscoveryTroubleshootingModal: React.FC<DiscoveryTroubleshootingMod
...
+        <Text component={TextVariants.p}>
+          Note: Boot time depends on several factors such as your hardware and network
+          configuration, and if you are booting from an ISO.
+        </Text>
+        <Text component={TextVariants.h2}>Verify if DHCP is enabled</Text>
         <Text component={TextVariants.h2}>SSH into your machine</Text>
-        <PrismCode code={`ssh core@<machine-ip>`} />
         <Text component={TextVariants.p}>
-          The SSH key provided when generating the discovery ISO is used to authenticate the user.
-          It should not be required to provide the password.
```
olivelakra, line 63: replace "It should not be required to provide the password." by "It is not mandatory to provide the password."

ohadlevy: I think you actually can't use a password to login in this case, only via ssh key.

jtomasek:
Screenshot 2020-12-16 at 13 43 25
Screenshot 2020-12-16 at 13 43 38

mareklibra (dismissed):

src/components/clusterConfiguration/DiscoveryTroubleshootingModal.tsx
```diff
@@ -50,78 +50,72 @@ export const DiscoveryTroubleshootingModal: React.FC<DiscoveryTroubleshootingMod
...
-          code={`[core@vm-11-47 ~]$ sudo su
-[core@vm-11-47 ~]$ podman ps -a | grep assisted-installer
-[core@vm-11-47 ~]$ podman logs <container id>`}
+          code={`sudo su
+podman ps -a | grep assisted-installer
+podman logs <container id>`}
         />
         <Text component={TextVariants.h2}>Check bootkube logs</Text>
-        <PrismCode code={`[core@vm-11-47 ~]$ journalctl -u bootkube`} />
+        <PrismCode code={`journalctl -u bootkube`} />
```
irosenzw, line 118: add `sudo` for consistency with the other `journalctl` commands

jtomasek: done

irosenzw (approved):

PR #270, opened 2020-10-21
OCPBUGSM-17787 Align cancel cluster condition to backend state machine

jtomasek:


src/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -30,35 +30,24 @@ import { AlertsContext } from '../AlertsContextProvider';
...
+    'pending-user-action',
+    'installing',
+    'installing-in-progress',
+    'installed',
+    'error',
+  ];
+  const { hosts = [] } = cluster;
+  return (
+    allowedClusterStates.includes(cluster.status) &&
+    hosts.find((host) => !allowedHostStates.includes(host.status))
```
jtomasek, line 49: According to this discussion it seems that it is not even necessary to check the host status https://coreos.slack.com/archives/CUPJTHQ5P/p1603284140247300?thread_ts=1603278706.226100&cid=CUPJTHQ5P

src/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -30,35 +30,25 @@ import { AlertsContext } from '../AlertsContextProvider';
...
-  }
-
-  return true;
+  const allowedClusterStates = [
+    'preparing-for-installation',
+    'installing',
+    'installing-in-progress',
+    'finalizing',
+  ];
+  const allowedHostStates = [
```
filanov, line 39: What about `resetting`?
I think that `pending-user-action` is `installing-pending-user-action`

jtomasek: `resetting` is done after the installation is cancelled so I don't think it makes sense to add it as a state from which we can cancel.

fixed to 'installing-pending-user-action' and added type to the arrays to avoid errors such as this.

RazForBiocatch (approved):

RazRegev (approved):

mareklibra (approved):

PR #271, opened 2020-10-22
MGMT-2593 Fix redirect for non-existing cluster

mareklibra:


src/components/clusters/ClusterPage.tsx
```diff
@@ -108,7 +108,7 @@ const ClusterPage: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
 
   if (uiState === ResourceUIState.ERROR) {
     if (errorDetail?.code === '404') {
-      return <Redirect to="/clusters" />;
+      return <Redirect to={`${routeBasePath}/clusters`} />;
```
mareklibra, line 111: This fix raises question if we want to redirect to the assisted-installer list and so skipping the "standard" 404 page of the OCM. Wdyt @jtomasek ?

jtomasek: I think this is a good fix for now and we can revisit when the cluster lists get merged

jtomasek (approved):

PR #272, opened 2020-10-22
Replace 'switch' with 'checkbox'

irosenzw:


jtomasek (changes requested):
I think we can keep the unused SwitchField component in the codebase.

irosenzw:
OK, I don't mind keeping it.

jtomasek (approved):

PR #273, opened 2020-10-22
OCPBUGSM-17474 PullSecret is expanded to all available height on New Cluster page

mareklibra:


jtomasek (commented):
This solution is not flexible as the calculated height will need to get adjusted whenever we do a change to the form or text in the page.

Alternatively we could make the textarea fit it's content but PatternFly's `TextArea` does not pass ref to textarea it renders so we'd need to fix this on PF side first.

mareklibra:
Agree and we should fix it properly this way. Anyway, it will take time, a lot of small steps need to be done. Can we workaround the issue till a proper fix is available?

nirfarkas:
Any chance for a screenshot? (: Thanks!

jtomasek (approved):

PR #274, opened 2020-10-22
Bug 1890504: Add select-all and clear-all checkbox for cluster events

irosenzw:
Add a single checkbox with both functionalities - clear all and select all

Screencast from 10-22-2020 07_01_31 PM

### DIFF
--- a/src/components/ui/ClusterEventsToolbar.tsx
+++ b/src/components/ui/ClusterEventsToolbar.tsx
@@ -31,6 +31,7 @@ export type ClusterEventsFiltersType = {
   severity: Event['severity'][];
   clusterLevel: boolean;
   orphanedHosts: boolean;
+  selectAll: boolean;
 };
 
 type ClustersListToolbarProps = {
@@ -46,7 +47,7 @@ const Placeholder = ({ text }: { text: string }) => (
   </>
 );
 
-const NO_HOSTS = 'deselect-all-hosts-action';
+const SELECT_ALL = 'select-all-actions';
 const CLUSTER_LEVEL = 'cluster-level-action';
 const ORPHANS = 'deleted-hosts-action';
 
@@ -69,6 +70,7 @@ export const getInitialClusterEventsFilters = (cluster: Cluster): ClusterEventsF
   severity: [],
   clusterLevel: true,
   orphanedHosts: true,
+  selectAll: true,
 });
 
 const getEventsCount = (severity: Event['severity'], events: Event[]) =>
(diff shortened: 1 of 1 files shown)


src/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -104,13 +124,18 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
...
     switch (value) {
-      case NO_HOSTS:
-        setFilters({
-          ...filters,
-          hosts: [], // clear all selection whenever clicked
-          clusterLevel: false,
-          orphanedHosts: false,
-        });
+      case SELECT_ALL:
+        if (filters.selectAll) {
```
mareklibra, line 128: nit: Can we use `isChecked` instead? And so flip these if-else blocks.

src/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -104,13 +124,18 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
...
+          setFilters({
+            ...filters,
+            clusterLevel: false,
+            orphanedHosts: false,
+            selectAll: false,
+            hosts: [],
+          });
+        } else {
+          setFilters(getInitialClusterEventsFilters(cluster));
```
mareklibra, line 137: With this, the `Severity` and `Text search` are emptied when `Select all hosts` is checked. We should just mark all hosts here and keep other subfilters untouched.

src/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -83,6 +85,24 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
...
+      setFilters({ ...filters, selectAll: false });
+    } else if (
+      filters.orphanedHosts &&
+      filters.clusterLevel &&
+      filters.hosts.length === cluster.hosts?.length &&
+      !filters.selectAll
+    ) {
+      setFilters({ ...filters, selectAll: true });
+    }
+  });
```
mareklibra, line 104: Having this effect like this, it is called with *every* render cycle, which we can avoid by proper setting within `onHostSelect()` flows.

mareklibra (changes requested):

src/components/ui/ClusterEventsToolbar.tsx
irosenzw: done.

src/components/ui/ClusterEventsToolbar.tsx
irosenzw: done.

mareklibra (approved):

PR #276, opened 2020-10-26
MGMT-2379: Advanced network configuration - use checkbox

irosenzw:
- Replace radio buttons with checkbox.
 - Remember adv settings modifications when switching
   back to 'Basic'.

### DIFF
--- a/src/components/clusterConfiguration/NetworkConfiguration.tsx
+++ b/src/components/clusterConfiguration/NetworkConfiguration.tsx
@@ -1,11 +1,13 @@
 import React from 'react';
 import { useFormikContext } from 'formik';
-import { TextContent, Radio, Text, FormGroup } from '@patternfly/react-core';
+import { TextContent, Text, Checkbox } from '@patternfly/react-core';
 import BasicNetworkFields from './BasicNetworkFields';
 import AdvancedNetworkFields from './AdvancedNetworkFields';
 import { HostSubnets, ClusterConfigurationValues } from '../../types/clusters';
 import { InputField, CheckboxField, SelectField } from '../ui/formik';
 import { ManagedDomain, Cluster } from '../../api/types';
+import { CLUSTER_DEFAULT_NETWORK_SETTINGS } from '../../config/constants';
+import { isAdvConf } from './utils';
 
 type NetworkConfigurationProps = {
   cluster: Cluster;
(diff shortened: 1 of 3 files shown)


mareklibra:
It does not correctly handle initial state:
- click on Advanced networking (check it)
- change a value there (Cluster Network Host Prefix from 23 to 22)
- Save & validate
- reload page (F5)
- the checkbox is unchecked (should be checked and so advanced section rendered)

---

The defaults are not correctly set which is confusing for the user:
- once a cluster is created, the backend sets defaults to the 3 advanced networking properties, including the Host prefix to 23
- repeat the scenario above (means: the user changes the prefix to 22)
- check-uncheck the Advanced netowrking

This leads the user to a state with the advanced networking unchecked (means: section is hidden), so the user expects that system defaults will be used. But when checked one more time, the Host prefix contains 22 (set by user) and not the system default (23).

---

I think we should compare the received values with system defaults. Please file a backend issue to provide them via API. Untill this is implemented and bubbled to production, we need to hardcode system defaults in the UI as a workaround.

mareklibra (dismissed):

src/types/clusters.ts
```diff
@@ -17,6 +17,7 @@ export type ClusterConfigurationValues = ClusterUpdateParams & {
   hostSubnet: string;
   useRedHatDnsService: boolean;
   shareDiscoverySshKey: boolean;
+  useAdvancedNetworking?: boolean;
```
jtomasek, line 20:
```suggestion
  useAdvancedNetworking: boolean;
```

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -81,30 +82,14 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
...
-          isChecked={isAdvanced}
-          onChange={() => setAdvanced(true)}
-          label="Advanced"
-          description="Configure a custom networking type and CIDR ranges."
-          isLabelWrapped
-        />
-      </FormGroup>
       <BasicNetworkFields cluster={cluster} hostSubnets={hostSubnets} />
-      {isAdvanced && <AdvancedNetworkFields />}
+      <CheckboxField
```
jtomasek, line 86: It would be better to leave the checkbox outside of the Formik form values so it does not have an effect on 'validate and  save' button. Just render `Checkbox` component here and onChange do `setAdvanced` to toggle show/hide of the 'AdvancedNetworkConfiguration`. + execute `toBasic` if it got unchecked - just reset the values to defaults instead of initial values.

The initial value for `isAdvanced` can be calculated using `isAdvConf`

jtomasek (changes requested):

src/components/clusterConfiguration/NetworkConfiguration.tsx
irosenzw: It's a bit problematic to leave it outside of formik since then it won't remember the checkbox value if the page refreshes along with the changed values.

jtomasek: Unless I misunderstood, the checkbox value should be initialised like this:
```
const [isAdvanced, setAdvanced] = React.useState(isAdvConf(cluster));
```

irosenzw: done. please review.

jtomasek (approved):
Works great!

PR #284, opened 2020-11-03
Move Day 2 action to OCM cluster detail subtab

mareklibra:
OCM counterpart: https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/1748


mareklibra:
---

---

src/components/hosts/HostsCount.tsx
```diff
@@ -8,15 +8,15 @@ type HostsCountProps = {
   valueId?: string;
 };
 
-export const getEnabledHostsCount = (hosts: Host[] = []) =>
-  hosts.filter((h) => h.status != 'disabled').length;
+export const getEnabledHostsCount = (hosts?: Host[]) =>
+  (hosts || []).filter((h) => h.status != 'disabled').length;
 
 const HostsCount: React.FC<HostsCountProps> = ({
-  hosts = [],
```
mareklibra, line 15: it can be `null`

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
```diff
@@ -1,25 +1,50 @@
...
+  ref: React.RefObject<unknown>;
+};
+
+export const getOCMClusterDetailTab = (
+  tabs: ClusterDetailTabType[],
+  addBareMetalTabRef: React.RefObject<unknown>,
+  displayAddBareMetalHosts: boolean,
+): ClusterDetailTabType => ({
+  key: tabs.reduce((max, tab) => (tab.key > max ? tab.key : max), tabs[0].key) + 1,
+  title: 'Add Hosts',
```
andybraren, line 28:
```suggestion
  title: 'Bare Metal',
```

I know we want to highlight this new feature so that it gets used (especially until we have something in OCP) but `Add Hosts` is more action-ey than the other tab labels and doesn't quite feel right. Renaming this `Bare Metal` is a little better, but I still find myself thinking this page should eventually just be an action someday (like it would be in OCP).

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
```diff
@@ -29,23 +27,27 @@ const AddBareMetalHosts: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
...
-    </>
+        <AlertsSectionGroup />
+        <Toolbar id="cluster-toolbar">
+          <ToolbarButton
+            variant={ButtonVariant.primary}
+            name="install"
+            onClick={handleHostsInstall}
+            isDisabled={!isKnownHost(cluster)}
+          >
+            Install ready hosts
```
andybraren, line 46: This might be out of scope of this PR, but WDYT of at least the UI reflecting the `Known` status of hosts as either `Ready` or `Ready to install` instead? I left [a comment here](https://docs.google.com/document/d/1jxNMTlotmJ0GFZ1GUEQ3hfOOVLzMT6mvY8Ufo5RdErY/edit?disco=AAAAGlR2YNc) about this a while back and still feel like it would be clearer to the user that the hosts are ready/green to proceed. `Known` sounds like a backend state.

image

mareklibra: We can change it on UI in a follow-up. Consistently for both Day 1 and 2.

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
mareklibra: @andybraren , can you please very briefly elaborate what do you mean by turning this feature into an action? This PR is actually about moving it from a kebab-action to a subtab. Maybe we don't want this to be merged ... Or we can rework.

This "tiny" title change gives us option to much wider use case in the future, anyway.

nirfarkas: Maybe a middle ground can be adding a plus icon to the right of the "Add Hosts"?
It will make this "Tab" look & feel different (like an action):
image

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
mareklibra: https://github.com/mareklibra/facet-lib/pull/288

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
```diff
@@ -29,23 +27,27 @@ const AddBareMetalHosts: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
   };
 
   return (
-    <>
-      <ClusterBreadcrumbs clusterName={cluster.name} />
-      <PageSection variant={PageSectionVariants.light} isMain>
+    <Card>
+      <CardTitle>
+        <Title headingLevel="h2" size="lg" className="card-title">
+          Bare Metal Inventory
```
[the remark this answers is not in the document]
mareklibra: done

src/config/constants.ts
```diff
@@ -132,7 +132,8 @@ export const HOST_STATUS_DETAILS: { [key in Host['status']]: string } = {
   resetting: 'This host is resetting the installation.',
   'resetting-pending-user-action':
     'Host already booted from disk during previous installation. To finish resetting the installation please boot the host into Discovery ISO.',
-  'added-to-existing-cluster': 'This host has been added to an existing cluster.',
+  'added-to-existing-cluster':
+    "This Host has booted RHCOS image, please use the Nodes section of the cluster's console to monitor further progress of adding this host to the cluster.",
```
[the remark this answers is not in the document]
mareklibra: done

mareklibra:
Some of the wording comments are addressed now.
In addition, the `HostsCount` component is refactored to better cover both Day 1 and 2 scenarios.

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
andybraren: I just discovered that the Support tab has a pattern that's very similar to what we're trying to do in this PR. Notice that the tab name is a noun - `Support` - but it has a section for support cases with a button to create a new one:

image

So here's a revised mockup with a tweaked card title (`Bare Metal Discovery` like the [future wizard step](https://marvelapp.com/prototype/7ce7ib3/screen/73593972)) that seems to align us pretty well with OCM's pattern:

2-ocm-add-hosts

The Actions menu in the top-right corner is disabled in OCM for me, but I'd be interested to see if `Open support case` is an action in there. If it's in there, maybe we should keep "Add Bare Metal Hosts" in there as well and link users to this page/tab when clicked. If not, then maybe we can skip that for now and Nir & I can sync with OCM designers and follow up later.

What do you both think?

src/config/constants.ts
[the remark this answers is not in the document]
andybraren: Nice! 👏

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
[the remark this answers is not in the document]
andybraren: Changed my mind again, sorry 😞, see the comment/mockup below.

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
andybraren: Nice, thank you! 👍

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
mareklibra: These kebab actions lead to modal dialogs, as we had the intermediary modal in first version. Unless there is a pattern for it, I would not add this action leading the user to the subtab. Seems to me confusing considering context. In addition, we would need to refactor OCM codebase.

IIUC, the expected change is just the subtab title, right?

andybraren: Sounds good, no action then, just a tab titled `Bare Metal`. Sorry for the long thread to get to that conclusion. 🙂

andybraren: And the card re-titled Bare Metal Discovery*

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
```diff
@@ -29,23 +27,27 @@ const AddBareMetalHosts: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
   };
 
   return (
-    <>
-      <ClusterBreadcrumbs clusterName={cluster.name} />
-      <PageSection variant={PageSectionVariants.light} isMain>
+    <Card>
+      <CardTitle>
+        <Title headingLevel="h2" size="lg" className="card-title">
+          Add Bare Metal Hosts
```
andybraren, line 33:
```suggestion
          Bare Metal Discovery
```

Changing my previous change

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
```diff
@@ -1,14 +1,12 @@
-import { ButtonVariant, PageSectionVariants } from '@patternfly/react-core';
+import { ButtonVariant, Card, CardBody, CardTitle, Title, Toolbar } from '@patternfly/react-core';
 import React from 'react';
 import { useDispatch } from 'react-redux';
 import { Cluster, getErrorMessage, handleApiError, installHosts } from '../../api';
 import { addAlert } from '../../features/alerts/alertsSlice';
 import { updateCluster } from '../../features/clusters/currentClusterSlice';
-import ClusterToolbar from '../clusters/ClusterToolbar';
-import ClusterBreadcrumbs from '../clusters/ClusterBreadcrumbs';
 import { isKnownHost } from '../hosts/utils';
```
jtomasek, line 7: renaming this utility function to `hasKnownHosts` would greatly help code readability. I had to look at the function to know what exactly it does.

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
mareklibra: done

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
```diff
@@ -1,25 +1,50 @@
...
+  isVisible: boolean;
+}> = ({ cluster, isVisible }) => {
   const [error, setError] = React.useState<string>();
-  const [redirect, setRedirect] = React.useState<Cluster['id']>();
+  const [day2Cluster, setDay2Cluster] = React.useState<Cluster | null>();
   const pullSecret = usePullSecretFetch();
 
   React.useEffect(() => {
-    if (isOpen && pullSecret) {
+    if (cluster && pullSecret && day2Cluster === undefined && isVisible) {
```
jtomasek, line 44: The day2 cluster is not re-fetched when user navigates back to 'Bare Metal' tab. I think it would be good to always refresh data on navigating to the tab.

The polling seems not to stop when user navigates to another tab because the tab content component is not unmounted when switching tabs. This is probably ok.

src/components/ui/AlertsSection/index.tsx
```diff
@@ -1,29 +1,36 @@
...
 import { AlertGroup, AlertActionCloseButton, Alert } from '@patternfly/react-core';
+import PageSection from '../PageSection';
 import { AlertProps } from '../../../features/alerts/alertsSlice';
+import { AlertsContext } from '../../AlertsContextProvider';
 
 import './AlertsSection.css';
-import { AlertsContext } from '../../AlertsContextProvider';
 
-const AlertsSection: React.FC = () => {
+export const AlertsSectionGroup: React.FC<{ className?: string }> = ({ className }) => {
```
jtomasek, line 9:
```suggestion
export const Alerts: React.FC<{ className?: string }> = ({ className }) => {
```
Since this has nothing to do with the page section how about using this name?

jtomasek (changes requested):

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
```diff
@@ -1,25 +1,50 @@
...
+): ClusterDetailTabType => ({
+  key: tabs.reduce((max, tab) => (tab.key > max ? tab.key : max), tabs[0].key) + 1,
+  title: 'Bare Metal',
+  contentId: 'addBareMetalHostsContent',
+  id: 'addBareMetalHosts',
+  show: displayAddBareMetalHosts,
+  ref: addBareMetalTabRef,
+});
+
+export const OCMClusterDetailTabContent: React.FC<{
```
jtomasek, line 35: This component name is quite confusing when being used on OCM side. I'd prefer calling it 'AddBareMetalHostsClusterDetailTab`

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
```diff
@@ -1,25 +1,50 @@
...
+type ClusterDetailTabType = {
+  key: number;
+  title: string;
+  contentId: string;
+  id: string;
+  show: boolean;
+  ref: React.RefObject<unknown>;
+};
+
+export const getOCMClusterDetailTab = (
```
jtomasek, line 22: is this required to be in facet-lib? I feel like this is very specific to OCM tabs implementation so it should live there.

jtomasek (changes requested):

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
mareklibra: All of these components are OCM specific and will most probably not used anywhere else. Having it here, it is easier for us to update if needed. Moreover, it conforms the tab-component.

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
mareklibra: done

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
mareklibra: Good point, changing to `BareMetalHostsClusterDetailTab`

src/components/AddBareMetalHosts/OCMClusterDetailTabContent.tsx
mareklibra: In OCM, all tabs are always rendered and data fetched, just the `hidden` attribute is changing.

Since this approach does not scale, let's stop at least our polling when not needed.
I have added a new commit doing that.

jtomasek (approved):

PR #286, opened 2020-11-04
Tweak add worker host wording

andybraren:


src/components/AddBareMetalHosts/BareMetalInventoryAddHost.tsx
```diff
@@ -11,9 +11,8 @@ import {
 
 const HostRequirementsContent = ({ worker = {} }: { worker?: HostRequirementsType['worker'] }) => (
   <Text component="p">
-    Worker hosts are required to be equiped by at least {worker.cpuCores || 2} CPU cores,{' '}
```
mareklibra, line 14: Same wording is in `BaremetalInventory.tsx`, I think we should stay consistent.

andybraren: The wording in `BaremetalInventory.tsx` seems a little different already since it describes how 2+ workers with the minimum specs are recommended for cluster creation.

```
  <Text component="p">
    Three master hosts are required with at least {master.cpuCores || 4} CPU cores,{' '}
    {master.ramGib || 16} GB of RAM, and {master.diskSizeGb || 120} GB of filesystem storage each.
    Two or more additional worker hosts are recommended with at least {worker.cpuCores || 2} CPU
    cores, {worker.ramGib || 8} GB of RAM, and {worker.diskSizeGb || 120}
    GB of filesystem storage each.
  </Text>
```

mareklibra: Right, I was just pointing to the language style difference. But it's nitpicking.

mareklibra (approved):

PR #288, opened 2020-11-04
MGMT-2792 Rename host's Known state to "Ready to install"

mareklibra:
These changes should be accompanied by changes within backend messages: https://github.com/openshift/assisted-service/pull/643


andybraren:
Thanks! I think this will be a lot clearer for users. FYI @nirfarkas

jtomasek (approved):
We should

PR #302, opened 2020-11-12
OCPBUGSM-20067 Add View Cluster Events to Day 2

mareklibra:


nirfarkas:
Isn't the role suppose to be always "Worker"? Do we think users will add masters on day 2?

mareklibra:
Eagle eye, @nirfarkas  :-) Right, it is recently `Worker` only but I was told that we might want to add `Master` in the future. Anyway, this is what API recently allows and especially **returns**. Rewording `Automaitc` to `Worker` or even automatically changing it by UI would not be a good idea.

jtomasek (approved):

PR #317, opened 2020-11-19
Add components for downloading Assisted Installer Live ISO

jtomasek:


jkilzi:
@jtomasek is this still relevant?

PR #319, opened 2020-11-20
Update API

jtomasek:


src/api/swagger.yaml
```diff
@@ -82,6 +82,11 @@ paths:
           type: boolean
           required: false
           default: false
+        - in: query
+          name: openshift_cluster_id
```
mareklibra, line 86: Adding to my TODO: we should newly query Day 2 based on this

mareklibra (approved):

PR #320, opened 2020-11-20
MGMT-792 Render virtual machine indication

mareklibra:
In addition, CPU details are reordered - more important things first

### DIFF
--- a/src/components/hosts/HostRowDetail.tsx
+++ b/src/components/hosts/HostRowDetail.tsx
@@ -133,6 +133,13 @@ export const HostDetail: React.FC<HostDetailProps> = ({ inventory, host }) => {
   }
   bmcAddress = bmcAddress || DASH;
 
+  let virtualText = DASH;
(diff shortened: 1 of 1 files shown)


andybraren:
Thanks for the screenshot!

This is a small detail, but for the `Serial number` WDYT of `None (virtual machine)` for this scenario? A dash like `-` sort of looks like the serial wasn't detected, but in this case the field just isn't applicable.

Since the table columns go CPU/Memory/Disk I think I lean toward `Memory capacity` being below the CPU-related fields (even though it's arguably more important), but my opinion isn't a strong one. @nirfarkas @AdamJ curious what you think, or any other ideas. See the [old layout here](https://github.com/mareklibra/facet-lib/pull/231) for reference.

mareklibra:
>  WDYT of None (virtual machine) for this scenario
Makes sense. I will update together with other changes which might come up based on raised questions.

jtomasek (approved):

PR #323, opened 2020-11-22
Bug 1882281: Add vip validation

irosenzw:
Require user to enter API vip and Ingress vip
when DHCP allocation is not checked.

### DIFF
--- a/src/components/ui/formik/validationSchemas.ts
+++ b/src/components/ui/formik/validationSchemas.ts
@@ -3,6 +3,7 @@ import { ClusterConfigurationValues, HostSubnets } from '../../../types/clusters
 import { Host } from '../../../api/types';
 import { ProxyFieldsType } from '../../clusterConfiguration/types';
 import { trimSshPublicKey } from './utils';
+import { NO_SUBNET_SET } from '../../../config/constants';
 
 const CLUSTER_NAME_REGEX = /^[a-z]([-a-z0-9]*[a-z0-9])?$/;
 const SSH_PUBLIC_KEY_REGEX = /^(ssh-rsa|ssh-ed25519|ecdsa-[-a-z0-9]*) AAAA[0-9A-Za-z+/]+[=]{0,3}( .+)?$/;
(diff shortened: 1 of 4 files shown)


src/components/ui/formik/validationSchemas.ts
```diff
@@ -96,6 +97,18 @@ const requiredOnceSet = (initialValue?: string, message?: string) =>
     (value) => value || !initialValue,
   );
 
+const vipRequired = (hostSubnet: string) =>
+  Yup.string().test(
+    'vip-required-no-dhcp',
+    'Requird. Please provide an IP address',
```
jtomasek, line 103: Typo
```suggestion
    'Required. Please provide an IP address',
```

src/components/ui/formik/validationSchemas.ts
```diff
@@ -105,7 +118,8 @@ export const vipValidationSchema = (
     is: (value) => !value,
     then: requiredOnceSet(initialValue)
       .concat(vipRangeValidationSchema(hostSubnets, values))
-      .concat(vipUniqueValidationSchema(hostSubnets, values)),
+      .concat(vipUniqueValidationSchema(hostSubnets, values))
+      .concat(vipRequired(values.hostSubnet)),
```
jtomasek, line 122: `when` can be used instead of combining `concat`, `test` and using function to pass the hostSubnet value.
```suggestion
      .when('hostSubnet', {
        is: (value: string) => value !== NO_SUBNET_SET,
        then: Yup.string().required('Required. Please provide an IP address.'),
      }),
```

jtomasek (changes requested):

src/components/ui/formik/validationSchemas.ts
irosenzw: done.

src/components/ui/formik/validationSchemas.ts
irosenzw: done.

jtomasek (approved):

PR #325, opened 2020-11-23
MGMT-2316: Provide DNS entries instructions

irosenzw:
Screencast from 11-29-2020 03_06_58 PM

### DIFF
--- a/src/components/clusterDetail/ConsoleModal.tsx
+++ b/src/components/clusterDetail/ConsoleModal.tsx
@@ -3,16 +3,15 @@ import {
   Modal,
   Button,
   ButtonVariant,
-  TextContent,
   Text,
-  TextList,
-  TextListItem,
   ModalVariant,
+  ExpandableSection,
 } from '@patternfly/react-core';
-import { Cluster } from '../../api/types';
-import { removeProtocolFromURL } from '../../api/utils';
+import { Cluster, Host, Interface, Inventory } from '../../api/types';
+import { removeProtocolFromURL, stringToJSON } from '../../api/utils';
 import { ToolbarButton } from '../ui/Toolbar';
 import { InfoCircleIcon } from '@patternfly/react-icons';
+import PrismCode from '../ui/PrismCode';
 
 type WebConsoleHintProps = {
   cluster: Cluster;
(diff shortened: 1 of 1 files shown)


nirfarkas:
image

Can we tweak the expandables behavior so when we click on one of them the other one collapses?
image

ohadlevy:
@nirfarkas I don't understand, didnt we agree that in both cases we would not want to see it at the same time on the screen?

nirfarkas:
@ohadlevy AFAIR we've said that we want to make it clear that the user can choose one solution from the two we're offering. I think this design is providing the information in a good manner... 
I've brought this up and shared it with other UX designers and all suggested using this UX rather than tabs because the tab component is for showing different content about something and not for providing two ways to fix a problem. We've also explored windows and Mac help docs and discovered that this is the UX they chose for similar situations.

ohadlevy:
I'm less worried on tabs vs not, but from the fact that both are visible at the same time? should clicking on option 2 close option 1 ?

irosenzw:
@ohadlevy it can be done of course. I just concerned about the UX of such a behavior. isn't it a bit annoying for the user ?

irosenzw:
modal first view: 
image

modal view when click on Option 2 ( or on Option 1 ) :
image

nirfarkas:
Thanks, Ido! @ohadlevy I think that the text together with the single item expansion does the job of communicating it's either one of the solutions and not both.

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -140,8 +224,8 @@ export const ConsoleModal: React.FC<ConsoleModalProps> = ({
     >
       Launch OpenShift Console
     </Button>,
-    <Button variant={ButtonVariant.secondary} onClick={() => closeModal()} key="cancel">
-      Cancel
+    <Button variant={ButtonVariant.secondary} onClick={() => closeModal()} key="close">
+      close
```
jtomasek, line 228:
```suggestion
      Close
```

src/components/clusterDetail/ConsoleModal.css
```diff
@@ -0,0 +1,3 @@
+.console-modal-expandable {
```
jtomasek, line 1: It should be possible to avoid using custom styles by using `Stack` layout for the expandable items https://www.patternfly.org/v4/layouts/stack#with-gutter

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -30,39 +32,121 @@ type ConsoleModalProps = WebConsoleHintProps & {
...
+  optionalList?: string[];
+};
+
+const ModalExpandableSection: React.FC<ModalExpandableSectionProps> = (props) => {
+  const { requiredList, optionalList, ...otherProps } = props;
+  return (
+    <ExpandableSection {...otherProps}>
+      <TextContent>
+        <TextList>
+          <Text component="p">
```
jtomasek, line 50: React will warn when putting `div`s inside `p`. (`Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.`)

We should use `TextListItem` but PatternFly does not provide list-style-type: none `li` :/

Alternatively, since the content of this is going to be usually copied by the user, why not render the lists in a `<PrismCode />` component?

<TextContent>
  <Text component="h6">Title if needed<Text>
  <PrismCode .... />
  ...

jtomasek (changes requested):

src/components/clusterDetail/ConsoleModal.tsx
irosenzw: PrismCode is an excellent idea!

andybraren (dismissed):
Thanks for including the great GIF of the behavior, and I like that the first `(recommended)` option is expanded by default. LGTM! 👍

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -30,39 +31,122 @@ type ConsoleModalProps = WebConsoleHintProps & {
...
+
+const ModalExpandableSection: React.FC<ModalExpandableSectionProps> = (props) => {
+  const { requiredList, optionalList, ...otherProps } = props;
+
+  return (
+    <ExpandableSection {...otherProps}>
+      <TextContent>
+        <TextList>
+          <PrismCode code={formatMultilineString(requiredList)} language="yaml" />
+          <Text component="p">Optional:</Text>
```
jtomasek, line 56:
```suggestion
          <Text component="h6">Optional:</Text>
```

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -30,39 +31,122 @@ type ConsoleModalProps = WebConsoleHintProps & {
...
+    : '';
+
+const ModalExpandableSection: React.FC<ModalExpandableSectionProps> = (props) => {
+  const { requiredList, optionalList, ...otherProps } = props;
+
+  return (
+    <ExpandableSection {...otherProps}>
+      <TextContent>
+        <TextList>
+          <PrismCode code={formatMultilineString(requiredList)} language="yaml" />
```
jtomasek, line 55: `markup` seems to work best for the plain configuration file

```suggestion
          <PrismCode code={formatMultilineString(requiredList)} language="markup" />
```

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -30,39 +31,122 @@ type ConsoleModalProps = WebConsoleHintProps & {
...
+          <Text component="p">Optional:</Text>
+          <PrismCode code={formatMultilineString(optionalList)} language="yaml" />
```
jtomasek, line 57:
```suggestion
          <PrismCode code={formatMultilineString(optionalList)} language="markup" />
```

jtomasek (changes requested):
Looks nice, a few nits inline.

jtomasek (approved):

PR #327, opened 2020-11-24
MGMT-3077 Add support for NTP configuration

jtomasek:
- Add NTP source input field in advanced network configuration
- Add support for ntp-synced backend network validation
- Add NTP status to host detail

Depends on https://github.com/openshift/assisted-service/pull/675

Failing validation + NTP status unreachable:
Screenshot from 2020-11-24 14-11-27

Configuring NTP source:
Screenshot from 2020-11-24 14-12-18

NTP synced:
Screenshot from 2020-11-24 13-58-42

### DIFF
--- a/src/components/hosts/HostRowDetail.tsx
+++ b/src/components/hosts/HostRowDetail.tsx
@@ -7,10 +7,13 @@ import { Host, Inventory, Disk, Interface } from '../../api/types';
 import { getHostRowHardwareInfo } from './hardwareInfo';
 import { DASH } from '../constants';
 import { DetailList, DetailListProps, DetailItem } from '../ui/DetailList';
+import { ValidationsInfo } from '../../types/hosts';
+import NtpValidationStatus from './NtpValidationStatus';
 
 type HostDetailProps = {
   inventory: Inventory;
   host: Host;
+  validationsInfo: ValidationsInfo;
 };
 
 type SectionTitleProps = {
@@ -123,7 +126,7 @@ const NicsTable: React.FC<NicsTableProps> = ({ interfaces }) => {
   );
 };
 
-export const HostDetail: React.FC<HostDetailProps> = ({ inventory, host }) => {
+export const HostDetail: React.FC<HostDetailProps> = ({ inventory, host, validationsInfo }) => {
   const { id } = host;
   const rowInfo = getHostRowHardwareInfo(inventory);
   const disks = inventory.disks || [];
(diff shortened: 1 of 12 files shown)


src/components/hosts/HostRowDetail.tsx
```diff
@@ -146,6 +149,10 @@ export const HostDetail: React.FC<HostDetailProps> = ({ inventory, host }) => {
         <DetailItem title="CPU architecture" value={inventory.cpu?.architecture || DASH} />
         <DetailItem title="CPU model name" value={inventory.cpu?.modelName || DASH} />
         <DetailItem title="CPU cores and clock speed" value={rowInfo.cpuSpeed} />
+        <DetailItem
```
mareklibra, line 152: Can we move this to the third column?
This one is under redesign in other open PRs.

jtomasek: Done

nirfarkas:
I think what's missing in the notification is a way to fix this issue. Maybe something like: "Please specify an NTP server in the Advanced Networking section".
Also, does the NTP field always shown? if so, it should be mandatory and auto-populated (when everything is in sync, and empty for input if there's a problem.

mareklibra:
How is NTP handled in case of Day 2?

jtomasek:
> I think what's missing in the notification is a way to fix this issue. Maybe something like: "Please specify an NTP server in the Advanced Networking section".

I can add `notificationHints` constant to allow adding client side messages to failing validations. Currently only the backend provides the message.

> Also, does the NTP field always shown? if so, it should be mandatory and auto-populated (when everything is in sync, and empty for input if there's a problem.

Yes, the field is always shown. The field is by default pre-populated with `0.rhel.pool.ntp.org` and if the validation fails, user can provide different value. In general, user can fix the validation either by configuring the ntp sources manually on the host or configure ntp source via the UI.
I don't quite understand why the field should have additional logic then just enabling user to configure one ntp source for all hosts.

jtomasek:
> How is NTP handled in case of Day 2?

Technically it is still possible to make update_cluster API calls, however we don't do any operation like this on day 2 yet so we'd have to introduce a dialog to set this which may be quite confusing IMO. So for day 2 I'd prefer to notify user and let them manually fix the problem and ensure that host is ntp synced.

@YuviGold

jtomasek:
Updated the `ntpSource` field label and description, added client side hint to `ntp-synced` validation failure.

Screenshot 2020-11-25 at 15 56 16

Screenshot 2020-11-25 at 16 01 35

nirfarkas:
Hmmm... shouldn't the field be mandatory? I think it can't be empty.

jtomasek:
It can be empty. In some AI setups it has an empty default value.

nirfarkas:
Thanks, Jiri!

mareklibra (approved):

mareklibra (approved):

PR #330, opened 2020-11-26
OCPBUGSM-20382 Day 2 post-installation instructions

mareklibra:


src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
```diff
@@ -18,7 +18,10 @@ import { Alerts } from '../ui/AlertsSection';
 import { EventsModalButton } from '../ui/eventsModal';
 import BaremetalInventoryAddHosts from './BareMetalInventoryAddHost';
 
-const AddBareMetalHosts: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
+const AddBareMetalHosts: React.FC<{ cluster: Cluster; ocpConsoleUrl?: string }> = ({
+  cluster,
+  ocpConsoleUrl,
```
jtomasek, line 23: It would be better to provide the ocpConsoleUrl via context instead of passing it through several nested components on which the number of props it just passes down is getting large.

Maybe even `OcpConsoleNodesSectionLink` itself could take care of getting the console url? Even by doing a separate API call?

jtomasek (changes requested):

src/components/AddBareMetalHosts/AddBareMetalHosts.tsx
mareklibra: I think it's better to pass this info from the higher components down instead of local single-purpose request.
Anyway, it makes sense to use React `Context` for it.

mareklibra: done

jtomasek (approved):

jtomasek (dismissed):

PR #334, opened 2020-11-26
OCPBUGSM-20923 A failed day2 host can be reseted

mareklibra:
So installation can be re-triggered.

- [x] depends on ~~#327~~ #336  due to API changes


nirfarkas:
Do we show a confirmation popup after selecting "Reset"? Should we?

mareklibra:
Thanks @nirfarkas , added

jtomasek (approved):

PR #339, opened 2020-12-01
HTML ID change: button-download-discovery-iso button ID is prefixed

mareklibra:
To make it always unique.


mareklibra:
Cc @cdvultur

jtomasek (approved):

PR #343, opened 2020-12-01
Fix AlertSection padding

jtomasek:
Before:
Screenshot from 2020-12-01 15-56-43

After:
Screenshot from 2020-12-01 15-54-57


andybraren:
Not really related to this PR, but I think this alert should be using inline styling rather than the toast styling:
https://www.patternfly.org/v4/components/alert/design-guidelines#usage

irosenzw (approved):

PR #349, opened 2020-12-03
Simplify look-up of the Day 2 cluster

mareklibra:
Latest changes around openshift_cluster_id query look-up were based on wrong
assumption that the look-up is implemented consistently
for both Day 1 and 2 in the backend.

This patch conforms the backend best.


mareklibra:
In parallel to this fix/revert, I will start a discussion about fixing backend.

jtomasek (approved):

PR #351, opened 2020-12-03
move ip-address package to be dependency and not a peerDep

irosenzw:


mareklibra (approved):

mareklibra:
If it works in uhc-portal, we will need to removed this dependency in the `facet` as well.

jtomasek (approved):

irosenzw:
> If it works in uhc-portal, we will need to removed this dependency in the `facet` as well.

yes. we should also remove `netmask` package from facet.

PR #355, opened 2020-12-07
BZ 1905079: Fix null value of Cluster.hosts from API

mareklibra:


mareklibra:
In a very rare case, right after cluster's creation, the API can return `null` instead of an empty array for a cluster's host list.

jtomasek (approved):

PR #385, opened 2020-12-16
MGMT-3235: Move Network config into a wizard step

mareklibra:
For follow-ups:
- implement transition functions (especially decide on backend validations)
- alerts (relevant for a step)
- rename the `clusterConfiguration` subfolder
- other steps + final summary
- test, test, test - transitions and data consistency after partial updates


mareklibra:
Rebased with many conflicts :-(

src/components/clusterConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+          );
+        };
+        const onSshKeyBlur = () => {
+          if (values.sshPublicKey) {
+            setFieldValue('sshPublicKey', trimSshPublicKey(values.sshPublicKey));
+          }
+        };
+
+        const form = (
+          <Form>
```
jtomasek, line 150: Wrap the `Form` with a grid to limit fields width
```suggestion
          <Grid hasGutter>
            <GridItem span={12} lg={10} xl={9} xl2={7}>
              <Form>
```

src/components/clusterWizard/constants.ts
```diff
@@ -0,0 +1,13 @@
+// export const CLUSTER_STEP_DETAIL = 'cluster-detail';
+export const CLUSTER_STEP_CONFIGURATION = 'cluster-configuration';
```
jtomasek, line 2: Do we need these contstants? Would defining a type and enforce it where needed be enough?
```
type ClusterSteps = 'cluster-configuration' | 'networking' | ...
```

src/components/clusterWizard/wizardTransition.tsx
```diff
@@ -0,0 +1,58 @@
...
+    uiValidation = isValid && !isSubmitting;
+  }
+
+  // TODO(mlibra): check backend validations
+  const backendValidation = true;
+
+  return uiValidation && backendValidation;
+};
+
+const getNextClusterWizardStep = (step: string) =>
```
jtomasek, line 39: I feel like with this module, we're trying to teach the wizard to know about the steps order and be able to go to next/prev step from anywhere. It is very similar to what PatternFly implementation does - it introduces a Wizard component which knows how to transition from step to step.

I think this is not really needed. We can follow simple rules:
- Each step knows what is it's next step
- It is possible to go to any previous step from each step
- It is only possible to the next step by clicking 'next' in a specific step

I think there is no need to generalize the transition functions (using TransitionProps) let's define canNext* functions on each step - these functions can be completely tailored to each individual step without needing to generalize it's parameters.

jtomasek: We can have a 'utils' module where we keep the functions which are used by individual steps such as `getClusterNetworkStatus`, `getClusterBaremetalInventoryStatus`...

src/components/clusterWizard/wizardTransition.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const useClusterWizardTransitionFunctions = (step: string) => {
+  const { setCurrentStepId } = React.useContext(ClusterWizardContext);
+
+  const stepBack = getBackClusterWizardStep(step);
+  const stepNext = getNextClusterWizardStep(step);
+  const onBack = stepBack ? () => setCurrentStepId(stepBack) : undefined;
+  const onNext = stepNext ? () => setCurrentStepId(stepNext) : undefined;
+
+  return { onBack, onNext };
```
jtomasek, line 57: I really feel like these are not necessary when in each step we can simply call `setCurrentStepId('the-step-which-goes-after-me`).
Having these functions would be good if we had a wrapping entity above the steps and individual steps wouldn't know what are the next steps but that's not the case.

src/components/clusterConfiguration/ClusterConfigurationForm.tsx
```diff
@@ -222,84 +136,15 @@ const ClusterConfigurationForm: React.FC<ClusterConfigurationFormProps> = ({
...
-                )}
-              </ClusterToolbar>
+              <ClusterWizardToolbar
+                cluster={cluster}
+                errors={errors}
+                dirty={dirty}
+                isSubmitting={isSubmitting}
+                isNextDisabled={!canNextClusterConfiguration({ isValid, isSubmitting, cluster })}
+                onNext={submitForm}
+                onBack={onBack}
```
jtomasek, line 146: As described in other comments, I'd prefer to just use `() => setCurrentStepId('networking')`

src/components/clusterConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+        params.machineNetworkCidr = cidr;
+      }
+
+      const { data } = await patchCluster(cluster.id, params);
+      formikActions.resetForm({
+        values: getNetworkInitialValues(data, managedDomains),
+      });
+      dispatch(updateCluster(data));
+
+      if (canNextNetwork({ cluster })) {
```
jtomasek, line 108: This is the example that trying to make `canNextNetwork` reusable is confusing. `canNextNetwork` is checking if form is submitting and valid but this time we're passing `cluster` prop to it. So we're trying to have a single function doing 2 very different things. I'd much rather prefer this here:

```suggestion
      if (getClusterHardwareStatus(cluster) === 'ready') {
        setCurrentStepId('networking')
        ...
      
```

If there are more things to check, we wrap it in
```
canTransitionToNextStep = (cluster) => boolean
```

jtomasek (changes requested):

src/components/clusterWizard/wizardTransition.tsx
mareklibra: Ok, agree. Unnecessary abstraction, will simplify

mareklibra:
Rebased after 1.5.6

jtomasek (approved):

PR #394, opened 2021-01-03
MGMT-3487 - Use IPv6 short form in subnet selection dropdown

omertuc:
This is just a suggestion, feel free to close the PR / make changes.

Instead of:
image

Show `fe70::/64 (fe80::-fe80::ffff:ffff:ffff:ffff)`

The subnet, along with the full start and end address, look rather awkward inside the dropdown because they form a very long string.

### DIFF
--- a/src/components/clusterConfiguration/utils.ts
+++ b/src/components/clusterConfiguration/utils.ts
@@ -25,7 +25,7 @@ const getHumanizedSubnet = (subnet: Address6 | Address4 | null) => {
   if (subnet) {
     const subnetStart = subnet.startAddress().correctForm();
     const subnetEnd = subnet.endAddress().correctForm();
-    return `${subnet.address} (${subnetStart}-${subnetEnd})`;
+    return `${subnet.address} (${subnetStart} - ${subnetEnd})`;
   }
 
   return '';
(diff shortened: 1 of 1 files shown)


src/components/clusterConfiguration/utils.ts
```diff
@@ -21,10 +21,15 @@ export const getSubnet = (cidr: string): Address6 | Address4 | null => {
...
 
-const getHumanizedSubnet = (subnet: Address6 | Address4 | null) =>
-  subnet
-    ? `${subnet.address} (${subnet.startAddress().address}-${subnet.endAddress().address})`
-    : '';
+const getHumanizedSubnet = (subnet: Address6 | Address4 | null) => {
+  if (subnet) {
+    const subnetStart = subnet.startAddress().correctForm();
+    const subnetEnd = subnet.endAddress().correctForm();
+    return `${subnet.address} (${subnetStart}-${subnetEnd})`;
```
jtomasek, line 28:
```suggestion
    return `${subnet.address} (${subnetStart} - ${subnetEnd})`;
```

jtomasek (changes requested):
Thanks! Small nit inline.

src/components/clusterConfiguration/utils.ts
omertuc: Actually wanted to do that myself but felt it was too opinionated :sweat_smile:

jtomasek (approved):

PR #396, opened 2021-01-04
MGMT-2840: add VM Reboot config info

mareklibra:


src/components/AddBareMetalHosts/BareMetalInventoryAddHost.tsx
```diff
@@ -42,8 +43,9 @@ const BaremetalInventoryAddHosts: React.FC = () => {
           <HostsNotShowingLink setDiscoveryHintModalOpen={setDiscoveryHintModalOpen} />
         </Text>
         <HostRequirements ContentComponent={HostRequirementsContent} />
+        <FormatDiskWarning />
+        <VMRebootConfigurationInfo hosts={cluster.hosts} />
       </TextContent>
-      {/* <FormatDiskWarning /> */}
```
mareklibra, line 46: @jtomasek Please confirm this should be already re-enabled

mareklibra:
Cc @andybraren @nirfarkas  wdyt?
I have co-located the warnings.

andybraren (dismissed):
Co-locating the two inline warnings makes sense to me, thanks Marek, LGTM! 👍

src/components/AddBareMetalHosts/BareMetalInventoryAddHost.tsx
jtomasek: Yes, it should be enabled.

src/components/hosts/VMRebootConfigurationInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+  <>
+    <Text component={TextVariants.p}>
+      Your libvirt virtual machines should be configured to restart automatically after a reboot.
+      You can check this by running:
+    </Text>
+    <Text component={TextVariants.p}>
+      <code>
+        virsh dumpxml {'<'}name{'>'} | grep on_reboot{' '}
+      </code>
+    </Text>
```
jtomasek, line 17: It would be better to use `PrismCode` component for code snippets. See `DiscoveryTroubleshootingModal` component.

jtomasek (changes requested):

src/components/hosts/VMRebootConfigurationInfo.tsx
mareklibra: Done. What about changing the default theme to something better fitting the rest of the page? https://github.com/PrismJS/prism-themes
But in a follow-up.

mareklibra:
With Prism:


What about changing the default theme to something better fitting the rest of the page? https://github.com/PrismJS/prism-themes
But in a follow-up. @jtomasek , @andybraren , @nirfarkas

src/components/hosts/VMRebootConfigurationInfo.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { Button, ButtonVariant, Popover, Text, TextVariants } from '@patternfly/react-core';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
+import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens';
+import { Host, stringToJSON, Inventory } from '../../api';
+import PrismCode from '../ui/PrismCode';
+
+const Hint = () => (
+  <>
```
jtomasek, line 9:
```suggestion
  <TextContent>
```

src/components/hosts/VMRebootConfigurationInfo.tsx
```diff
@@ -0,0 +1,46 @@
...
+      You can check this by running:
+    </Text>
+    <PrismCode code="virsh dumpxml <name> | grep on_reboot" />
+    <Text component={TextVariants.p}>And verify that this is the output:</Text>
+    <PrismCode code="<on_reboot>restart</on_reboot>" />
+    <Text component={TextVariants.p}>
+      If not, please start your VMs with the following configuration:
+    </Text>
+    <PrismCode code="--events on_reboot=restart" />
+  </>
```
jtomasek, line 21:
```suggestion
  </TextContent>
```

nirfarkas:
@mareklibra I think it looks clearer with the Prism theme! It's easier to understand where is the code you need to copy.

jtomasek:
Thinking about the alerts, shouldn't we change the icons to `InfoCircleIcon` rather than using warning icons? By the nature of those the alerts are not going away so maybe presenting it just as something to note instead of warning would be better? @nirfarkas

src/components/hosts/VMRebootConfigurationInfo.tsx
mareklibra: done

mareklibra:
@nirfarkas I was thinking about changing the theme for PrismCode. Here and in the `DiscoveryTroubleshootingModal` as well.

Examples, my preference is in the `github` theme:

nightOwlLight:
01_nightOwlLight

---

github:
02_github

---

ultramin:
03_ultramin

andybraren:
Since we already use PrismCode the `github` theme makes sense to me for now, but a copy-able `<pre><code>` block with syntax highlighting seems like something that could be a "native" PatternFly component in the future (unless I'm just not finding it). Do you agree?

The read-only [Clipboard copy component](https://www.patternfly.org/v4/components/clipboard-copy) with `isCode` styling (like in [this example](https://www.patternfly.org/v4/components/clipboard-copy#json-object-wrap-code-with-pre)) is pretty close and could probably be used for the first and third blocks in this popover to make it easier to copy each command, but a form input isn't quite a "code block" and wouldn't work for the DNS troubleshooting modal's larger boxes.

I'll ask around, but https://github.com/patternfly/patternfly-design/issues/836 seems to be the most relevant PF issue for this. The same pattern will likely be needed for certain parts of the OCP Web Console, and probably areas of c.rh.c as well.

Also FYI, I know the Monaco editor is going to be used for a native Code Editor component in the near future (see [this PR](https://github.com/patternfly/patternfly-react/pull/5294)) and we'll need Monaco or that new component for the Assisted Installer's YAML manifest editor pretty soon. In that case, would using Monaco's [Monarch syntax highlighter](https://microsoft.github.io/monaco-editor/monarch.html) make sense to use instead of PrismCode? 🤷‍♂️

andybraren:
> Thinking about the alerts, shouldn't we change the icons to `InfoCircleIcon` rather than using warning icons? By the nature of those the alerts are not going away so maybe presenting it just as something to note instead of warning would be better? @nirfarkas

@jtomasek See [MGMT-2840](https://issues.redhat.com/browse/MGMT-2840) for some discussion about this.

nirfarkas:
@jtomasek I'd rather use warnings here because it's information that the user shouldn't disregard. It was nice if we had some feedback mechanism to check this but this is really out of scope so the warning will be there as long as there are hosts that are VMs

nirfarkas:
> Since we already use PrismCode the `github` theme makes sense to me for now, but a copy-able `<pre><code>` block with syntax highlighting seems like something that could be a "native" PatternFly component in the future (unless I'm just not finding it). Do you agree?
> 
> The read-only [Clipboard copy component](https://www.patternfly.org/v4/components/clipboard-copy) with `isCode` styling (like in [this example](https://www.patternfly.org/v4/components/clipboard-copy#json-object-wrap-code-with-pre)) is pretty close and could probably be used for the first and third blocks in this popover to make it easier to copy each command, but a form input isn't quite a "code block" and wouldn't work for the DNS troubleshooting modal's larger boxes.
> 
> I'll ask around, but [patternfly/patternfly-design#836](https://github.com/patternfly/patternfly-design/issues/836) seems to be the most relevant PF issue for this. The same pattern will likely be needed for certain parts of the OCP Web Console, and probably areas of c.rh.c as well.
> 
> Also FYI, I know the Monaco editor is going to be used for a native Code Editor component in the near future (see [this PR](https://github.com/patternfly/patternfly-react/pull/5294)) and we'll need Monaco or that new component for the Assisted Installer's YAML manifest editor pretty soon. In that case, would using Monaco's [Monarch syntax highlighter](https://microsoft.github.io/monaco-editor/monarch.html) make sense to use instead of PrismCode? 🤷‍♂️

From a user experience perspective, it doesn't matter what theme we'll use as long it's clear that some parts are instructions and the other parts are code (that can be copied, optionally)

jtomasek (approved):

mareklibra:
@andybraren Thank you for the thorough search of code-snippet formatting options.

I am keeping this PR at recent `nightOwl` theme with intention to change it in a follow-up.

The Monaco editor looks great, but we would need to check if we can use it just for syntax highlighting. Mounting whole editor just for a pop-up seems overkilling to me.
In any case, I have found following on there homepage:
```
The Monaco editor is not supported in mobile browsers or mobile web frameworks. 
```
So let's skip it for this use-case.

Regarding https://github.com/patternfly/patternfly-design/issues/836, let's see result of their ding and how it fits the one-liner instruction use-case.

In the intended follow-up, let's explore possibility of either new PrismCode theme or `<pre><code>`.

andybraren:
> From a user experience perspective, it doesn't matter what theme we'll use as long it's clear that some parts are instructions and the other parts are code (that can be copied, optionally)

@nirfarkas Agreed, from a portfolio consistency and code maintainability perspective I'm just biased to think "if something could be used in multiple places/products across multiple teams and isn't in PatternFly yet, maybe it should be." I'm just noticing the potential gap/opportunity here. 🙂

Good catch @mareklibra about Monaco's statement about mobile. I'll need to learn more about how the PatternFly component will handle that, since most of the time our components are mobile/touch-accessible. And agreed that we wouldn't want to load the full editor in this case. Will follow up.

PR #400, opened 2021-01-07
Replace OpenshiftVersionsContext by useOpenshiftVersions() React hook

mareklibra:


src/components/clusters/NewClusterPage.tsx
```diff
@@ -40,14 +40,24 @@ type NewClusterFormProps = {
 
 const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '' }) => {
   const { addAlert, clearAlerts } = React.useContext(AlertsContext);
-  const { versions, getDefaultVersion } = React.useContext(OpenshiftVersionsContext);
+  const {
+    error: errorOCPVersions,
+    loading: loadingOCPVersions,
+    versions,
+    getDefaultVersion,
+  } = useOpenshiftVersions();
```
jtomasek, line 48: I'd move this into `NewCluster` component (same as we have pullSecret fetching) and pass the `versions` to a `NewClusterForm` as props. Doing this allows the versions prop to be always defined (`[]` in case of an error) and avoids us having to use `enableReinitialize` on the form.

src/components/clusters/NewClusterPage.tsx
```diff
@@ -40,14 +40,24 @@ type NewClusterFormProps = {
...
   const history = useHistory();
 
   const nameInputRef = React.useRef<HTMLInputElement>();
   React.useEffect(() => {
     nameInputRef.current?.focus();
   }, []);
 
+  React.useEffect(() => errorOCPVersions && addAlert(errorOCPVersions), [
```
jtomasek, line 56: This could be done in `NewCluster` component

src/components/clusters/NewClusterPage.tsx
```diff
@@ -85,13 +95,27 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '' }) => {
     }
   };
 
-  const getOpenshiftVersionHelperText = (value: string) =>
-    versions.find((version) => version.value === value)?.supportLevel !== 'production' ? (
+  const getOpenshiftVersionHelperText = (value: string) => {
+    return versions &&
```
jtomasek, line 99: This wouldn't be needed if change described above is done.

src/components/clusters/NewClusterPage.tsx
```diff
@@ -85,13 +95,27 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '' }) => {
...
+      versions.find((version) => version.value === value)?.supportLevel !== 'production' ? (
       <>
         <ExclamationTriangleIcon color={warningColor.value} size="sm" />
         &nbsp;Please note that this version is not production ready.
       </>
     ) : null;
+  };
+
+  const ocpVersionOptions = versions
```
jtomasek, line 108: This wouldn't be needed as well.

src/components/clusters/NewClusterPage.tsx
```diff
@@ -102,6 +126,7 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '' }) => {
           openshiftVersion: getDefaultVersion(),
           pullSecret: pullSecret,
         }}
+        enableReinitialize
```
jtomasek, line 129: Not needed with the above change

jtomasek (changes requested):

src/components/clusters/NewClusterPage.tsx
mareklibra: Done. New `OpenshiftVersionSelect` component is introduced and all this logic co-located there

src/components/clusters/OpenshiftVersionSelect.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
+import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens';
+import SelectField from '../ui/formik/SelectField';
+import { AlertsContext } from '../AlertsContextProvider';
+import { useOpenshiftVersions } from '../fetching/openshiftVersions';
+import { useFormikContext } from 'formik';
+import { ClusterCreateParams } from '../../api';
+
+const OpenshiftVersionSelect: React.FC = () => {
```
jtomasek, line 10: Doing this in a field level component introduces several problems:
- the initial value of the form is incorrect and if we need to reset the form it won't work correctly
- In case the cluster already exists and we want to use this to find a matching version this won't work
- we're updating the form from within the field which mangles the form properties such as touched, dirty etc.

IMHO we should fetch all necessary data for the form in a wrapping component (`NewCluster`) as we already do with pullSecret and then pass the data to the form component and let Formik initialize with the values based on the data.

jtomasek (changes requested):

jtomasek (approved):
Looks great, thanks!

PR #401, opened 2021-01-07
Use github prism theme as default

mareklibra:
Follow-up for #396.


mareklibra:
Former state:



---

mareklibra:
From the discussed options I like using of the PrismCode but with different theme at most:



---

mareklibra:
Cc @andybraren , @nirfarkas , @jtomasek

jtomasek (approved):

nirfarkas:
Looking Great!

PR #406, opened 2021-01-12
MGMT-3290, MGMT-3338 Add disk management to host detail

jtomasek:
MGMT-3290 Add disk limitations and roles to host detail
MGMT-3338 Add disk role setting functionality to host detail

disk-role-selection

### DIFF
--- a/src/components/hosts/DiskLimitations.tsx
+++ b/src/components/hosts/DiskLimitations.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import { Alert, AlertVariant, Button, ButtonVariant, Popover } from '@patternfly/react-core';
+import { WarningTriangleIcon } from '@patternfly/react-icons';
+import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens';
+import { Disk } from '../../api/types';
+
+type DiskLimitationsProps = {
+  disk: Disk;
+};
+const DiskLimitations: React.FC<DiskLimitationsProps> = ({ disk }) => {
+  if (!disk.installationEligibility?.eligible) {
+    return (
+      <Popover
+        headerContent={<div>Disk limitations</div>}
+        bodyContent={<DiskLimitationsPopoverContent {...disk.installationEligibility} />}
+        minWidth="30rem"
+        maxWidth="50rem"
+      >
+        <Button
+          variant={ButtonVariant.link}
+          icon={<WarningTriangleIcon color={warningColor.value} />}
+          isInline
+        >
+          Limitations
+        </Button>
+      </Popover>
+    );
+  }
+  return null;
+};
+
+export default DiskLimitations;
+
+type DiskLimitationsPopoverContentProps = {
+  eligible?: boolean;
+  notEligibleReasons?: string[];
+};
+
+const DiskLimitationsPopoverContent: React.FC<DiskLimitationsPopoverContentProps> = ({
+  eligible,
+  notEligibleReasons,
+}) => {
+  const alerts = [];
+
+  if (!eligible) {
+    alerts.push(
+      <Alert
+        key="install-eligibility"
+        variant={AlertVariant.warning}
+        title="Disk is not eligible for installation"
+        isInline
+      >
+        {notEligibleReasons && (
+          <ul>
+            {notEligibleReasons.map((reason: string) => (
+              <li key={reason}>{reason}</li>
+            ))}
+          </ul>
+        )}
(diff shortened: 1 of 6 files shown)


omertuc:
Looks great!

One suggestion - maybe add bullet points for the reasons?

jtomasek:
> One suggestion - maybe add bullet points for the reasons?
Yeah, PatternFly removes them for some reason :/

jtomasek:
Fixed the alert list bullets.

jtomasek:
Updated disk limitations list:
Screenshot 2021-01-12 at 16 03 03

and removed the warning icon next to the disk limitations count:
Screenshot 2021-01-12 at 16 02 53

omertuc:
When the service gets its first inventory it automatically sets the `installation_disk_path` of the host to one of the disks in the inventory. Will this automatically be reflected in the UI? i.e. will there automatically be a disk with an "installation" role, as dictated by the host installation_disk_path field in the api, even without the user making any manual choice?

jtomasek:
> When the service gets its first inventory it automatically sets the `installation_disk_path` of the host to one of the disks in the inventory. Will this automatically be reflected in the UI? i.e. will there automatically be a disk with an "installation" role, as dictated by the host installation_disk_path field in the api, even without the user making any manual choice?

Yes

omertuc:
> and removed the warning icon next to the disk limitations count:

I liked the limitations counts, it made the ineligible disks stand-out at a glance. Is it possible to bring it back / gray-out the rows for disks that are ineligible?

jtomasek:
@nirfarkas ^

nirfarkas:
> Is it possible to bring it back / gray-out the rows for disks that are ineligible? Or maybe do something else that will make them stand out?
@omertuc 
I would advise not doing this because when we show the warning icon it looks like there's something wrong with the specific disk (which is not true). 
A second reason for removing the icon and keeping all the rows identical is that although currently, we have only one role ("Installation Disk") in the future we're planning to have more roles from OCS. If we highlight some rows as not eligible for installation we would have to highlight also disks not eligible for other roles and maybe distinguish between them... 
Also, "Installation Disk" is not eligible for OCS usage so should we also highlight it?
Last (but not least), one of the principles of UX design is to reduce "noise" to it's minimum - we don't want to add colors and indications for everything (just for the important ones). A disk might have role limitations, it doesn't imply that it is disabled (:

omertuc:
> I would advise not doing this because when we show the warning icon it looks like there's something wrong with the specific disk (which is not true).
A second reason for removing the icon and keeping all the rows identical is that although currently, we have only one role ("Installation Disk") in the future we're planning to have more roles from OCS. If we highlight some rows as not eligible for installation we would have to highlight also disks not eligible for other roles and maybe distinguish between them...

Makes sense.

> Also, "Installation Disk" is not eligible for OCS usage so should we also highlight it?
Last (but not least), one of the principles of UX design is to reduce "noise" to it's minimum - we don't want to add colors and indications for everything (just for the important ones). A disk might have role limitations, it doesn't imply that it is disabled (:

Fair enough.

So a second proposal to consider:

If a disk is completely useless for all roles (for example, this will always happen to the ISO/Media disk which is always there for every host), highlight it in some way as irrelevant.

I'd argue that in the context of the installer, disks that are not fit for any role can be considered "noise".

Maybe change the role from "None" to something else (can't come up with good ideas right now - some synonym to irrelevant, useless, unfit, etc)? Maybe remove the arrow from the "None" to clarify that there's no point even trying to click it? Maybe gray out the "role" or the entire row? 

The reason I bring it up is because as it is now, the disk role column is very "inviting" for the user to try and interact with, but it will yield no results and will just be empty.

src/components/hosts/DiskLimitations.tsx
```diff
@@ -0,0 +1,67 @@
...
+  disk: Disk;
+};
+const DiskLimitations: React.FC<DiskLimitationsProps> = ({ disk }) => {
+  if (!disk.installationEligibility?.eligible) {
+    const limitationsCount = disk.installationEligibility?.notEligibleReasons?.length || 0;
+    return (
+      <Popover
+        headerContent={<div>Disk limitations</div>}
+        bodyContent={<DiskLimitationsPopoverContent {...disk.installationEligibility} />}
+        minWidth="30rem"
```
mareklibra, line 23: This is a repeating pattern, we should prbably introduce a wrapping component

mareklibra (approved):

PR #416, opened 2021-01-22
Bug 1901944 - [4.6][UI] Severity filter in cluster events clears all check-boxes if one is deselected

jkilzi:
### Fixes
- Uncheking one of the severity filter checkboxes toggled off the rest
- Selecting only one host along with deleted and cluster-level events checkboxes made the `Select All` checkbox in the hosts filter to be toggled on

Peek 2021-01-22 02-46


src/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -90,21 +90,42 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
...
           ? filters.selectAll
-          : isChecked &&
-            allOtherSelectedHosts.length === filters.hosts?.length &&
+          : filters.hosts.length === totalHostsInCluster &&
             filters.orphanedHosts &&
             filters.clusterLevel,
     });
+
+    setFilters(
+      _.flow(
```
mareklibra, line 124: Future bug-fixer will love you ;-)

jkilzi: This is just performing plain functional composition (`flow` and `curry` are just there in order to make the order of the  updates more declarative - the functions are applied from left to right inside `flow`, just like a pipe `|` operator would do in `bash`).  Anyway, just to make everybody happy I would like to suggest the following change, which does exactly the same:

```javascript
const updatedFilters = setNextSelectedValues(type, isChecked, value, filters);
const nextFilters = setNextSelectAllValue(cluster.hosts?.length ?? 0, updatedFilters);

setFilters(nextFilters);
```

WDYT?

jtomasek: lgtm

mareklibra: It's easier to read now

jkilzi: resolved in 76b9bf976d7d66d1f83ffb240361ffa48bd45452

src/components/ui/ClusterEventsList.test.tsx
```diff
@@ -0,0 +1,249 @@
+import React from 'react';
+import { render, screen, within } from '@testing-library/react';
+import ClusterEventsList, { ClusterEventsListProps } from './ClusterEventsList';
+
+// TODO: this fixture needs a warm place where to live in the project.
+const props = Object.freeze<ClusterEventsListProps>({
```
mareklibra, line 6: Per our former discussions, we would like to avoid the need to maintain mocks. The API is still not stable and such effort will kill us. I totally agree that we need easy-to-execute tests but we need to consider whole flow as well.

Let's use unit-tests for functionality which does not require (much of) mocking. At least till the API is claimed to be stable. Then we can re-evaluate.

For the rest of functionality, let's implement such testing using Cypress in the `assisted-ui`. And when writing such tests there, let's design them for reusability within e2e downstream flow which does not run on top of the test-infra but deploys its own environment and clusters (in other words: write the test parametrized).

mareklibra, line 6: Since this test has already been written, let's keep it here. We can either maintain or remove later.
Anyway, I would prefer to rewrite that test to the Cypress in a follow-up.

mareklibra (approved):

src/components/ui/ClusterEventsList.test.tsx
jkilzi: Things are starting to get scrambled here so I'll do my best in order to take things back to the right context.

First of all the conversation we had around "mocking" involved a mock-back-end server. The fixture you see here is not meant to be provided by such facility. What you see here is a minimal snapshot of the data required to test part of the functionality of this component in isolation.

The API not being stable can affect testing at any level not only unit-test-level. So yes, if the API changes and is not versioned then you'll have to change the code and that also includes unit-tests along with mock data / fixtures.

What I'm concerned about here is the process:
- when you stumble upon a bug who's duty is to prove it is reproducible?
- who's duty is to write an automated test in order to assure, if there is a regression, you get alerted fast?

Systems change by nature, making the system easy to change is part of the development team effort - not QE. (more on this [here](https://martinfowler.com/articles/practical-test-pyramid.html#:~:text=The%20%22Test%20Pyramid%22%20is%20a,into%20buckets%20of%20different%20granularity.&text=It%20shows%20which%20kinds%20of,how%20these%20can%20be%20implemented.) and [here](https://www.agilealliance.org/glossary/unit-test/#q=~(infinite~false~filters~(postType~(~'page~'post~'aa_book~'aa_event_session~'aa_experience_report~'aa_glossary~'aa_research_paper~'aa_video)~tags~(~'unit*20test))~searchTerm~'~sort~false~sortDirection~'asc~page~1)))

You also said: "Let's use unit-tests for functionality which does not require (much of) mocking"
But at unit-test-level and at integration test level, that's the whole point! You use mock data in order to not make I/O operations (e.g. API calls) and you're doing that in order to be able to test components in isolation.

We need to discuss this issue in depth, otherwise things will get harder to maintain.

PR #419, opened 2021-01-22
Launch dev environment with one command

jkilzi:
This PR introduces a new script which can be run using `yarn concurrently` and allows for running the 3~ish commands needed to start the development environment.

RecApp-2021-01-22-20_39_33


package.json
```diff
@@ -14,7 +14,8 @@
   "scripts": {
     "prebuild": "run-s test",
     "build": "microbundle-crl --no-compress --format modern,cjs --css-modules false",
-    "start": "echo Consider running: yarn sync-to-ui ; echo ; microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
+    "start": "microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
+    "concurrently": "./scripts/concurrently.sh",
```
mareklibra, line 18: I would rename it to `start-asisted-ui` to be clear about consequences. We are developing even the `uhc-portal` exclusively, without `assisted-ui` involved.

jkilzi: @mareklibra I'm not sure I understand what you mean by "We are developing even the uhc-portal exclusively, without assisted-ui involved" (What do you call `uhc-portal`?). I named it this way following @jtomasek recommendation. Personally I would rather name it `start:dev` just to keep it short and also to reflect the fact that the script is modifying the way it launches the project. Anyway, it can be whatever you decide :)

mareklibra: When developing the `assisted-ui-lib`, I recently choose from following options based on actual context:
- see changes in assisted-ui (this is what this PR recently addresses)
- see changes in uhc-portal only (I do not care about assisted-ui and don't want it to be running)
- see changes in both

Since uhc-portal dev env execution brings another level of complexity, I am fine in simplifying just the `assisted-ui` flow now. My suggestion about renaming the yarn script is motivated by keeping other flows seamlessly in the play.

jkilzi: Now I understand your point.

Honestly I couldn't adapt this solution for the `uhc-portal` since I'm not familiar (yet?) with that project/codebase. I'll be more than happy to make such adaptation if you'll find it useful though.

If I rename the script to `start-assisted-ui` will that be enough for now?

mareklibra: yes :-)

mareklibra (approved):

jkilzi:
@mareklibra  now I this output makes more sense to me :)
```bash
[sync-to-ui] Sync to ../uhc-portal/node_modules/openshift-assisted-ui-lib skipped
[sync-to-ui] cp: cannot create regular file '../uhc-portal/node_modules/openshift-assisted-ui-lib/package.json': No such file or directory
```

jtomasek:
Looks good, note that the script will fail to start the assisted-ui dev server unless user exports REACT_APP_API_URL

jkilzi:
@jtomasek agree. But this was the case also before this PR.

PR #421, opened 2021-01-25
MGMT-2974: Allow the user to install Single Node OpenShift

irosenzw:
**new cluster page:**
(The checkbox is unchecked by default)
new cluster page

------------------------------------------------------------------------------------------------
**cluster configuration page (No subnet, API-VIP and Ingress VIP selection):**
readyToInstall

------------------------------------------------------------------------------------------------
**installing page ( no  API-VIP and Ingress VIP properties) :**
installing phase

### DIFF
--- a/src/components/clusterConfiguration/BaremetalInventory.tsx
+++ b/src/components/clusterConfiguration/BaremetalInventory.tsx
@@ -10,6 +10,7 @@ import {
   DiscoveryTroubleshootingModal,
 } from './DiscoveryTroubleshootingModal';
 import FormatDiskWarning from './FormatDiskWarning';
+import { isSingleNodeCluster } from './utils';
 
 const HostRequirementsContent = ({
   worker = {},
@@ -24,6 +25,17 @@ const HostRequirementsContent = ({
     Two or more additional worker hosts are recommended with at least {worker.cpuCores || 2} CPU
     cores, {worker.ramGib || 8} GB of RAM, and {worker.diskSizeGb || 120}
     GB of filesystem storage each.
+  </Text>
+);
+
+const SingleHostRequirementsContent = ({
+  master = {},
+}: {
+  master?: HostRequirementsType['master'];
+}) => (
+  <Text component="p">
+    One host is required with at least {master.cpuCores || 4} CPU cores, {master.ramGib || 16} GB of
+    RAM, and {master.diskSizeGb || 120} GB of filesystem storage.
   </Text>
 );
 
@@ -45,7 +57,11 @@ const BaremetalInventory: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
           Hosts connected to the internet will be inspected and automatically appear below.{' '}
           <HostsNotShowingLink setDiscoveryHintModalOpen={setDiscoveryHintModalOpen} />
         </Text>
-        <HostRequirements ContentComponent={HostRequirementsContent} />
+        {isSingleNodeCluster(cluster) ? (
+          <HostRequirements ContentComponent={SingleHostRequirementsContent} />
+        ) : (
+          <HostRequirements ContentComponent={HostRequirementsContent} />
+        )}
         <FormatDiskWarning />
         <VMRebootConfigurationInfo hosts={cluster.hosts} />
       </TextContent>
--- a/src/components/clusterConfiguration/NetworkConfiguration.tsx
+++ b/src/components/clusterConfiguration/NetworkConfiguration.tsx
@@ -6,7 +6,7 @@ import AdvancedNetworkFields from './AdvancedNetworkFields';
 import { HostSubnets, ClusterConfigurationValues } from '../../types/clusters';
 import { InputField, CheckboxField, SelectField } from '../ui/formik';
 import { ManagedDomain, Cluster } from '../../api/types';
-import { getClusterDefaultSettings, isAdvConf } from './utils';
+import { getClusterDefaultSettings, isAdvConf, isSingleNodeCluster } from './utils';
 
 type NetworkConfigurationProps = {
   cluster: Cluster;
@@ -83,7 +83,9 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
           isRequired
         />
       )}
-      <BasicNetworkFields cluster={cluster} hostSubnets={hostSubnets} />
+      {!isSingleNodeCluster(cluster) && (
+        <BasicNetworkFields cluster={cluster} hostSubnets={hostSubnets} />
+      )}
       <Checkbox
         id="useAdvancedNetworking"
         label="Use Advanced Networking"
--- a/src/components/clusters/NewClusterPage.tsx
+++ b/src/components/clusters/NewClusterPage.tsx
@@ -32,6 +32,8 @@ import { usePullSecretFetch } from '../fetching/pullSecret';
 import PullSecret from './PullSecret';
 import { useOpenshiftVersions } from '../fetching/openshiftVersions';
 import { OpenshiftVersionOptionType } from '../../types/versions';
+import SingleNodeCheckbox from '../ui/formik/SingleNodeCheckbox';
+import { ocmClient } from '../../api/axiosClient';
 
 import './NewClusterPage.css';
 
@@ -107,6 +109,7 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versio
           name: '',
           openshiftVersion: versions[0]?.value,
           pullSecret: pullSecret,
+          highAvailabilityMode: 'Full',
         }}
         validationSchema={validationSchema}
         onSubmit={handleSubmit}
@@ -130,6 +133,7 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versio
                       </Text>
                     </TextContent>
                     <InputField ref={nameInputRef} label="Cluster Name" name="name" isRequired />
+                    {!ocmClient && <SingleNodeCheckbox name="highAvailabilityMode" />}
                     <SelectField
                       label="OpenShift Version"
                       name="openshiftVersion"
--- a/src/components/ui/DetailList.tsx
+++ b/src/components/ui/DetailList.tsx
@@ -26,6 +26,7 @@ export type DetailItemProps = {
       }[]
     | React.ReactNode;
   idPrefix?: string;
+  isHidden?: boolean;
 };
 
 export const DetailList: React.FC<DetailListProps> = ({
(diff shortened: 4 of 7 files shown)


src/components/clusters/NewClusterPage.tsx
```diff
@@ -49,6 +51,10 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versio
     nameInputRef.current?.focus();
   }, []);
 
+  const [highAvailabilityMode, setHighAvailabilityMode] = React.useState<
```
mareklibra, line 54: I think the state can be managed as any other Formik's value.

irosenzw: it's a bit problematic with Formik since formik's checkbox only have boolean values while in this case we need a string ('Full' or 'None') as a value.
moreover, since the values ([here](https://github.com/openshift-assisted/assisted-ui-lib/pull/421/files/cbbb5598b40f99d92a0f01b856ce3f9736beac49#diff-7ef7e9bedb8f975ebe0cde62f30576ce0adf82924b76e97cc314259a4953a453R69)) are `ClusterCreateParams` `highAvailabiltyMode` cannot be boolean.

jtomasek: I agree with Marek that the field should be managed by Formik as any other field. There are 2 options:

1/ Use existing `CheckboxField` component and use it to `setFieldValue` as we do for example in `NetworkConfiguration` compnent to toggle RHDnsService.

2/ Define custom `SingleNodeCheckbox` component which will be very similar to `Checkbox` component but it will have custom `isChecked` and `onChange` use field's `setValue` formik helper to set it's value.

Second option is IMHO better as it requires only one new form value.

jtomasek: Without having the value tracked by formik form's 'touched' and other helpers won't work as they won't take the single node value in account.

jtomasek (changes requested):

osherdp:
Boolean of "create single node cluster" doesn't make sense to me
I think it should be availability mode, and you can choose between full availability and single-node/standalone

src/components/clusters/NewClusterPage.tsx
```diff
@@ -130,6 +132,10 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versio
                       </Text>
                     </TextContent>
                     <InputField ref={nameInputRef} label="Cluster Name" name="name" isRequired />
+                    <SingleNodeCheckbox
+                      name="highAvailabilityModeChkbox"
+                      label="Create Single-Node Cluster"
```
omertuc, line 137: Feels like an `a` is missing, i.e. `"Create a Single-Node Cluster"`. 

Also not sure about the capitalization, maybe `"Create a single-node cluster"` ?

WDYT?

irosenzw: > Feels like an `a` is missing, i.e. `"Create a Single-Node Cluster"`.
> 
> Also not sure about the capitalization, maybe `"Create a single-node cluster"` ?
> 
> WDYT?

Good point.

irosenzw:
> Boolean of "create single node cluster" doesn't make sense to me
> I think it should be availability mode, and you can choose between full availability and single-node/standalone

Honestly, I haven't dedicated too much thought to name this checkbox.
@mareklibra @jtomasek @nirfarkas @andybraren any ideas?

osherdp:
> > Boolean of "create single node cluster" doesn't make sense to me
> > I think it should be availability mode, and you can choose between full availability and single-node/standalone
> 
> Honestly, I haven't dedicated too much thought to name this checkbox.
> @mareklibra @jtomasek @nirfarkas @andybraren any ideas?

I mean, I'm not sure that a checkbox is right here. Either radio-buttons (probably preferred) or dropdown

irosenzw:
> > > Boolean of "create single node cluster" doesn't make sense to me
> > > I think it should be availability mode, and you can choose between full availability and single-node/standalone
> > 
> > 
> > Honestly, I haven't dedicated too much thought to name this checkbox.
> > @mareklibra @jtomasek @nirfarkas @andybraren any ideas?
> 
> I mean, I'm not sure that a checkbox is right here. Either radio-buttons (probably preferred) or dropdown

Dropdown/Radio buttons for 2 options it's just bad UI/UX IMO as you can have the same result in 1 checkbox.
We can have a tooltip/popover that explains the meaning behind it if the label isn't self-explanatory enough.

osherdp:
> > > > Boolean of "create single node cluster" doesn't make sense to me
> > > > I think it should be availability mode, and you can choose between full availability and single-node/standalone
> > > 
> > > 
> > > Honestly, I haven't dedicated too much thought to name this checkbox.
> > > @mareklibra @jtomasek @nirfarkas @andybraren any ideas?
> > 
> > I mean, I'm not sure that a checkbox is right here. Either radio-buttons (probably preferred) or dropdown
> 
> Dropdown/Radio buttons for 2 options it's just bad UI/UX IMO as you can have the same result in 1 checkbox.
> We can have a tooltip/popover that explains the meaning behind it if the label isn't self-explanatory enough.
> 
> 

Availability mode is enum, right now containing two modes but not necessarily will stay this way in the future

src/components/ui/DetailList.tsx
```diff
@@ -41,32 +42,40 @@ export const DetailList: React.FC<DetailListProps> = ({
...
-  </>
-);
+export const DetailItem: React.FC<DetailItemProps> = ({
+  title,
+  value = '',
+  idPrefix,
+  isHidden = false,
+}) =>
+  isHidden ? (
+    <></>
```
mareklibra, line 52: Can we return `null` instead?

src/components/clusterConfiguration/BaremetalInventory.tsx
```diff
@@ -45,7 +45,9 @@ const BaremetalInventory: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
           Hosts connected to the internet will be inspected and automatically appear below.{' '}
           <HostsNotShowingLink setDiscoveryHintModalOpen={setDiscoveryHintModalOpen} />
         </Text>
-        <HostRequirements ContentComponent={HostRequirementsContent} />
+        {cluster.highAvailabilityMode === 'Full' && (
```
mareklibra, line 48: nit: `!isSingleNodeCluster(cluster)` ?

mareklibra (changes requested):

src/components/ui/formik/SingleNodeCheckbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const SingleNodeCheckbox: React.FC<CheckboxFieldProps> = ({
+  label,
+  helperText,
+  onChange,
+  validate,
+  idPostfix,
+  ...props
+}) => {
+  const [field] = useField({ name: props.name, validate });
```
jtomasek, line 17:
```suggestion
  const [field, meta, helpers] = useField<'None' | 'Full'>({ name: props.name, validate });
 
   const { value } = meta;
   const { setValue } = helpers;
```

src/components/ui/formik/SingleNodeCheckbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const fieldId = getFieldId(props.name, 'input', idPostfix);
+
+  const { setFieldValue } = useFormikContext<ClusterCreateParams>();
```
jtomasek, line 20: not needed

src/components/ui/formik/SingleNodeCheckbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Checkbox
+      {...field}
+      {...props}
+      id={fieldId}
+      label={label}
```
jtomasek, line 27: Since this is custom field just for Single node option the label and helperText can be defined directly here and simplified. No need to pass them as props

src/components/ui/formik/SingleNodeCheckbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+      aria-describedby={`${fieldId}-helper`}
+      description={
+        typeof helperText === 'string' ? (
+          helperText
+        ) : (
+          <HelperText fieldId={fieldId}>{helperText}</HelperText>
+        )
+      }
+      isChecked={field.value}
```
jtomasek, line 36: something like:
```suggestion
      isChecked={value === 'None'}
```

src/components/ui/formik/SingleNodeCheckbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+      onChange={(value, event) => {
+        setFieldValue('highAvailabilityMode', value ? 'None' : 'Full');
+        field.onChange(event);
+        onChange && onChange(value, event);
+      }}
```
jtomasek, line 41:
```suggestion
      onChange={(checked, event) => {
        setValue(checked ? 'None' : 'Full');
        field.onChange(event);
      }}
```

src/components/clusters/NewClusterPage.tsx
```diff
@@ -130,6 +132,11 @@ const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versio
                       </Text>
                     </TextContent>
                     <InputField ref={nameInputRef} label="Cluster Name" name="name" isRequired />
+                    <SingleNodeCheckbox
```
jtomasek, line 135: This checkbox needs to be rendered only on standalone. We don't want to expose this option in OCM yet. 

```
  if (!ocmClient) {
...
```

jtomasek (changes requested):
Few comments in Formik usage.

I think it would be good to create an initial PR just with adding the checkbox and then have the discussion with the backend on how we need to update other UI parts. Hiding the Network fields etc may cover problems which need to be solved properly.

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -83,7 +83,9 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
           isRequired
         />
       )}
-      <BasicNetworkFields cluster={cluster} hostSubnets={hostSubnets} />
+      {!isSingleNodeCluster(cluster) && (
```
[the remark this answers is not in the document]
irosenzw: the SNO selects the subnets for the user - the user doesn't need to select a subnet at this point.
In the future, from what I understood, the user will have to option to do so.

jtomasek (approved):
Thanks!

mareklibra (approved):

PR #423, opened 2021-01-27
OCPBUGSM-23658 Ntp action in host detail property

jtomasek:
Enable setting NTP source from NtpValidationStatus
    
- Move host dialogs into a state provided through context to enable triggering the dialog from different places
- Convert the dialogs state to useReducer and improve the API

Peek 2021-02-02 09-17

### DIFF
--- a/src/components/hosts/HostDialogsContext.tsx
+++ b/src/components/hosts/HostDialogsContext.tsx
@@ -1,0 +1,88 @@
+import React from 'react';
+import { Host, Inventory } from '../../api/types';
+import dialogsReducer, {
+  openDialog as openDialogAction,
+  closeDialog as closeDialogAction,
+} from '../../features/dialogs/dialogsSlice';
+
+type HostIdAndHostname = {
+  hostId: Host['id'];
+  hostname: string;
+};
+
+type EditHostProps = {
+  host: Host;
+  inventory: Inventory;
+};
+
+type DialogsDataTypes = {
+  eventsDialog: HostIdAndHostname;
+  editHostDialog: EditHostProps;
+  deleteHostDialog: HostIdAndHostname;
+  resetHostDialog: HostIdAndHostname;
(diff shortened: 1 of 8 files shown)


src/components/hosts/HostDialogsContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+      open: (data: DialogsDataTypes[typeof dialogId]) =>
+        getOpenDialog<DialogsDataTypes[typeof dialogId]>(dialogId)(data),
+      close: () => getCloseDialog(dialogId)(),
+      data: dialogsState[dialogId],
+    };
+    return context;
+  }, {});
+
+  return (
+    <HostDialogsContext.Provider value={context as HostDialogsContextType}>
```
mareklibra, line 74: nit: it would be safer for future maintenance to ensure that `context` is already of `HostDialogsContextType` type, especially since it is calculated.

jtomasek: I can do the type conversion right after the reduce. Would that be sufficient? If I add the type to context definition, the initial value in `reduce` (`{}`) is required to match the `HostDialogsContextType` which is not what we want to do.

mareklibra: That would not bring the intended help.
We can add `?` at line 34, that is safe enough since we are passing through all `dialogIds` here.
Or provide an inital default prior calling `reduce`.
Or do something like that round the line 63:
```
const myTemp: NewTypeForSingleDialogData = {
  isOpen: !!dialogsState[dialogId],
  ...
};
context[dialogId] = myTemp;
```

mareklibra: I mean ^^^ as a code quality improvement, it's not a must.
The computation of context is not the simplest one and future maintenance of it could lead to issues which can be discovered by type-checking.

mareklibra (approved):

PR #426, opened 2021-01-28
MGMT-3233 Wizard cluster details step

jtomasek:
Followups:
- Align DetailsStep to use ClusterWizardToolbar
- Resolve pull-secret editing logic
- Enable wizard on NewClusterPage


src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -0,0 +1,273 @@
...
+    try {
+      const params = _.omit(values, ['useRedHatDnsService']);
+
+      if (pullSecretSet) {
+        delete params.pullSecret;
+      }
+
+      const { data } = await patchCluster(cluster.id, params);
+      dispatch(updateCluster(data));
+      setCurrentStepId('cluster-configuration');
```
mareklibra, line 113: I think we should check relevant backend validations again before transitioning

src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -0,0 +1,273 @@
...
+                    submitForm();
+                  }
+                }}
+              >
+                <InputField ref={nameInputRef} label="Cluster Name" name="name" isRequired />
+                <SelectField
+                  label="OpenShift Version"
+                  name="openshiftVersion"
+                  options={[{ label: openshiftVersion, value: openshiftVersion }]}
+                  // getHelperText={getOpenshiftVersionHelperText}
```
mareklibra, line 159: Why is this commented out?

src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -0,0 +1,273 @@
...
+          <Stack hasGutter>
+            {!!alerts.length && (
+              <StackItem>
+                <Alerts />
+              </StackItem>
+            )}
+            <StackItem>
+              <ClusterToolbar>
+                <ToolbarButton
+                  name="save"
```
mareklibra, line 207: This logic + component composition are repeating on every step, so they are co-located in the `ClusterWizardToolbar`.

src/components/clusterWizard/ClusterWizard.tsx
```diff
@@ -17,11 +18,15 @@ const ClusterWizard: React.FC<ClusterWizardProps> = ({ cluster }) => {
 
   const renderCurrentStep = React.useCallback(() => {
     switch (currentStepId) {
-      //case CLUSTER_STEP_DETAIL:
+      case 'cluster-details':
+        return <ClusterDetails cluster={cluster} />;
       // TODO(mlibra): use different component!!
```
mareklibra, line 23: This TODO can be already removed

mareklibra (changes requested):

src/components/clusterWizard/ClusterDetails.tsx
jtomasek: Agreed, however I am not sure if there any that we should check for this step?

src/components/clusterWizard/ClusterDetails.tsx
jtomasek: At this point the field is disabled, I am going to update this in a follow up when I mix this with NewClusterPage functionality. Also I think I'll need to fetch the versions here and try to compare to what is set on the cluster.

src/components/clusterWizard/ClusterDetails.tsx
jtomasek: Yes, I'll conform to it in a follow up.

src/components/clusterWizard/ClusterWizard.tsx
jtomasek: done

src/components/clusterWizard/ClusterDetails.tsx
mareklibra: Just for consistency. If nothing else to have it among the transition functions, so we can think about all of them together.
Can be follow-up.

mareklibra:
Follow-ups:
- use of `getOpenshiftVersionHelperText`
- validation before `setCurrentStepId('cluster-configuration');`
- use of `ClusterWizardToolbar` in `ClusterDetails`

PR #436, opened 2021-02-02
MGMT-3824 Wizard transition functions

mareklibra:
TODO:
- [ ] find a way to be notified about changes in the list of backend validations
- [x] verify
- [x] click-through navigation to steps to fix validations (backwards only)
- [ ] empty baseDomain should not pass validation

### DIFF
--- a/src/components/clusterWizard/ClusterWizardStep.tsx
+++ b/src/components/clusterWizard/ClusterWizardStep.tsx
@@ -12,7 +12,6 @@ import {
   canNextNetworkBackend,
   ClusterWizardStepsType,
 } from './wizardTransition';
-import './ClusterWizardStep.css';
 
 type ClusterWizardStepProps = {
   cluster: Cluster;
@@ -27,12 +26,11 @@ const wizardSteps: ClusterWizardStepsType[] = [
 ];
 
 const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = ({
-  content,
-  isDisabled,
-  isCurrent,
   isValid = () => true,
-  ...rest
+  ...props
 }) => {
+  const { content, isDisabled, isCurrent } = props;
+
   let validatedLinkName = content;
   if (!isDisabled && !isCurrent && !isValid()) {
     validatedLinkName = (
@@ -47,14 +45,7 @@ const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = ({
     );
   }
 
-  return (
-    <WizardNavItem
-      content={validatedLinkName}
-      isDisabled={isDisabled}
-      isCurrent={isCurrent}
-      {...rest}
-    />
-  );
+  return <WizardNavItem {...props} content={validatedLinkName} />;
 };
 
 const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ cluster, footer, children }) => {
@@ -66,7 +57,7 @@ const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ cluster, footer, 
         key="cluster-details"
         content="Cluster Details"
         isCurrent={currentStepId === 'cluster-details'}
-        isValid={() => !cluster || canNextClusterDetails({ cluster })}
+        isValid={() => canNextClusterDetails({ cluster })}
         isDisabled={false}
         step={0}
         onNavItemClick={() => setCurrentStepId('cluster-details')}
--- a/src/components/clusterWizard/ClusterWizardStep.tsx
+++ b/src/components/clusterWizard/ClusterWizardStep.tsx
@@ -1,11 +1,21 @@
 import React from 'react';
-import { WizardBody, WizardNav, WizardNavItem } from '@patternfly/react-core';
+import { WizardBody, WizardNav, WizardNavItem, WizardNavItemProps } from '@patternfly/react-core';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Wizard/wizard';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
+import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens';
+import { Cluster } from '../../api/types';
 import ClusterWizardContext from './ClusterWizardContext';
-import { ClusterWizardStepsType } from './wizardTransition';
+import {
+  canNextBaremetalDiscovery,
+  canNextClusterDetails,
+  canNextNetworkBackend,
+  ClusterWizardStepsType,
+} from './wizardTransition';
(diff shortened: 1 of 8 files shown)


jtomasek (changes requested):

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -8,14 +15,108 @@ export type ClusterWizardStepsType =
...
+      clusterValidationsInfo[group].find((validation) => id === validation.id)?.status ===
+      'success',
+  );
+
+const allClusterValidations = (
+  clusterValidationsInfo: ClusterValidationsInfo,
+  group: 'hostsData' | 'network' | 'configuration',
+) => clusterValidationsInfo[group]?.every((validation) => validation.status === 'success');
+
+const selectedHostValidations = (
```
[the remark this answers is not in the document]
mareklibra: Sure, we can flatten. Anyway, the received validations are "an object containing arrays". So to save one 2-value enum parameter, we need to do additional sequence of manupilations internally. I do not consider that as a benefit. If nothing else, there's performance penalty.
In any case, I think it is good to fail even if the required validation is not found in the input (I know, it should be `pending`, but what if backend processing changes in the future ...)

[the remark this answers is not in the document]
mareklibra: Renaming the function

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -8,14 +15,108 @@ export type ClusterWizardStepsType =
...
+  hostValidationsInfo: HostValidationsInfo,
+  group: 'hardware' | 'network',
+  selectedIds: HostValidationId[],
+): boolean =>
+  !selectedIds.every(
+    (id) =>
+      hostValidationsInfo[group]?.find((validation) => id === validation.id)?.status === 'success',
+  );
+
+const allHostValidations = (
```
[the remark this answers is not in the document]
mareklibra: ok

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -8,14 +15,111 @@ export type ClusterWizardStepsType =
...
+
+const checkHostValidations = (
+  hostValidationsInfo: HostValidationsInfo,
+  group: 'hardware' | 'network',
+  requiredIds: HostValidationId[],
+): boolean =>
+  requiredIds.every((id) => {
+    const validation = hostValidationsInfo[group]?.find((validation) => id === validation.id);
+    return validation?.status === 'success';
+  });
```
jtomasek, line 47: Following up on my previous comment, my intention was to remove the requirement to specify the group to allow running this function only once when we need to check validations picked from several groups which can happen on subsequent steps.

We are already flattening the validations info on some places:
https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/components/clusterConfiguration/ClusterValidationSection.tsx#L48

So for our case it would be:
```suggestion
  {
    const requiredValidations = _.values(hostValidationsInfo).flat().filter(v => requiredIds.includes(v.id))
    return requiredValidations.every(v => v.status === 'success);
    }
```

jtomasek (changes requested):

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -16,36 +26,70 @@ const wizardSteps: ClusterWizardStepsType[] = [
   'review',
 ];
 
-const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ footer, children }) => {
+const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = ({
+  content,
+  isDisabled,
+  isCurrent,
+  isValid = () => true,
```
jtomasek, line 33: Any reason why `isValid` is not a boolean?

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -16,36 +26,70 @@ const wizardSteps: ClusterWizardStepsType[] = [
...
+  ...rest
+}) => {
+  let validatedLinkName = content;
+  if (!isDisabled && !isCurrent && !isValid()) {
+    validatedLinkName = (
+      <>
+        <ExclamationTriangleIcon
+          className="wizard-nav-item-warning-icon"
```
jtomasek, line 41: seems not to be used

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -16,36 +26,70 @@ const wizardSteps: ClusterWizardStepsType[] = [
...
+    );
+  }
+
+  return (
+    <WizardNavItem
+      content={validatedLinkName}
+      isDisabled={isDisabled}
+      isCurrent={isCurrent}
+      {...rest}
+    />
```
jtomasek, line 56:
```suggestion
    <WizardNavItem {...props} content={validatedLinkContent}/>
```

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -16,36 +26,70 @@ const wizardSteps: ClusterWizardStepsType[] = [
...
 ];
 
-const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ footer, children }) => {
+const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = ({
+  content,
+  isDisabled,
+  isCurrent,
+  isValid = () => true,
+  ...rest
+}) => {
```
jtomasek, line 35:
```suggestion
const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = (props) => {
  {
    content,
    isDisabled,
    isCurrent,
    isValid = () => true
} = props;
```

jtomasek (changes requested):

src/components/clusterWizard/ClusterWizardStep.tsx
mareklibra: Just performance reason, to postpone validation function call.

src/components/clusterWizard/ClusterWizardStep.tsx
mareklibra: yep, not any more

src/components/clusterWizard/ClusterWizardStep.tsx
mareklibra: done in a similar way

src/components/clusterWizard/wizardTransition.ts
mareklibra: ok, done.
Adding an extra check for the case that the input does not contain the required validation at all.

jtomasek (approved):

PR #438, opened 2021-02-03
MGMT-2632 Adds missing warning

jkilzi:
- Adds a missing warning when the cluster is in `installing-pending-user-action` (e. g. wrong boot order).


jtomasek (approved):
Nice catch

mareklibra (approved):

PR #439, opened 2021-02-03
Introduce feature gates

mareklibra:
First usage: enable/disable Single Node OpenShift

Related PRs:
- https://github.com/openshift-assisted/assisted-ui/pull/395
- OCM 2117

### DIFF
--- a/src/components/clusters/NewClusterPage.tsx
+++ b/src/components/clusters/NewClusterPage.tsx
@@ -34,9 +34,9 @@ import { useOpenshiftVersions } from '../fetching/openshiftVersions';
 import { OpenshiftVersionOptionType } from '../../types/versions';
 import SingleNodeCheckbox from '../ui/formik/SingleNodeCheckbox';
 import {
-  FeatureGateContext,
   FeatureGateContextProvider,
   FeatureListType,
+  useFeature,
 } from '../../features/featureGate';
 
 import './NewClusterPage.css';
@@ -48,7 +48,7 @@ type NewClusterFormProps = {
 
 const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versions }) => {
   const { addAlert, clearAlerts } = React.useContext(AlertsContext);
-  const { isFeatureEnabled } = React.useContext(FeatureGateContext);
+  const isSingleNodeOpenshiftEnabled = useFeature('ASSISTED_INSTALLER_SNO_FEATURE');
   const history = useHistory();
 
   const nameInputRef = React.useRef<HTMLInputElement>();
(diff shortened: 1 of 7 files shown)


src/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -121,7 +121,7 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
     });
 
     let nextFilters = setNextSelectedValues(type, isChecked, value, filters);
-    nextFilters = setNextSelectAllValue(cluster.hosts?.length ?? 0, nextFilters);
+    nextFilters = setNextSelectAllValue(cluster.hosts?.length || 0, nextFilters);
```
mareklibra, line 124: Our development build does not transpile this operator and OCM is not ready for it.

irosenzw: +1. 
bumped into that too which working against uhc-portal

src/features/featureGate.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+// Must conform Unleash constants
+type AssistedInstallerFeatureType = 'ASSISTED_INSTALLER_SNO_FEATURE';
```
irosenzw, line 4: nit: do we need the `ASSISTED_INSTALLER` prefix ? 
can we just call it `SNO_FEATURE` or if we do need that prefix: `AI_SNO_FEATURE`

src/features/featureGate.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const FeatureGateContext = React.createContext<FeatureGateContextType>({
+  isFeatureEnabled: () => false,
+});
+
+export const FeatureGateContextProvider: React.FC<{
+  features: FeatureListType;
+}> = ({ features, children }) => {
+  const isFeatureEnabled = (feature: AssistedInstallerFeatureType) => {
+    // return feature === 'ASSISTED_INSTALLER_SNO_FEATURE' && window.location.hostname.includes('qaprodauth.');
```
irosenzw, line 31: leftover ?

src/features/featureGate.tsx
mareklibra: The same constant is added to OCM as well. I would keep the names same to simplify `git grep` by future maintainers.

src/features/featureGate.tsx
mareklibra: thanks, removed

src/components/clusters/NewClusterPage.tsx
```diff
@@ -44,6 +48,7 @@ type NewClusterFormProps = {
 
 const NewClusterForm: React.FC<NewClusterFormProps> = ({ pullSecret = '', versions }) => {
   const { addAlert, clearAlerts } = React.useContext(AlertsContext);
+  const { isFeatureEnabled } = React.useContext(FeatureGateContext);
```
jtomasek, line 51:
```
const isSingleNodeEnabled = useFeature('ASSISTED_INSTALLER_SNO_FEATURE')
```
would be a nicer API.

this hook could be exported from FeatureGateContext module. See HostDialogsContext.tsx for simmilar solution.

jtomasek (changes requested):

jtomasek (approved):

jtomasek (approved):

PR #440, opened 2021-02-04
MGMT-3234 Convert NewClusterPage to use wizard

jtomasek:
- NewClusterPage now renders NewClusterWizard instead of NewClusterForm
- ClusterDetails wizard step is capable of creating/updating the cluster
- single node, openshift version and pull secret are not allowed to be
  updated

Next steps:
- use ClusterWizardToolbar in ClusterDetails
- validate cluster name + dns domain for uniqueness rather than just cluster name
- explore possibility to edit pull secret?
- align to wireframes

### DIFF
--- a/src/components/clusterWizard/ClusterWizardStep.tsx
+++ b/src/components/clusterWizard/ClusterWizardStep.tsx
@@ -14,7 +14,7 @@ import {
 } from './wizardTransition';
 
 type ClusterWizardStepProps = {
-  cluster: Cluster;
+  cluster?: Cluster;
   footer?: React.ReactNode;
 };
 
@@ -57,7 +57,7 @@ const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ cluster, footer, 
         key="cluster-details"
         content="Cluster Details"
         isCurrent={currentStepId === 'cluster-details'}
-        isValid={() => canNextClusterDetails({ cluster })}
+        isValid={() => !cluster || canNextClusterDetails({ cluster })}
         isDisabled={false}
         step={0}
         onNavItemClick={() => setCurrentStepId('cluster-details')}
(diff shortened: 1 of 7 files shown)


src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -57,7 +65,7 @@ const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ cluster, footer,
         key="cluster-details"
         content="Cluster Details"
         isCurrent={currentStepId === 'cluster-details'}
-        isValid={() => canNextClusterDetails({ cluster })}
+        isValid={validWithoutCluster(canNextClusterDetails, cluster)}
```
mareklibra, line 68:
```suggestion
        isValid={() => !cluster || canNextClusterDetails({ cluster })}
```

mareklibra (changes requested):

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -14,9 +14,9 @@ export type ClusterWizardStepsType =
   | 'networking'
   | 'review';
 
-export const CLUSTER_WIZARD_FIRST_STEP: ClusterWizardStepsType = 'cluster-details';
+export const CLUSTER_WIZARD_FIRST_STEP: ClusterWizardStepsType = 'baremetal-discovery';
```
mareklibra, line 17: This will not work for the "edit" flow.
I will try to fix it in a patch I am currently stacking on top of that PR.

src/components/clusterWizard/ClusterWizardStep.tsx
jtomasek: done

src/components/clusterWizard/wizardTransition.ts
jtomasek: We could indicate the desired step via means of react router (state or query).

mareklibra (approved):

PR #446, opened 2021-02-09
Make wizard first step dynamic based on current flow

mareklibra:
Depends on #440

### DIFF
--- a/src/components/clusterWizard/ClusterDetails.tsx
+++ b/src/components/clusterWizard/ClusterDetails.tsx
@@ -150,7 +150,7 @@ const ClusterDetailsForm: React.FC<ClusterDetailsFormProps> = (props) => {
 
     try {
       const { data } = await postCluster(params);
-      history.push(`${routeBasePath}/clusters/${data.id}`);
+      history.push(`${routeBasePath}/clusters/${data.id}?flow=new`);
     } catch (e) {
       handleApiError<ClusterCreateParams>(e, () =>
         addAlert({ title: 'Failed to create new cluster', message: getErrorMessage(e) }),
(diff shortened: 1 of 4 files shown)


src/components/clusterWizard/wizardTransition.ts
```diff
@@ -14,7 +14,10 @@ export type ClusterWizardStepsType =
   | 'networking'
   | 'review';
 
-export const CLUSTER_WIZARD_FIRST_STEP: ClusterWizardStepsType = 'baremetal-discovery';
+export const getClusterWizardFirstStep = (): ClusterWizardStepsType => {
+  const searchParams = new URLSearchParams(window.location.search);
```
mareklibra, line 18: Another approach can be via react-router, either by `useLocation()` or by parsing query params at the top of component hierarchy (close to Router). The so far proposed solution requires minimum code changes.

src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -147,7 +147,7 @@ const ClusterDetailsForm: React.FC<ClusterDetailsFormProps> = (props) => {
 
     try {
       const { data } = await postCluster(params);
-      history.push(`${routeBasePath}/clusters/${data.id}`);
+      history.push(`${routeBasePath}/clusters/${data.id}?flow=new`);
```
jtomasek, line 150: to avoid exposing the logic to the user we could use `location.state`
```suggestion
      history.push(`${routeBasePath}/clusters/${data.id}`, { wizardFlow: 'new' });
```

src/components/clusterWizard/ClusterWizardContext.tsx
```diff
@@ -1,13 +1,13 @@
...
+import { ClusterWizardStepsType, getClusterWizardFirstStep } from './wizardTransition';
 
 type ClusterWizardContextType = {
   currentStepId: ClusterWizardStepsType;
   setCurrentStepId: (stepId: ClusterWizardStepsType) => void;
 };
 
 const ClusterWizardContext = React.createContext<ClusterWizardContextType>({
-  currentStepId: CLUSTER_WIZARD_FIRST_STEP,
+  currentStepId: getClusterWizardFirstStep(),
```
jtomasek, line 10: IIUC we could put just 'cluster-details' here as this value is only used in case when context provider is not rendered.

jtomasek (approved):

mareklibra:
Rebased only

PR #451, opened 2021-02-11
BZ-1877838 CIDR fields display wrong validation message

jkilzi:
Previously the message bellow couldn't be displayed because we were validating the input is a valid IP address (v4 or v6) but not a valid CIDR string.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
   "dependencies": {
     "axios-case-converter": "^0.6.0",
     "ip-address": "^7.1.0",
+    "is-cidr": "^4.0.2",
     "is-in-subnet": "^3.1.0",
     "unique-names-generator": "^4.2.0"
   }
--- a/src/components/ui/formik/validationSchemas.ts
+++ b/src/components/ui/formik/validationSchemas.ts
@@ -6,6 +6,7 @@ import { NO_SUBNET_SET } from '../../../config/constants';
 import { trimCommaSeparatedList, trimSshPublicKey } from './utils';
 import { Address4, Address6 } from 'ip-address';
 import { isInSubnet } from 'is-in-subnet';
+import isCIDR from 'is-cidr';
 
 const CLUSTER_NAME_REGEX = /^[a-z]([-a-z0-9]*[a-z0-9])?$/;
 const SSH_PUBLIC_KEY_REGEX = /^(ssh-rsa|ssh-ed25519|ecdsa-[-a-z0-9]*) AAAA[0-9A-Za-z+/]+[=]{0,3}( .+)?$/;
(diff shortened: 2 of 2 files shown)


nirfarkas:
Thanks @jkilzi !
One nit:
Instead of: "Examples:" --> "For example:".

jkilzi:
I've updated the message as per Nir's request

jkilzi:
Added additional validations, please take a look:

ezgif com-gif-maker

package.json
```diff
@@ -82,6 +82,7 @@
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^2.5.0",
     "install-peers-cli": "^2.2.0",
+    "is-cidr": "^4.0.2",
```
mareklibra, line 85: Move to `dependencies`, please

mareklibra (changes requested):

package.json
jkilzi: @mareklibra  same issue here with microbundle as [here](https://github.com/openshift-assisted/assisted-ui-lib/pull/455#discussion_r576960873)

mareklibra: indeed :-)

jtomasek (changes requested):
Please split the dependency update into a separate commit for easier wizard branch rebase.

yarn.lock
```diff
@@ -5683,20 +5685,6 @@ http-proxy-middleware@0.19.1:
   version "0.19.1"
   resolved "https://registry.yarnpkg.com/http-proxy-middleware/-/http-proxy-middleware-0.19.1.tgz#183c7dc4aa1479150306498c210cdaf96080a43a"
   integrity sha512-yHYTgWMQO8VvwNS22eLLloAkvungsKdKTLO8AJlftYIKNfJr3GK3zK0ZCfzDDGUBttdGc8xFy1mCitvNKQtC3Q==
-  dependencies:
-    http-proxy "^1.17.0"
-    is-glob "^4.0.0"
-    lodash "^4.17.11"
-    micromatch "^3.1.10"
-
-http-proxy@^1.17.0:
```
mareklibra, line 5692: Why is that? I just tried `git add is-cidr` locally and it did not remove me anything from here.

jkilzi: Sorry, why did you expect it to remove anything?

mareklibra: I did not but it is being removed by this PR

jkilzi: @mareklibra  also this dependency has been removed `follow-redirects@^1.0.0`. Do you want me to revert the `yarn.lock`? Please let me know if you want me to do something else.

jkilzi:
OK, I made the following here:
1. Excluded the yarn.lock completely from the PR (I suspect that there are some leftovers there though...)
2. Squashed all the changes into one commit

mareklibra (approved):

PR #453, opened 2021-02-12
Introduce wizard step validation mapping

jtomasek:
- Use validation map to calculate cluster/host validationInfo and
  status
- Use calculated status to determine transition to next step
- Adjust ClusterWizardToolbar to display validations toggle properly
- Simplify 'canNext' functions
- Enable 'Next' button in case when form is dirty and valid to allow
  user saving the changes regardless of backend validations

### DIFF
--- a/src/components/clusterConfiguration/ClusterValidationSection.tsx
+++ b/src/components/clusterConfiguration/ClusterValidationSection.tsx
@@ -16,10 +16,12 @@ import {
 } from '@patternfly/react-core';
 import { TimesIcon } from '@patternfly/react-icons';
 import { Cluster } from '../../api/types';
-import { NetworkConfigurationValues, ValidationsInfo } from '../../types/clusters';
+import { NetworkConfigurationValues, Validation, ValidationsInfo } from '../../types/clusters';
 import { CLUSTER_FIELD_LABELS } from '../../config/constants';
 import { stringToJSON } from '../../api/utils';
 import './ClusterValidationSection.css';
+import { getWizardStepClusterValidationsInfo } from '../clusterWizard/wizardTransition';
+import ClusterWizardContext from '../clusterWizard/ClusterWizardContext';
 
 type ClusterValidationSectionProps = {
   cluster: Cluster;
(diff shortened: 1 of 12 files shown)


src/components/clusterConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -164,10 +164,10 @@ const NetworkConfigurationForm: React.FC<{
             <StackItem>
               <ClusterWizardToolbar
                 cluster={cluster}
-                errors={errors}
+                formErrors={errors}
                 dirty={dirty}
                 isSubmitting={isSubmitting}
-                isNextDisabled={!canNextNetwork({ isValid, isSubmitting, cluster })}
+                isNextDisabled={!(isValid && dirty) && !canNextNetwork({ cluster })}
```
mareklibra, line 170: If the form is dirty and valid, it should be possible to submit it no matter of former validation state.

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -23,8 +24,73 @@ export const getClusterWizardFirstStep = (
...
+  host: {
+    groups: [],
+    validationIds: [],
+  },
+};
+
+const baremetalDiscoveryStepValidationsMap: WizardStepValidationMap = {
+  cluster: {
+    groups: [],
+    validationIds: [/*'sufficient-masters-count',*/ 'ocs-requirements-satisfied'],
```
mareklibra, line 58: please reference https://bugzilla.redhat.com/show_bug.cgi?id=1928086 here

src/components/clusterConfiguration/ClusterValidationSection.tsx
```diff
@@ -93,7 +97,7 @@ const ClusterValidationSection: React.FC<ClusterValidationSectionProps> = ({
                 <FlexItem>
                   <List>
                     {failedValidations.map((validation) => (
-                      <ListItem key={validation.id}>{validation.message}</ListItem>
+                      <ListItem key={validation.id}>{validation?.message}</ListItem>
```
mareklibra, line 100: not needed

mareklibra (changes requested):

src/components/clusterConfiguration/ClusterValidationSection.tsx
jtomasek: done

src/components/clusterWizard/wizardTransition.ts
jtomasek: done

mareklibra (approved):

PR #455, opened 2021-02-15
BZ-1917396 Display sizes as per IEC standard

jkilzi:
- Fixes the memory and disk size display units as per IEC standard

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
   ],
   "dependencies": {
     "axios-case-converter": "^0.6.0",
+    "filesize.js": "^2.0.0",
     "ip-address": "^7.1.0",
     "is-cidr": "^4.0.2",
     "is-in-subnet": "^3.1.0",
--- a/src/components/AddBareMetalHosts/BareMetalInventoryAddHost.tsx
+++ b/src/components/AddBareMetalHosts/BareMetalInventoryAddHost.tsx
@@ -14,8 +14,8 @@ import { AddBareMetalHostsContext } from './AddBareMetalHostsContext';
 
 const HostRequirementsContent = ({ worker = {} }: { worker?: HostRequirementsType['worker'] }) => (
   <Text component="p">
-    Worker hosts must have at least {worker.cpuCores || 2} CPU cores, {worker.ramGib || 8} GB of
-    RAM, and {worker.diskSizeGb || 120} GB of filesystem storage.
+    Worker hosts must have at least {worker.cpuCores || 2} CPU cores, {worker.ramGib || 8} Gib of
+    RAM, and {worker.diskSizeGb || 120} Gib of filesystem storage.
   </Text>
 );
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -81,6 +81,7 @@
     "eslint-plugin-prettier": "^3.1.2",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^2.5.0",
+    "filesize.js": "^2.0.0",
```
mareklibra, line 84: please remove `humanize-plus` and `@types/humanize-plus`. They don't seem to be used.

mareklibra: in addition, this should be among `dependencies`

mareklibra (changes requested):
In addition to comments above, please update all `HostRequirements`-content components, they have hardcoded `GB`.

package.json
jkilzi: 1. If I put this inside the `dependencies` section then `microbundle` won't include it in the final bundle. I found an explanation on why this happens and why should I put it inside the `devDependencies` in [microbundle's official docs](https://github.com/developit/microbundle/wiki/How-Microbundle-decides-which-dependencies-to-bundle). I don't know why microbundle works in this way, it's against the standard (and I hate this the same as you do).
2. We still use `humanize-plus` in `hardwareInfo.ts`

mareklibra: 1. But this is as expected. By having it among `dependencies`, the yarn will resolve that at `yarn install` time of the consuming package (i.e. `assisted-ui`). I understand this is a chicken-egg issue at development time, but you can workaround it by
- simply trust since you have it already tested within `devDependencies`
- do tentative `yarn add` of the package within `assisted-ui` (do not commit that change there)
Once we release the library, everything will be good again.

By inlining all dependencies, our dist would be unnecessary big. And we would dictate all the transpiling rules even on the places where the consuming application build process should be used. Another disadvantage will be in no-resuse of that 3rd party code within the consuming application - consider the case that the same library is used by another dependency as well. The tree shaking will not work here.

2. You are right, for CPU speed. Let's keep it then.

jtomasek (changes requested):
Please split the dependency update into a separate commit for easier wizard branch rebase.

package.json
jkilzi: Moved `filesize.js` to dependencies here: 72026c95c77f190735363a8481f0cb31d2fca92e

yarn.lock
```diff
@@ -5683,20 +5683,6 @@ http-proxy-middleware@0.19.1:
   version "0.19.1"
   resolved "https://registry.yarnpkg.com/http-proxy-middleware/-/http-proxy-middleware-0.19.1.tgz#183c7dc4aa1479150306498c210cdaf96080a43a"
   integrity sha512-yHYTgWMQO8VvwNS22eLLloAkvungsKdKTLO8AJlftYIKNfJr3GK3zK0ZCfzDDGUBttdGc8xFy1mCitvNKQtC3Q==
-  dependencies:
```
mareklibra, line 5686: Same comment as https://github.com/openshift-assisted/assisted-ui-lib/pull/451#discussion_r577476247

mareklibra (approved):

yarn.lock
jkilzi: `yarn.lock` has been removed from the PR

PR #456, opened 2021-02-15
Wizard: Add validation sections to the Review step

mareklibra:
Depends on:
- [x] #453
- [x] #457

### DIFF
--- a/src/components/clusterConfiguration/ReviewValidations.tsx
+++ b/src/components/clusterConfiguration/ReviewValidations.tsx
@@ -20,6 +20,7 @@ import ClusterWizardContext from '../clusterWizard/ClusterWizardContext';
 import { ClusterWizardStepsType, findValidationFixStep } from '../clusterWizard/wizardTransition';
 import { wizardStepNames } from '../clusterWizard/ClusterWizardStep';
 import { Cluster, Host, stringToJSON } from '../../api';
+import { CLUSTER_VALIDATION_LABELS, HOST_VALIDATION_LABELS } from '../../config';
 
 const AllValidationsPassed = () => (
   <>
@@ -48,14 +49,18 @@ const FailingValidation: React.FC<{
   clusterGroup?: ClusterValidationGroup;
   hostGroup?: HostValidationGroup;
 }> = ({ validation, clusterGroup, hostGroup }) => {
-  const issue = `Checking of ${_.lowerFirst(validation.message).replace(/\.$/, '')}, failed. `;
+  const issue = `${
+    HOST_VALIDATION_LABELS[validation.id] ||
+    CLUSTER_VALIDATION_LABELS[validation.id] ||
+    validation.id
+  } check failed. `;
 
   let fix;
   const step = findValidationFixStep({ id: validation.id, clusterGroup, hostGroup });
   if (step) {
     fix = (
       <>
-        Fix can be done in the <ValidationActionLink step={step} /> step.
+        It can be fixed in the <ValidationActionLink step={step} /> step.
       </>
     );
   } else {
--- a/src/components/clusterConfiguration/ReviewValidations.tsx
+++ b/src/components/clusterConfiguration/ReviewValidations.tsx
@@ -1,0 +1,148 @@
+import React from 'react';
+import _ from 'lodash';
+import { Button, ButtonVariant } from '@patternfly/react-core';
+import { CheckCircleIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
+import {
+  global_success_color_100 as okColor,
+  global_danger_color_100 as dangerColor,
+} from '@patternfly/react-tokens';
+import {
+  ValidationsInfo as ClusterValidationsInfo,
+  ValidationGroup as ClusterValidationGroup,
+  Validation as ClusterValidation,
+} from '../../types/clusters';
+import {
+  ValidationsInfo as HostValidationsInfo,
+  ValidationGroup as HostValidationGroup,
+  Validation as HostValidation,
+} from '../../types/hosts';
+import ClusterWizardContext from '../clusterWizard/ClusterWizardContext';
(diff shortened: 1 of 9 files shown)


mareklibra:
---

mareklibra:
Cc @nirfarkas

src/components/clusterConfiguration/ReviewValidations.tsx
```diff
@@ -0,0 +1,148 @@
...
+    </Button>
+  );
+};
+
+const FailingValidation: React.FC<{
+  validation: HostValidation | ClusterValidation;
+  clusterGroup?: ClusterValidationGroup;
+  hostGroup?: HostValidationGroup;
+}> = ({ validation, clusterGroup, hostGroup }) => {
+  const issue = `Checking of ${_.lowerFirst(validation.message).replace(/\.$/, '')}, failed. `;
```
mareklibra, line 51: Cc @nirfarkas

mareklibra: We can tweak messages on the backend. But it would be good to agree on a generic structure to be used here.

nirfarkas: I think the warning icon should be replaced with an error icon 
image

mareklibra: ok, I will update that in left-navigation as well

mareklibra:
Rebased

src/components/clusterWizard/ClusterWizard.tsx
```diff
@@ -32,6 +32,7 @@ const ClusterWizard: React.FC<ClusterWizardProps> = ({ cluster }) => {
         return <ReviewStep cluster={cluster} />;
       case 'cluster-details':
       default:
+        // return <ReviewStep cluster={cluster} />;
```
jtomasek, line 35: leftover?

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -35,9 +42,9 @@ const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = ({
   if (!isDisabled && !isCurrent && !isValid()) {
     validatedLinkName = (
       <>
-        <ExclamationTriangleIcon
+        <ExclamationCircleIcon
```
jtomasek, line 45: Wireframes have the icon after the content https://marvelapp.com/prototype/fjj6g57/screen/76752444

src/components/clusterConfiguration/ReviewValidations.tsx
```diff
@@ -0,0 +1,148 @@
...
```
jtomasek, line 51: IMO it would be better to use HOST_VALIDATION_LABELS[validation.id] (and define labels for cluster validations too) as current approach does not work with all validation messages. e.g.:
Screenshot from 2021-02-18 15-04-23

jtomasek (changes requested):
Works great!

src/components/clusterConfiguration/ReviewValidations.tsx
mareklibra: I wanted to avoid that due to the need of subsequent maintenance. But you are right, we will probably not able to provide good wording this way.

mareklibra: done

src/components/clusterWizard/ClusterWizard.tsx
mareklibra: right

src/components/clusterWizard/ClusterWizardStep.tsx
mareklibra: done

src/components/clusterWizard/ClusterWizardStep.css
```diff
@@ -0,0 +1,7 @@
+.cluster-wizard-step .pf-c-wizard__nav-item {
+  white-space: nowrap;
```
jtomasek, line 2: This will lead to longer step names not getting wrapped and the warning icon potentially not being visible. How about just spacing the icon from the text with `&nbsp;` - that will prevent from breaking the line between the text and icon while allowing the text to break in case it is long.

src/components/clusterWizard/ClusterWizardStep.css
```diff
@@ -0,0 +1,7 @@
...
+}
+
+.wizard-nav-item-warning-icon {
+  margin-left: var(--pf-global--spacer--sm);
```
jtomasek, line 6: `&nbsp;` between the text and icon has the same effect AFAIK

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -35,12 +42,12 @@ const NavItem: React.FC<WizardNavItemProps & { isValid?: () => boolean }> = ({
   if (!isDisabled && !isCurrent && !isValid()) {
     validatedLinkName = (
       <>
-        <ExclamationTriangleIcon
+        {content}
+        <ExclamationCircleIcon
```
jtomasek, line 46:
```suggestion
        {content}&nbsp;
        <ExclamationCircleIcon
```

jtomasek (changes requested):

src/components/clusterWizard/ClusterWizardStep.css
mareklibra: Exactly. But we do not have such names now and we should prevent using them in the future.

src/components/clusterWizard/ClusterWizardStep.css
mareklibra: right, will do. But this is more PF-firendly way I think.

jtomasek (approved):

PR #459, opened 2021-02-16
Introduce ASSISTED_INSTALLER_OCS_FEATURE

mareklibra:
Defaults for features are provided, so they can be configured via Unleash without additional change in the assisted-ui-lib code.

### DIFF
--- a/src/components/Router.tsx
+++ b/src/components/Router.tsx
@@ -4,9 +4,9 @@ import { Switch, Route, Redirect } from 'react-router-dom';
 import { Clusters, ClusterPage, NewClusterPage } from './clusters';
 import { store } from '../store';
 import { isSingleClusterMode, routeBasePath } from '../config/constants';
+import { FeatureGateContextProvider, FeatureListType } from '../features/featureGate';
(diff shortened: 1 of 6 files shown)


src/components/Router.tsx
```diff
@@ -17,7 +17,7 @@ export const AssistedUiRouter: React.FC<{ allEnabledFeatures: FeatureListType }>
 );
 
 export const Router: React.FC<{ features: FeatureListType }> = ({ features, children }) => (
-  <>
+  <FeatureGateContextProvider features={features}>
```
jtomasek, line 20: This will have to be moved as we're integrating with OCM

jtomasek (approved):

src/components/Router.tsx
mareklibra: Right. In general, I prefer to handle the context somewhere high in the hierarchy than copying it to each sub-component and dealing with passing of `features` props on multiple places. I can even imagine handling of the context provider wihtin OCM code-base. Let's consider it on particular code when it's available.

PR #460, opened 2021-02-16
MGMT-4110 SNO is only available in selected OCP versions

jtomasek:
- If user enables SNO, the list of available OCP versions is reduced to 4.8 and higher
- SNO option will be hidden if there is no version which supports it available

### DIFF
--- a/src/components/clusters/NewClusterPage.tsx
+++ b/src/components/clusters/NewClusterPage.tsx
@@ -14,8 +14,6 @@ import {
 import PageSection from '../ui/PageSection';
 import { ToolbarButton } from '../ui/Toolbar';
 import ClusterToolbar from './ClusterToolbar';
-import { ExclamationTriangleIcon } from '@patternfly/react-icons';
-import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens';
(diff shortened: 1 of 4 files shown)


src/components/clusters/NewClusterPage.tsx
```diff
@@ -25,21 +23,17 @@ import { getClusters, postCluster } from '../../api/clusters';
...
 import SingleNodeCheckbox from '../ui/formik/SingleNodeCheckbox';
-import {
-  FeatureGateContextProvider,
-  FeatureListType,
-  useFeature,
-} from '../../features/featureGate';
+import { FeatureGateContextProvider, FeatureListType } from '../../features/featureGate';
 
 import './NewClusterPage.css';
+import OpenShiftVersionSelect from '../clusterConfiguration/OpenShiftVersionSelect';
```
mareklibra, line 36: nit: order

mareklibra (dismissed):

src/components/clusters/NewClusterPage.tsx
jtomasek: done

mareklibra (approved):

PR #462, opened 2021-02-16
[WIP] MGMT-3651 Remove host names from event messages

jkilzi:
Before this PR The UI attempts to resolve the hostname in every event message if it is available, otherwise, it shows the host's UUID. The idea is to perform this logic on the server.

- DO NOT MERGE YET! (waiting for [this task](https://issues.redhat.com/browse/MGMT-4131) to be finished on BE side)

### DIFF
--- a/src/components/hosts/utils.ts
+++ b/src/components/hosts/utils.ts
@@ -137,6 +137,12 @@ export const getEnabledHosts = (hosts: Host[] = []) =>
 export const getHostname = (host: Host, inventory: Inventory) =>
   host.requestedHostname || inventory.hostname;
 
+export const findHostname = (hostId: string, hosts: Host[] = []): string => {
+  const host = hosts.find(({ id }) => id === hostId);
+
+  return host?.requestedHostname || hostId;
+};
+
 export const getHardwareTypeText = (inventory: Inventory) => {
   let hardwareTypeText = DASH;
   const { systemVendor } = inventory;
--- a/src/components/ui/EventsList.tsx
+++ b/src/components/ui/EventsList.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { EventList, Event } from '../../api/types';
+import { EventList, Event, HostList } from '../../api/types';
 import { Label } from '@patternfly/react-core';
 import { TableVariant, Table, TableBody, breakWord } from '@patternfly/react-table';
 import {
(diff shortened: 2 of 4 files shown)


src/components/hosts/utils.ts
```diff
@@ -120,6 +120,15 @@ export const hasKnownHost = (cluster: Cluster) =>
 export const getHostname = (host: Host, inventory: Inventory) =>
   host.requestedHostname || inventory.hostname;
 
+export const findHostname = (hostId: string, hosts: Host[] = []): string | null => {
+  const host = hosts.find(({ id }) => id === hostId);
+  if (!host) {
+    return null;
```
mareklibra, line 126: since `host.id` is still valid corner-case result of this function (see bellow), why not use it even in that case?

src/components/ui/EventsList.tsx
```diff
@@ -63,28 +68,32 @@ const EventsList: React.FC<EventsListProps> = ({ events }) => {
...
-            )}
-            {event.message}
-          </>
-        ),
-      },
-    ],
-    props: { event },
-  }));
+  const rows = sortedEvents.map((event) => {
+    const hostname = event.hostId ? findHostname(event.hostId, hosts) : null;
```
mareklibra, line 72: IMO, it would be more comfortable if the `findHostname` handles false-like `event.hostId` as well

src/components/ui/EventsList.tsx
```diff
@@ -63,28 +68,32 @@ const EventsList: React.FC<EventsListProps> = ({ events }) => {
...
+    return {
+      cells: [
+        {
+          title: <strong>{getHumanizedDateTime(event.eventTime)}</strong>,
+        },
+        {
+          title: (
+            <>
+              {event.severity !== 'info' && (
+                <>
```
mareklibra, line 83: no need for fragment here

src/components/hosts/utils.ts
jkilzi: As agreed I refactored this function so it always returns a value. 
See here 25a9c1865a9448d45fdaed6178ad57c5876c5bc9

src/components/ui/EventsList.tsx
jkilzi: Following our discussion we agreed that in that case whoever wants to use `findHostname` will need to peek into the function definition in order to see when this function returns `null`.
See here 25a9c1865a9448d45fdaed6178ad57c5876c5bc9 for the fix.

jkilzi:
This PR is using Redux in order to add the hostnames to the events. We prefer not to use Redux.
It also has some cross dependencies that had been affected by the folder structure refactor.
Since work on this issue has not been prioritized in the BE yet, we should rethink how to achieve that without using Redux and organize the code following the new folder structure convention.

PR #467, opened 2021-02-18
BZ-1917396 Fixes disk and memory display units

jkilzi:
- Disk sizes are displayed in GB (b * 1000) and memory in GiB (b * 1024)

### DIFF
--- a/src/components/hosts/HostRowDetail.tsx
+++ b/src/components/hosts/HostRowDetail.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import filesize from 'filesize.js';
 import { TextContent, Text, TextVariants, Grid, GridItem } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody, TableVariant } from '@patternfly/react-table';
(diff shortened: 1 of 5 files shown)


src/components/hosts/HostRowDetail.tsx
```diff
@@ -86,7 +86,7 @@ const DisksTable: React.FC<DisksTableProps> = ({ cluster, host, disks, installat
         },
         { title: <DiskLimitations disk={disk} /> },
         disk.driveType,
-        filesize(disk.sizeBytes || 0, 2, 'iec'),
+        filesize(disk.sizeBytes || 0, 2, 'si').replace(/b$/, 'B'),
```
mareklibra, line 89: nit: can we have a new function in utils for that? To be sure we use that conversion consistently even in the future.

jkilzi: sure I'll do that

jkilzi: solved at c82add81f8f5b6f30515be9fba07eb2c8c31fda9

jtomasek (approved):

PR #470, opened 2021-02-22
MGMT-4200 Fix misplaced wizard toolbar alerts

jtomasek:


mareklibra:
We worked on the same issue in parallel :-(
The solution here works but what about moving the section closer to alerts? https://github.com/openshift-assisted/assisted-ui-lib/pull/472

Let's choose one those solutions ...

The #472 handles some other issues as well, they would need to be addressed here.

mareklibra:
Let's merge this and finish the resp within #472

mareklibra (approved):

PR #471, opened 2021-02-22
MGMT-4203 Handle empty validationsInfo right after cluster creation

mareklibra:


src/components/clusterWizard/wizardTransition.ts
```diff
@@ -167,14 +167,17 @@ export const getWizardStepHostStatus = (
 ): Host['status'] => {
   const { status } = host;
   if (['insufficient', 'pending-for-input'].includes(host.status)) {
-    const validationsInfo = stringToJSON<HostValidationsInfo>(host.validationsInfo) || {};
+    const validationsInfo = stringToJSON<HostValidationsInfo>(host.validationsInfo);
+    if (!validationsInfo) {
```
jtomasek, line 171: This is not necessary as `checkHostValidationGroups` and `checkHostValidations` will return false if `validationsInfo` is empty object. See wizardTransitions.test.ts

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -167,14 +167,17 @@ export const getWizardStepHostStatus = (
 ): Host['status'] => {
   const { status } = host;
   if (['insufficient', 'pending-for-input'].includes(host.status)) {
```
jtomasek, line 169: To be consistent with the other change:
```suggestion
  if (['insufficient', 'pending-for-input'].includes(status)) {
```

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -210,7 +213,12 @@ export const getWizardStepClusterStatus = (
 ): Cluster['status'] => {
   const { status } = cluster;
   if (['insufficient', 'pending-for-input'].includes(cluster.status)) {
```
jtomasek, line 215: To be consistent with the other change:
```suggestion
  if (['insufficient', 'pending-for-input'].includes(status)) {
```

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -210,7 +213,12 @@ export const getWizardStepClusterStatus = (
...
-    const validationsInfo = stringToJSON<ClusterValidationsInfo>(cluster.validationsInfo) || {};
+    const validationsInfo = stringToJSON<ClusterValidationsInfo>(cluster.validationsInfo);
+    if (!validationsInfo) {
```
jtomasek, line 217: Same as with host functions. This is not needed as check functions will return false if validationsInfo is empty.

src/components/clusterWizard/ClusterWizardStep.tsx
```diff
@@ -64,7 +64,7 @@ const ClusterWizardStep: React.FC<ClusterWizardStepProps> = ({ cluster, footer,
         key="cluster-details"
         content={wizardStepNames['cluster-details']}
         isCurrent={currentStepId === 'cluster-details'}
-        isValid={() => !cluster || canNextClusterDetails({ cluster })}
+        isValid={() => !cluster?.validationsInfo || canNextClusterDetails({ cluster })}
```
jtomasek, line 67: Not needed, see comments below.

jtomasek (changes requested):

jtomasek:
The root of the problem is that we're not calling `canNext...` function in `handleClusterCreate`. A result of that function call should decide what we set as `locationState` to make a correct redirect.

src/components/clusterWizard/ClusterWizardStep.tsx
mareklibra: let's keep it for the first step, but remove for others since it is dead code there.

src/components/clusterWizard/wizardTransition.ts
mareklibra: done within #472

src/components/clusterWizard/ClusterWizardStep.tsx
mareklibra: Handling `canNextClusterDetails` within the create step would cause additional timing issues for that flow.

I hope I am not mistaken, but there should not be the case where the cluster is created but relevant validations errors are returned. I believe, the only valid flow during cluster create is either success or an HTTP error which is already handled.

jtomasek (approved):

PR #472, opened 2021-02-22
MGMT-4200 Fix position of wizard validation section

mareklibra:
Depends on #470

### DIFF
--- a/src/components/clusterConfiguration/ClusterValidationSection.tsx
+++ b/src/components/clusterConfiguration/ClusterValidationSection.tsx
@@ -63,6 +63,8 @@ const ClusterValidationSection: React.FC<ClusterValidationSectionProps> = ({
     }
   });
 
+  const isContentToDisplay =
+    (!errorFields.length && dirty) || !!errorFields.length || !!failedValidations.length;
   return (
     <Split>
       <SplitItem isFilled>
@@ -102,6 +104,11 @@ const ClusterValidationSection: React.FC<ClusterValidationSectionProps> = ({
                   </List>
                 </FlexItem>
               </Flex>
+            </Alert>
+          )}
+          {!isContentToDisplay && (
+            <Alert variant={AlertVariant.info} title="Host validations are failing" isInline>
+              All relevant cluster-level validations are passing, inspect individual hosts.
             </Alert>
           )}
         </AlertGroup>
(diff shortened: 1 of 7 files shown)


src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -252,11 +254,21 @@ const ClusterDetailsForm: React.FC<ClusterDetailsFormProps> = (props) => {
                 <Alerts />
               </StackItem>
             )}
+            {isValidationSectionOpen && cluster && (
```
jtomasek, line 257: This is repeating on every step. Let's keep it in `ClusterWizardToolbar`. We could move whole `<Stack>` into a `ClusterWizardToolbar` component. Including Alerts.

jtomasek (changes requested):

src/components/clusterWizard/wizardTransition.ts
```diff
@@ -166,7 +166,7 @@ export const getWizardStepHostStatus = (
   wizardStepId: ClusterWizardStepsType,
 ): Host['status'] => {
   const { status } = host;
-  if (['insufficient', 'pending-for-input'].includes(host.status)) {
+  if (['insufficient', 'pending-for-input', 'disabled', 'discovering'].includes(host.status)) {
```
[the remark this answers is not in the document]
mareklibra: It depends on the actual step. I.e. in case of the first `Cluster details`, we should not block on those states.

src/components/clusterWizard/ClusterDetails.tsx
mareklibra: good point, moved

src/components/clusterWizard/wizardTransition.ts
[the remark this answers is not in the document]
mareklibra: To fix, adding a list of allowed statuses into `WizardStepValidationMap`

src/components/clusterConfiguration/ClusterValidationSection.tsx
```diff
@@ -104,6 +106,11 @@ const ClusterValidationSection: React.FC<ClusterValidationSectionProps> = ({
               </Flex>
             </Alert>
           )}
+          {!isContentToDisplay && (
```
mareklibra, line 109: Improves UX in corner cases or bugs, otherwise that should be covered by cluster validations.
Anyway, this is not happening recently, i.e. due to `sufficient-masters-count` workaround.

jtomasek (approved):

PR #478, opened 2021-02-24
MGMT-4202: Introduce soft validations

mareklibra:


mareklibra:
---

src/components/hosts/HardwareStatus.tsx
```diff
@@ -19,8 +20,18 @@ const HardwareStatus: React.FC<HardwareStatusProps> = (props) => {
...
     'baremetal-discovery',
   );
+  const sublabel = getFailingClusterWizardSoftValidationIds(validationsInfo, 'baremetal-discovery')
+    .length
+    ? 'Some validations failed'
+    : undefined;
+
   return (
-    <HostStatus {...props} statusOverride={hardwareStatus} validationsInfo={validationsInfo} />
+    <HostStatus
```
jtomasek, line 29: I think it is not necessary to calculate the sublabel here. Since we're passing the `validationsInfo` which contains only validations relevant for the given step, to determine the sublabel a `HostStatus` component can calculate the sublabel by flattening the validationsInfo values and checking if only failing validations are the 'soft' ones (very similar to `checkHostValidations` function.

jtomasek (changes requested):
Is it necessary to define soft validations per step? Eventually the backend is going to provide a flag for each validation to identify if it is 'soft' or not. I think that in the meantime we could just have 2 arrays of soft validations, one for host validations and one for cluster validations.

For review step we just check if a validation is in the list and adjust the icon.

src/components/hosts/HardwareStatus.tsx
mareklibra: The answer depends on complexity of the calculation (discussion bellow). If the definition of soft validations is simplified, then yes, we can calculate that sublabel here along the two ramaining cases. But if it is dependent on the context (I mean wizard step), then I would keep `HostStatus` generic.

mareklibra:
> Is it necessary to define soft validations per step? Eventually the backend is going to provide a flag for each validation to identify if it is 'soft' or not. 

Good question. Out of consistency. Once backend implements that (when is it due?), we will need to change/delete that code anyway. I remember discussion about transition states when we tried to define rules declaratively and at one place only, so I built on top of that thinking. I agree that we are effectively talking just about one validation now, but we do not know about timing for the backend changes and if something else will need to be handled the same way soon.

jtomasek (approved):

nirfarkas:
I think that even hardware requirements will be converted to "Soft" for the "Sandbox" profile at some point because we don't want to prevent installations for this profile and just to warn about performance.

PR #497, opened 2021-03-03
Min height 200 pull secret text area

omertuc:
When the window height is too small the pull secret text area gets squished down to height 0 and it's impossible/hard to interact with it.

Not sure if this is the correct way to fix it (and also not sure how to test it, it worked locally in my browser dev window). 

Feel free to modify the PR / fix it / close it / recreate it however you see fit


omertuc:
Like this? (see new diff)

jtomasek (approved):

PR #499, opened 2021-03-03
MGMT-3891 Default config UI sync

jkilzi:
Retrieves the default configuration dynamically from `/cluster/default-config` endpoint

### DIFF
--- a/src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
+++ b/src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
@@ -1,0 +1,47 @@
+import _ from 'lodash';
+import React, { useContext, useEffect, useState } from 'react';
+import { handleApiError, getErrorMessage } from '../../api';
+import { getClustersDefaultConfiguration } from '../../api/clusters';
+import { ClusterDefaultConfig } from '../../api/types';
+import { AlertsContext } from '../AlertsContextProvider';
+
+export const ClusterDefaultConfigurationContext = React.createContext<ClusterDefaultConfig>({});
+
+export const ClusterDefaultConfigurationProvider: React.FC = ({ children }) => {
+  const [defaultConfiguration, setDefaultConfiguration] = useState<ClusterDefaultConfig>({});
+  const { addAlert } = React.useContext(AlertsContext);
+
+  useEffect(() => {
+    let mounted = true;
+    const fetchAndSetDefaultConfiguration = async () => {
+      try {
+        const { data } = await getClustersDefaultConfiguration();
+        if (mounted) {
+          setDefaultConfiguration(data);
+        }
+      } catch (e) {
+        handleApiError(e, () =>
+          addAlert({
+            title: 'Failed to retrieve the default configuration',
+            message: getErrorMessage(e),
+          }),
+        );
+      }
+    };
+    fetchAndSetDefaultConfiguration();
+
+    return () => {
+      mounted = false;
+    };
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  return (
+    <ClusterDefaultConfigurationContext.Provider value={defaultConfiguration}>
+      {children}
+    </ClusterDefaultConfigurationContext.Provider>
+  );
+};
+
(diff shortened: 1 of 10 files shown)


src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  return (
+    <ClusterDefaultConfigurationContext.Provider value={defaultConfiguration}>
+      {children}
+    </ClusterDefaultConfigurationContext.Provider>
+  );
+};
+
+export const useDefaultConfiguration = (keys: Array<keyof ClusterDefaultConfig>) => {
```
jtomasek, line 45: It might be nice to make the `keys` parameter optional so it is possible to also select the values like this:
```
const { clusterNetworkCidr, serviceNetworkCidr } = useDefaultConfiguration();

File: src/components/clusterConfiguration/utils.ts [hunk 2]
```diff
@@ -61,40 +57,19 @@ export const getSubnetFromMachineNetworkCidr = (machineNetworkCidr?: string) =>
...
-  const defaultNetworkSettings = getDefaultNetworkSettings(cluster.clusterNetworkCidr as string);
-  return (
-    cluster.clusterNetworkCidr !== defaultNetworkSettings.clusterNetworkCidr ||
-    cluster.clusterNetworkHostPrefix !== defaultNetworkSettings.clusterNetworkHostPrefix ||
-    cluster.serviceNetworkCidr !== defaultNetworkSettings.serviceNetworkCidr
-  );
-};
+export const isAdvConf = (
+  cluster: Cluster,
+  defaultNetworkSettings: Partial<ClusterNetworkDefaultSettings>,
```
jtomasek, line 62: Since the `ClusterNetworkDefaultSettings` is always used as Partial, maybe we can change the type to have the keys optional and stop using Partial?

jtomasek (changes requested):

src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
jkilzi: @jtomasek this was my first intention but then I found myself doing stuff like this multiple times.

```typescript
const { clusterNetworkCidr, serviceNetworkCidr } = useDefaultConfiguration();
const defaultNetworkSettings = { clusterNetworkCidr, serviceNetworkCidr };
```
And as long as you want to pick a larger subset of these properties from the default config the code gets more and more verbose (and bundle-size-wise larger). So I came up with this API (which also has intellisense) in order to facilitate the destructuring and regrouping of those properties.

Please  let me know if you still want me to do the change

jtomasek: I was thinking it would be nice to have both options, so in case when we need only one value it is possible to do
```
const { myValue } = useDefaultConfiguration();
```
instead of more verbose
```
const { myValue } = useDefaultConfiguration(['myValue;]);
```

src/components/clusterConfiguration/utils.ts
jkilzi: @jtomasek if `ClusterNetworkDefaultSettings` would contain 1000 properties I would end up making 1000 changes...
Using `Partial` looks more convenient IMHO.

src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
jkilzi: But then you might leave the door open for using destructuring and then regrouping... If this patterns repeats itself it will bloat the final bundle...

src/components/clusterConfiguration/utils.ts
jkilzi: solved at 40e8458acf363613d6c773a5db5901ea5b129192

src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
jtomasek: ok, lets go with required keys parameter

jtomasek (approved):

PR #503, opened 2021-03-07
BZ-1931572 Setting role for a SNO node shows an error

jkilzi:
- Disables host role dropdown in SNO setup

### DIFF
--- a/src/components/hosts/HostsTable.tsx
+++ b/src/components/hosts/HostsTable.tsx
@@ -58,6 +58,7 @@ import {
   canReset,
   getHostname,
   canInstallHost,
+  canEditRole,
 } from './utils';
 import EditHostModal from './EditHostModal';
 import Hostname from './Hostname';
@@ -100,6 +101,7 @@ const hostToHostTableRow = (openRows: OpenRows, cluster: Cluster) => (host: Host
     (v) => v.id === 'has-cpu-cores-for-role',
   );
   const computedHostname = getHostname(host, inventory);
+  const hostRole = getHostRole(host);
 
   return [
     {
@@ -115,8 +117,10 @@ const hostToHostTableRow = (openRows: OpenRows, cluster: Cluster) => (host: Host
           sortableValue: computedHostname || '',
         },
         {
-          title: <RoleCell host={host} clusterStatus={cluster.status} />,
-          sortableValue: getHostRole(host),
+          title: (
+            <RoleCell host={host} readonly={!canEditRole(cluster, host.status)} role={hostRole} />
+          ),
+          sortableValue: hostRole,
         },
         {
           title: <HostStatus host={host} cluster={cluster} validationsInfo={validationsInfo} />,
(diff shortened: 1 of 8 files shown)


jtomasek (changes requested):
IMO this should be done on the `RoleCell` component level as part of `canEditRole` helper

src/components/hosts/HostsTable.tsx
```diff
@@ -115,8 +118,14 @@ const hostToHostTableRow = (openRows: OpenRows, cluster: Cluster) => (host: Host
           sortableValue: computedHostname || '',
         },
         {
-          title: <RoleCell host={host} clusterStatus={cluster.status} />,
-          sortableValue: getHostRole(host),
+          title: (
+            <RoleCell
+              host={host}
+              readonly={!canEditRole(cluster.status, host.status) || isSingleNodeCluster(cluster)}
```
jtomasek, line 124: Why taking the responsibility of calculating the hostRole and knowing if a role can be edited from the RoleCell?
Also the purpose of `canEditRole` function is to tall us if a role can be edited or not, why not including `isSingleNodeCluster` in it?

jkilzi: I've made this change after carefully inspecting what pieces of information each function/component needed: `RollCell`, `canEditRole` and `isSingleNodeCluster`.

### Rationale behind each decision:
1. Extract the `getHostRole` function from `RowCell`:
  Calling `getHostRole(host)` returned the same value inside the `RowCell` and outside of it (in its sole consumer,  `HostTable`). After the refactor I made `RowCell` doesn't depend on `getHostRole` anymore and also only one function call to `getHostRole` is being performed, so there is a small performance gain.
2. Purpose of `canEditRole` function:
  I totally agree with you what its purpose should be. 
  Being that said, lets discuss please where should we call this function.
  Imagine that I combine the result of `isSingleNodeCluster(cluster)` with its current logic as you mention. Then the signature of `canEditRole` would need to change from `canEditRole(clusterStatus, hostStatus)` to something like `canEditRole(cluster.status, host.status, cluster)` or perhaps just `canEditRole(cluster, host.status)`. In any case whoever is going to call `canEditRole` with this signature will need to pass the `cluster` object along the way. WDYT, who should be more concerned about the `cluster` object, the `HostTable` or the `RowCell`?

jkilzi: Also it turns out that `canEditHost` and `canEditDisks` are just aliases for `canEditRole`. If we have a SNO cluster should the disks not be editable? what about the host inside `HostRowDetail`?

jtomasek: 1. I agree
2. IMHO the purpose of `RoleCell` component is to know if it should be editable or not and to this we should align the props it requires, so changing the props to `cluster` and `host` is fine.
  `canEditRole`'s signature can be changed as well to match it's requirements. `canEditHost` and `canEditDisks` should be adjusted to match existing `canEditRole`

jtomasek (changes requested):

jtomasek (dismissed):

jtomasek (approved):

PR #509, opened 2021-03-10
BZ-1936360 Advanced networking checkbox is enabled when creating new cluster

jkilzi:
If the `clusters/default-config` API endpoint is not available the provider will render an "error" UI and a "loading" UI while it is fetching data.
The end goal is to prevent the form to be initially filled with empty data and then updated with default values coming from the API; that is what is enabling the checkbox.

### DIFF
--- a/src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
+++ b/src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
@@ -1,31 +1,52 @@
 import _ from 'lodash';
-import React, { useContext, useEffect, useState } from 'react';
-import { handleApiError, getErrorMessage } from '../../api';
+import React, { PropsWithChildren, useContext, useEffect } from 'react';
 import { getClustersDefaultConfiguration } from '../../api/clusters';
 import { ClusterDefaultConfig } from '../../api/types';
-import { AlertsContext } from '../AlertsContextProvider';
+
+type State<T> = { status: 'idle' | 'loading' | 'succeeded' | 'failed'; data: T };
+type Actions<T> = { type: 'request' } | { type: 'failure' | 'success'; data: T };
+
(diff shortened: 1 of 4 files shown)


src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
```diff
@@ -1,31 +1,52 @@
...
+      break;
+    case 'failure':
+      newState = { status: 'failed', data: action.data };
+      break;
+    default:
+      newState = state;
+      break;
+  }
+
+  return newState;
```
jtomasek, line 29: This could be simplified to:

```suggestion
  switch (action.type) {
    case 'request':
      return { status: 'loading', data: {} };
    case 'success':
      return { status: 'succeeded', data: action.data };
    case 'failure':
      return { status: 'failed', data: action.data };
    default:
      return state;
  }
```

jtomasek (commented):
As we discussed, this is a good enough solution for resolving the bug.

Following steps:
- Data fetching and UI state should be provided via a separate react hook which will let a consuming component (NewClusterPage, ClusterPage) decide when to fetch the data and how to display error and loading. 
```
const { fetchDefaultConfiguration, data: defaultConfiguration, uiState: defaultConfigUIState } = useFetchDefaultConfiguration()
```
-> this hook can be generalized so all data fetching is done by a standard way (based on existing useApi hook)

- ClusterDefaultConfigurationProvider should not render anything and just serve as a simple way to let components in the tree use the default configuration. It takes the data as prop.

src/components/clusterConfiguration/ClusterDefaultConfigurationContext.tsx
jkilzi: Agree. IMHO I don't like many `return` statements. I see this as an optimization which is performed anyway by webpack at build time.

jtomasek (dismissed):

jtomasek (approved):

PR #512, opened 2021-03-15
BZ 1938969: Disable Back button on BareMetalDiscovery

mareklibra:
when validationsInfo is not available.
That might happen right after cluster's creation.


mareklibra:
clusterValidationMissing

mareklibra:
@nirfarkas Wdyt about disabling the back button in the case when the `validationsInfo` is not yet available?

With recent implementation, the user can not go Back from the 2nd step for approx 20 seconds during the `Cluster creation` flow (not the case of `Edit`).
Another approach can be in allowing the Back button immediately but block the user to do Next afterwards.

jtomasek (approved):

PR #515, opened 2021-03-16
BZ 1939134 The Next button stays disabled on the Networking step

mareklibra:
The form stayed dirty after `handleSubmit`.


mareklibra:
With this change, the flow works.

Anyway, I see bad UX in waiting for the VIP validations to pass - the user can see all the data (including auto-allocated VIPs)  but still can not move on until VIP validations are passing. I think a spinner denoting "running validations" will be big improvement here. Such waiting can take up to 2 minutes and is completely fine from the backend design perspective.

The UX can be resolved by adding `running` state of validations on the backend.

Anyway, it will take some effort to get there, so what about adding text to the failing validations list telling sort of "Validations are executed asynchronously on the backend. Please wait, they might successfully finish in a few minutes". @nirfarkas , wdyt?

jtomasek:
Adding `enableReinitialize` means that any time we calculate new `initialValues` (e.g. when a new cluster arrives from polling) we reset the form. This means that we override any unsaved changes user did and it will introduce the resetting problems which autosave originally had.

If `vipDhcpAllocation` checkbox is enabled we discard the apiVip and ingressVip values on submit but they are still present in the form. Once DHCP allocation is finished the cluster vip values become different from the formik values and causes the form to be dirty. IMO a simple fix is to make the StaticField component sync the cluster value to the formik to ensure these are aligned: https://github.com/openshift-assisted/assisted-ui-lib/pull/518

PR #516, opened 2021-03-17
BZ 1934198: Change the reboot config text to info-level

mareklibra:
Prior this patch, the message was marked as warning.


mareklibra:
rebootInfo

PR #528, opened 2021-03-30
MGMT-3230 Cluster Configuration Wizard

jtomasek:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "openshift-assisted-ui-lib",
-  "version": "1.5.12",
+  "version": "0.0.1-wizard",
```
mareklibra, line 3: let's keep `1.5.12`

PR #532, opened 2021-04-06
Fix the operator popover styling

jtomasek:


src/components/clusterWizard/ClusterWizardToolbar.tsx
```diff
@@ -74,7 +74,7 @@ const ValidationSectionToggle: React.FC<ValidationSectionToggleProps> = ({
     return (
       <ToolbarButton variant={ButtonVariant.link} onClick={toggleValidationSection} isInline>
         <WarningTriangleIcon color={warningColor.value} />
-        &nbsp;<small>The configuration form has unresolved validation errors.</small>
+        &nbsp;<small>The form has unresolved validation errors.</small>
```
jtomasek, line 77: A small change to make the message a bit more universal.

nirfarkas: Where and when do we show this message?

jtomasek: Next to the wizard navigation buttons in case when the step's form has client side validations errors

nirfarkas: Hi Jiri, Can you please share a screen capture?

jtomasek: Screenshot 2021-04-08 at 16 24 12

mareklibra (approved):

PR #533, opened 2021-04-06
BZ 1938578 Disable LSO with OCS

jtomasek:


src/components/clusterWizard/BaremetalDiscovery.tsx
```diff
@@ -31,6 +31,8 @@ const BaremetalDiscovery: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
       enabledOlmOperatorsByName.ocs = { name: 'ocs' };
     } else {
       delete enabledOlmOperatorsByName.ocs;
+      // TODO(jtomasek): remove this once enabling OCS is moved into a separate storage step and LSO option is exposed to the user
+      delete enabledOlmOperatorsByName.lso;
```
mareklibra, line 35: When CNV operator is selected( #536 ), the LSO is added automatically by backend.
The user should be aware of it. Maybe simple text explanation in the CNV popover is enough, @nirfarkas ?

mareklibra: Let's not block that patch and resolve it within #536

mareklibra (approved):

PR #395, opened 2021-01-03
Add cluster details and progress cards to OCM overview tab 

irosenzw:

PR #535, opened 2021-04-06
Add AI cluster card to uhc-portal overview tab

mareklibra:
Rebased #395 and further updated.

For follow-ups:
- optimization of OCM fetching (see new comments in OCM's `clusterActions.js`)
- after `postCluster`, navigate to OCM via subscription ID instead of specific AI page (see comment in `clusterDetails.tsx`)
- improve wizard's placing in the card (styling)
- fix placing of redux Provider


nirfarkas:
Any chance for screen capture? (:

mareklibra:
@nirfarkas please check last comments in #395, I have put them there. Anyway, testing of this is in progress ...
I.e. bits around operators has been changed.

mareklibra:
There has been a bunch of conflicting changes on both OCM and AI side. But it seems it still works after rebase:



---

src/components/clusterDetail/ClusterInstallationProgressCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }, [cluster.id]);
+
+  React.useEffect(() => {
+    if (cluster.status === 'installed') {
+      fetchCredentials();
+    }
+  }, [cluster.status, fetchCredentials]);
+
+  return (
+    <Card id="ai-cluster-details-card" isExpanded={isCardExpanded}>
```
jkilzi, line 50: @mareklibra  if this `id` is there only for testings purposes then it would be better to make it a `data-testid`

src/components/clusterDetail/ClusterInstallationProgressCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+              setResetClusterModalOpen={setResetClusterModalOpen}
+            />
+          )}
+          {cluster.status === 'installed' && (
+            <ClusterCredentials
+              cluster={cluster}
+              credentials={credentials}
+              error={!!credentialsError}
+              retry={fetchCredentials}
+              idPrefix={getID('cluster-creds')}
```
jkilzi, line 109: @mareklibra same thing here for the idPrefix prop. This is going to be translated into some nested element id attr.
If this is used for e2e tests then please stick to `data-testid`

src/components/clusterDetail/ClusterInstallationProgressCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+const AIClusterDetailsCard: React.FC<AIClusterDetailsCardProps> = ({ clusterStringObj }) => {
+  const cluster: Cluster = stringToJSON<Cluster>(clusterStringObj) as Cluster;
+
+  const [cancelInstallationModalOpen, setCancelInstallationModalOpen] = React.useState(false);
+  const [resetClusterModalOpen, setResetClusterModalOpen] = React.useState(false);
+  const [credentials, setCredentials] = React.useState<Credentials>();
+  const [credentialsError, setCredentialsError] = React.useState();
+  const [isCardExpanded, setIsCardExpanded] = React.useState(cluster.status !== 'installed');
+
+  const fetchCredentials = React.useCallback(() => {
```
jtomasek, line 30: It would be good to better consolidate the `ClusterInstallationProgressCard` and `ClusterDetail` components as there is duplicated code and the components have diverged since addition of monitored operators.

jtomasek (changes requested):

mareklibra:
- rebased
- fixed data-testid
- fixed layout
- fixed extensive data-fetching
- added sharing of common code via a hook

src/components/clusterDetail/ClusterCredentials.tsx
```diff
@@ -68,11 +68,7 @@ const ClusterCredentials: React.FC<ClusterCredentialsProps> = ({
     );
   }
 
-  return (
-    <GridItem span={12} lg={10} xl={6}>
-      {credentialsBody}
-    </GridItem>
-  );
+  return <GridItem span={12}>{credentialsBody}</GridItem>;
```
jtomasek, line 71: This makes the password clipboard copy 100% width which is too much
Screenshot 2021-04-14 at 12 59 45

src/components/clusterDetail/ClusterDetailStatusVarieties.tsx
```diff
@@ -0,0 +1,122 @@
...
+    failedOlmOperators,
+    consoleOperator,
+    fetchCredentials,
+  };
+};
+
+const ClusterDetailStatusVarieties: React.FC<{
+  cluster: Cluster;
+  setResetClusterModalOpen: (isOpen: boolean) => void;
+  clusterVarieties: ClusterStatusVarieties;
```
jtomasek, line 73: How about calling the hook directly in this component rather than doing it in parent and passing it in as prop? It is not used in the parent in any way.

src/components/clusterDetail/ClusterInstallationProgressCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <CardBody>
+        <ClusterProgress cluster={cluster} minimizedView={!isCardExpanded} />
+        <ClusterDetailsButtonGroup cluster={cluster} />
+      </CardBody>
+      <CardExpandableContent>
+        <CardBody>
+          <Grid hasGutter>
+            <ClusterDetailStatusVarieties
+              cluster={cluster}
+              setResetClusterModalOpen={setResetClusterModalOpen}
```
jtomasek, line 80: To avoid having to pass setResetClusterModalOpen as prop in several layers we could create `ClusterDialogsContext` for cluster dialogs in a same way as we have `HostDialogsContext` for host dialogs

src/components/clusterDetail/ClusterDetailsButtonGroup.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { downloadClusterInstallationLogs } from './utils';
+import { AlertsContext } from '../AlertsContextProvider';
+
+type ClusterDetailsButtonGroupProps = {
+  cluster: Cluster;
+};
+
+const getID = (suffix: string) => `cluster-detail-${suffix}`;
+
+const ClusterDetailsButtonGroup: React.FC<ClusterDetailsButtonGroupProps> = ({ cluster }) => {
```
jtomasek, line 16: Isn't this group missing Abort installation button?

I'd really like to use this group in Standalone's `ClusterDetail` page and completely get rid of ClusterToolbar.

src/components/clusterDetail/ClusterDetailStatusVarieties.tsx
mareklibra: Actually it is - in the ClusterDetail flow, the credentials are passed to ClusterCredentials. Querying it in the parent component is one of the ways to ensure just a single query per component hierarchy.
We can wrap iit n a context or do another refactoring of the hierarchy, but the `ClusterDetailStatusVarieties` is not expected to be reused any farther in the future so this is the most straightforward way, imo.

src/components/clusterDetail/ClusterInstallationProgressCard.tsx
mareklibra: It's just one layer down, but I agree, the amount of new code will simplify reuse later.

mareklibra: refactored

src/components/clusterDetail/ClusterDetailsButtonGroup.tsx
mareklibra: right, added

mareklibra:
New change: Introduced new `AssistedInstallerDetailCard` component rendering either wizard or progress based on state. Handles polling.

src/components/clusterDetail/ClusterDetailStatusVarieties.tsx
jtomasek: So you're expecting the `useClusterStatusVarieties` hook to be reused somewhere else other than to populate data for `ClusterDetailStatusVarieties` component?

The hook is preparing data specific for the `ClusterDetailsStatusVarieties` component. I think trying to make it reusable is wrong.

src/components/hosts/utils.ts
```diff
@@ -103,6 +103,16 @@ export const canDownloadHostLogs = (host: Host) =>
 export const canDownloadClusterLogs = (cluster: Cluster) =>
   !!(cluster.hosts || []).find((host) => canDownloadHostLogs(host));
 
+export const canAbortInstallation = (cluster: Cluster) => {
```
jtomasek, line 106: This is cluster specific, it should go into `components/clusters/utils.ts

src/components/clusters/ClusterPage.tsx
```diff
@@ -196,19 +156,13 @@ const ClusterPage: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
...
-            onClose={() => setCancelInstallationModalOpen(false)}
-            clusterId={cluster.id}
-          />
-          <ResetClusterModal
-            isOpen={resetClusterModalOpen}
-            onClose={() => setResetClusterModalOpen(false)}
-            cluster={cluster}
-          />
-        </ClusterDefaultConfigurationProvider>
+        <HostDialogsContextProvider>
```
jtomasek, line 159: This is very confusing because of the `HostDialogsContextProvider` name which now includes cluster specific dialogs. I also suspect that nesting these provides may cause more problems?

jtomasek (changes requested):

src/components/clusters/ClusterPage.tsx
mareklibra: Right, what about renaming it? I don't see a point in having it specific, the IDs can be unique. Will double-check nesting, would be unintended.

jtomasek: ok renaming should be fine I think.

jtomasek: Since we'll have `DialogsContextProvider` which we use to track all dialogs state, we can render it just at `ClusterPage`, `AI Detail card` (+ other OCM related components if we have such) and `Clusters`.

We can remove it from HostsTable

mareklibra: done

src/components/hosts/utils.ts
mareklibra: oops

jtomasek (approved):

PR #537, opened 2021-04-08
Update API

mareklibra:


src/api/swagger.yaml
```diff
@@ -3262,16 +3274,31 @@ definitions:
       cpu_cores:
         type: integer
         description: Required number of CPU cores
-      ram_gib:
+      ram_mib:
```
jtomasek, line 3277: This most likely requires update in the UI code

src/api/types.ts
```diff
@@ -1105,7 +1168,7 @@ export type HostValidationId =
   | 'container-images-available'
   | 'lso-requirements-satisfied'
   | 'ocs-requirements-satisfied'
-  | 'sufficient-installation-disk-speed'
+  | 'sufficient-or-unknown-installation-disk-speed'
```
jtomasek, line 1171: This key needs to be updated in `config/constants.ts`

jtomasek (changes requested):

src/api/swagger.yaml
mareklibra: This is new addition to the API, so far not used in the UI.

src/api/types.ts
mareklibra: fixed

jtomasek (approved):

mareklibra:
Let's merge this after release today

jtomasek (approved):

PR #540, opened 2021-04-11
MGMT-4051 Use disk ID as disk identifier in UI

jkilzi:
1. When calling BE with disks_selected_config use the disk ID as the disk ID and not Name
2. When identifying the disks in the list - use Id and not path (/dev/Name)


jkilzi:
BE has some issues with their current implementation. However the changes introduced in this PR will work as expected once the BE solves the problem on their side.

jkilzi:
The BE fix is in stage now

mareklibra (approved):

PR #541, opened 2021-04-12
MGMT-4369 Use new default attribute of OpenShift versions endpoint 

mareklibra:
Depends on:
- [x] #537 API update


src/config/constants.ts
```diff
@@ -22,6 +22,7 @@ export const PULL_SECRET_INFO_LINK = CLUSTER_MANAGER_SITE_LINK;
 export const DEFAULT_OPENSHIFT_VERSION: OpenshiftVersionOptionType = {
```
jtomasek, line 22: How about removing this and not provide any default for `getBugzillaLink`? It is outdated and not really used.

jtomasek (changes requested):

src/config/constants.ts
mareklibra: Good point. The `DEFAULT_OPENSHIFT_VERSION` seized to server its original purpose.

jtomasek (approved):

PR #543, opened 2021-04-12
BZ 1931635 Don't include disabled hosts in host progress counts

jtomasek:


src/components/hosts/utils.ts
```diff
@@ -129,6 +129,9 @@ export const downloadHostInstallationLogs = async (
 export const hasKnownHost = (cluster: Cluster) =>
   !!cluster.hosts?.find((host) => host.status === 'known');
 
+export const getEnabledHosts = (cluster: Cluster) =>
+  (cluster.hosts || []).filter((host) => host.status !== 'disabled');
```
mareklibra, line 133: nit: it would be good to replace other similar cases by such a utility function to simplify maintenance. We use such filter explicitly quite often.

mareklibra (approved):

mareklibra (approved):

PR #545, opened 2021-04-13
BZ 1946313: Review&Create failing valition hint

mareklibra:


src/components/clusterConfiguration/ReviewValidations.tsx
```diff
@@ -67,7 +67,10 @@ const FailingValidation: React.FC<{
 
   let fix;
   const step = findValidationFixStep({ id: validation.id, clusterGroup, hostGroup });
-  if (step) {
+  if (step === 'review') {
```
jtomasek, line 70: Maybe we should just remove the 'all-hosts-are-ready-to-install' from `reviewStepValidationsMap`? User is not able to fix anything in the review step, so it makes no sense to include that validation there.

Doing so would highlight that validation as failed and link to a step would not show up as expected.

src/components/clusterConfiguration/ReviewValidations.tsx
```diff
@@ -67,7 +67,10 @@ const FailingValidation: React.FC<{
...
+    // no sooner step, so the user can not do anything about it ...
+    fix = 'Please wait till all validations are finished.';
```
jtomasek, line 72: This is not exactly true as `all-hosts-are-ready-to-install` validation is failed and won't get updated.

jtomasek (changes requested):

src/components/clusterConfiguration/ReviewValidations.tsx
mareklibra: Asking backend to confirm, but so far that was my understanding, so the fix would be ok. Let's see their response.

src/components/clusterConfiguration/ReviewValidations.tsx
mareklibra: But we need to disable `Install` in that case, don't we?

src/components/clusterConfiguration/ReviewValidations.tsx
mareklibra: confirmed, just temporary flow till the validation is re-executed

jtomasek (approved):

PR #547, opened 2021-04-14
MGMT-2754 Enable cluster/user-managed networking

jkilzi:
This PR allows the user to select between user-managed or cluster-managed networking

Screencast from 20-04-21 18_06_29

### DIFF
--- a/src/components/clusterConfiguration/NetworkConfiguration.tsx
+++ b/src/components/clusterConfiguration/NetworkConfiguration.tsx
@@ -1,21 +1,30 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useFormikContext } from 'formik';
 import { Checkbox } from '@patternfly/react-core';
-import BasicNetworkFields from './BasicNetworkFields';
 import AdvancedNetworkFields from './AdvancedNetworkFields';
 import { HostSubnets, NetworkConfigurationValues } from '../../types/clusters';
-import { Cluster } from '../../api/types';
+import { Cluster } from '../../api';
 import { isSingleNodeCluster } from '../clusters/utils';
 import { isAdvConf } from './utils';
 import { useDefaultConfiguration } from './ClusterDefaultConfigurationContext';
+import {
+  AvailableSubnetsControl,
+  ManagedNetworkingControlGroup,
+  UserManagedNetworkingTextContent,
+  VirtualIPControlGroup,
+} from '../clusterWizard/networkingSteps';
+import { RenderIf } from '../ui/RenderIf';
+import { NO_SUBNET_SET } from '../../config';
 
 type NetworkConfigurationProps = {
   cluster: Cluster;
   hostSubnets: HostSubnets;
 };
 
-const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({ cluster, hostSubnets }) => {
-  const { setFieldValue } = useFormikContext<NetworkConfigurationValues>();
+const NetworkConfiguration = ({ cluster, hostSubnets }: NetworkConfigurationProps) => {
+  const { setFieldValue, values, touched, validateField } = useFormikContext<
+    NetworkConfigurationValues
+  >();
   const defaultNetworkSettings = useDefaultConfiguration([
     'clusterNetworkCidr',
     'serviceNetworkCidr',
@@ -34,11 +43,54 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({ cluster, ho
     }
   };
 
+  const isMultiNodeCluster = !isSingleNodeCluster(cluster);
+  const isUserManagedNetworking = values.networkingType === 'userManaged';
+
+  useEffect(() => {
+    if (isUserManagedNetworking) {
+      const shouldValidate = true;
+
+      // We need to reset these fields' values in order to align with the values the server sends
+      setFieldValue('vipDhcpAllocation', false);
+      setFieldValue('ingressVip', '', !shouldValidate);
+      setFieldValue('apiVip', '', !shouldValidate);
+      if (!touched.hostSubnet || isMultiNodeCluster) {
+        setFieldValue('hostSubnet', NO_SUBNET_SET, !shouldValidate);
+      }
+    } else {
+      if (!values.vipDhcpAllocation) {
+        validateField('ingressVip');
+        validateField('apiVip');
+      }
+    }
+  }, [
+    touched.hostSubnet,
+    isMultiNodeCluster,
+    isUserManagedNetworking,
+    setFieldValue,
+    values.vipDhcpAllocation,
+    validateField,
+  ]);
+
   return (
     <>
-      {!isSingleNodeCluster(cluster) && (
-        <BasicNetworkFields cluster={cluster} hostSubnets={hostSubnets} />
-      )}
+      <ManagedNetworkingControlGroup disabled={!isMultiNodeCluster} />
+      <RenderIf condition={isUserManagedNetworking}>
+        <UserManagedNetworkingTextContent shouldDisplayLoadBalancersBullet={isMultiNodeCluster} />
+      </RenderIf>
+
+      <RenderIf condition={!(isMultiNodeCluster && isUserManagedNetworking)}>
+        <AvailableSubnetsControl
+          hostSubnets={hostSubnets}
+          hosts={cluster.hosts || []}
+          isRequired={!isUserManagedNetworking}
+        />
+      </RenderIf>
+
+      <RenderIf condition={!isUserManagedNetworking}>
+        <VirtualIPControlGroup cluster={cluster} hostSubnets={hostSubnets} />
+      </RenderIf>
+
       <Checkbox
         id="useAdvancedNetworking"
         label="Use Advanced Networking"
@@ -46,7 +98,9 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({ cluster, ho
         isChecked={isAdvanced}
         onChange={toggleAdvConfiguration}
       />
-      {isAdvanced && <AdvancedNetworkFields />}
+      <RenderIf condition={isAdvanced}>
+        <AdvancedNetworkFields />
+      </RenderIf>
     </>
   );
 };
--- a/src/components/clusterConfiguration/utils.ts
+++ b/src/components/clusterConfiguration/utils.ts
@@ -82,5 +82,6 @@ export const getNetworkInitialValues = (
     sshPublicKey: cluster.sshPublicKey || '',
     hostSubnet: getSubnetFromMachineNetworkCidr(cluster.machineNetworkCidr),
     vipDhcpAllocation: cluster.vipDhcpAllocation,
+    networkingType: cluster.userManagedNetworking ? 'userManaged' : 'clusterManaged',
   };
 };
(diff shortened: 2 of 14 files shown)


src/components/clusterConfiguration/utils.ts
```diff
@@ -86,5 +86,6 @@ export const getNetworkInitialValues = (
     sshPublicKey: cluster.sshPublicKey || '',
     hostSubnet: getSubnetFromMachineNetworkCidr(cluster.machineNetworkCidr),
     vipDhcpAllocation: cluster.vipDhcpAllocation,
+    userManagedNetworking: cluster.userManagedNetworking,
```
jtomasek, line 89: Just to be safe
```suggestion
    userManagedNetworking: !!cluster.userManagedNetworking, 
```

src/components/clusterWizard/networkingSteps/ManagedNetworkingControlGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Split, SplitItem } from '@patternfly/react-core';
+import { RadioField } from '../../ui/formik/RadioField';
+
+interface ManagedNetworkingControlGroupProps {
+  isUserManagedNetworking: boolean;
+  disabled: boolean;
+}
+
+export const ManagedNetworkingControlGroup = ({
+  isUserManagedNetworking = false,
```
jtomasek, line 11: not needed

src/components/clusterWizard/networkingSteps/ManagedNetworkingControlGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const GROUP_NAME = 'userManagedNetworking';
+
+  return (
+    <Split hasGutter>
+      <SplitItem>
+        <RadioField
+          name={GROUP_NAME}
+          id={'clusterManagedNetworking'}
+          disabled={disabled}
+          checked={!isUserManagedNetworking}
```
jtomasek, line 23: This can be calculated on the RadioField

src/components/ui/formik/RadioField.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const [, , { setValue }] = useField({
+    name: props.name,
+    type: 'radio',
+  });
+
+  const handleChange = React.useCallback(
+    (checked, event) => {
+      const { value } = event.currentTarget;
+      if (checked) {
+        setValue(value === 'userManaged');
```
jtomasek, line 17: We can't make this generic as the value check is hardcoded

src/components/ui/formik/RadioField.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { PropsWithoutRef } from 'react';
+import { FieldAttributes, useField } from 'formik';
+import { Radio, RadioProps } from '@patternfly/react-core';
+
+type RadioFieldProps = PropsWithoutRef<FieldAttributes<RadioProps>>;
+
+export const RadioField = (props: RadioFieldProps) => {
+  const [, , { setValue }] = useField({
+    name: props.name,
+    type: 'radio',
```
jtomasek, line 10: I had to remove this to get the 'field.value' working as expected.

Alternatively, and if we wanted to make the RadioField generic, we would have to let formik use it as radio which requires to let it set value as string which requires us to create a custom form value and derive the userManagedNetworking boolean from it in handleSubmit.

src/components/ui/formik/RadioField.tsx
```diff
@@ -0,0 +1,24 @@
...
+    (checked, event) => {
+      const { value } = event.currentTarget;
+      if (checked) {
+        setValue(value === 'userManaged');
+      }
+    },
+    [setValue],
+  );
+
+  return <Radio {...props} onChange={handleChange} />;
```
jtomasek, line 23:
```suggestion
  return <Radio
      {...props}
      isChecked={
        (props.value === 'userManaged' && field.value) ||
        (props.value === 'clusterManaged' && !field.value)
      }
      onBlur={field.onBlur}
      onChange={handleChange}
    />;
```

jtomasek: Note that the trick to avoid 'controlled -> uncontrolled' warnings was to use `isChecked` instead of `checked`

src/components/ui/formik/RadioField.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { PropsWithoutRef } from 'react';
+import { FieldAttributes, useField } from 'formik';
+import { Radio, RadioProps } from '@patternfly/react-core';
+
+type RadioFieldProps = PropsWithoutRef<FieldAttributes<RadioProps>>;
+
+export const RadioField = (props: RadioFieldProps) => {
+  const [, , { setValue }] = useField({
```
jtomasek, line 8:
```suggestion
  const [field, , { setValue }] = useField({
```

src/components/clusterWizard/networkingSteps/ManagedNetworkingControlGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+}: ManagedNetworkingControlGroupProps) => {
+  const GROUP_NAME = 'userManagedNetworking';
+
+  return (
+    <Split hasGutter>
+      <SplitItem>
+        <RadioField
+          name={GROUP_NAME}
+          id={'clusterManagedNetworking'}
+          disabled={disabled}
```
jtomasek, line 22:
```suggestion
          isDisabled={disabled}
```

src/components/clusterWizard/networkingSteps/ManagedNetworkingControlGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+          value={'clusterManaged'}
+          label="Cluster-Managed Networking"
+        />
+      </SplitItem>
+      <SplitItem style={{ marginRight: '2.4rem' }} />
+      <SplitItem>
+        <RadioField
+          name={GROUP_NAME}
+          id={'userManagedNetworking'}
+          disabled={disabled}
```
jtomasek, line 33:
```suggestion
          isDisabled={disabled}
```

jtomasek (changes requested):
Comments inline. In addition we'll need to update a form values when changing to 'user managed networking' for example set `vipDhcpAllocation` to false...

jtomasek:
Added a commit which makes the RadioField properly reusable.

There is still an issue with VIP validations, I'll continue looking into this. In addition I am also getting this from the server:
Screenshot 2021-04-15 at 13 05 55

Which IIUC means that there is a change required on the server as well.

src/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -0,0 +1,60 @@
...
+                  label: hn.humanized,
+                  value: hn.humanized,
+                  id: `form-input-hostSubnet-field-option-${index}`,
+                })),
+            ]
+          : [{ label: 'No subnets are currently available', value: NO_SUBNET_SET }]
+      }
+      getHelperText={getHelperText}
+      onChange={() => {
+        if (!values.vipDhcpAllocation) {
```
jtomasek, line 51:
```suggestion
        if (!values.vipDhcpAllocation && values.networkingType === 'clusterManaged') {
```

src/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -34,19 +47,67 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({ cluster, ho
...
+    [setFieldTouched, setFieldValue],
+  );
+  const handleAvailableSubnetsControlChange = React.useCallback<
+    Required<AvailableSubnetsControlProps>['onChange']
+  >(() => {
+    if (!values.vipDhcpAllocation && !isUserManagedNetworking) {
+      validateField('ingressVip');
+      validateField('apiVip');
+    }
+  }, [isUserManagedNetworking, validateField, values.vipDhcpAllocation]);
```
jtomasek, line 77: These could be better handled using useEffect when related values change. I think it would make the code a bit simpler.

Example here: https://formik.org/docs/examples/dependent-fields

jkilzi: This is what I did following that example. The ref was necessary in order to remember the value in the single-node cluster scenario (what happened is that after moving to the final step and pressing back the hostSubnet is reset). In multi-node cluster the next button is disabled. At this point I think that if we can't spot what's wrong we should move forward with what we have (at least it works right now...) we can put a task aside to refactor this part later.

```typescript
const hostSubnetRef = useRef(values.hostSubnet || NO_SUBNET_SET);

  useEffect(() => {
    const isTouched = true;
    const shouldValidate = true;

    if (isUserManagedNetworking) {
      setFieldValue('vipDhcpAllocation', false);
      setFieldValue('hostSubnet', hostSubnetRef.current, !shouldValidate);
      setFieldTouched('hostSubnet', !isTouched, !shouldValidate);
      setFieldValue('ingressVip', '');
      setFieldTouched('ingressVip', !isTouched, !shouldValidate);
      setFieldValue('apiVip', '');
      setFieldTouched('apiVip', !isTouched, !shouldValidate);
    } else {
      if (!values.vipDhcpAllocation) {
        validateField('ingressVip');
        validateField('apiVip');
      }
    }
  }, [
    isUserManagedNetworking,
    setFieldTouched,
    setFieldValue,
    validateField,
    values.vipDhcpAllocation,
  ]);
```

jkilzi:
The popover for the LB bullet is not relevant anymore

src/components/clusterConfiguration/NetworkConfiguration.tsx
jkilzi: never mind I managed to make it work :+1:

src/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -0,0 +1,111 @@
...
+    >
+      {'Hosts outside of this range will not be included in the new cluster.'}
+    </Alert>
+  );
+};
+
+export interface AvailableSubnetsControlProps {
+  hostSubnets: HostSubnets;
+  hosts: Host[];
+  onChange?: SelectFieldProps['onChange'];
```
jtomasek, line 65: Not needed any more

src/components/clusterWizard/networkingSteps/ManagedNetworkingControlGroup.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { RadioProps, Split, SplitItem } from '@patternfly/react-core';
+import { RadioField } from '../../ui/formik/RadioField';
+import { NetworkConfigurationValues } from '../../../types/clusters';
+
+export interface ManagedNetworkingControlGroupProps {
+  disabled: boolean;
+  onChange?: (nextValue: NetworkConfigurationValues['networkingType']) => void;
```
jtomasek, line 8: Not needed any more

jtomasek (changes requested):

src/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -0,0 +1,104 @@
...
+                  label: hn.humanized,
+                  value: hn.humanized,
+                  id: `form-input-hostSubnet-field-option-${index}`,
+                })),
+            ]
+          : [{ label: 'No subnets are currently available', value: NO_SUBNET_SET }]
+      }
+      getHelperText={getHelperText}
+      isDisabled={!hostSubnets.length}
+      isRequired
```
jtomasek, line 101: This should be true only for `networkingType === 'clusterManaged'`

jtomasek (changes requested):
2 Nits, works great

jtomasek (approved):

PR #551, opened 2021-04-19
BZ 1946579 Add OCS manual disk formatting hint

jtomasek:
Screenshot 2021-04-19 at 16 21 34


nirfarkas:
"All installation disks are going to be used for local storage.
The list of the disks is available under each host row in the table below.
Please format the disks before installing"

mareklibra:
The group of warnings and info messages is growing. I think we should rethink that in the mid-term.
For now, would it make sense to merge those two disk-related warnings into one?

nirfarkas:
@mareklibra You're right! I think we should be careful with not overloading the UI with these.
This warning only appears when OCS is checked, so we can't merge it with the bootable warning.

mareklibra (approved):

mareklibra (approved):

PR #536, opened 2021-04-07
MGMT-4752: Add CNV operator

mareklibra:
TODO:
- [x] depends on #554
- [x] retrieve operator requirements dynamically  - change API MGMT-4220, [MGMT-4477](https://issues.redhat.com/browse/MGMT-4477) - https://github.com/openshift/assisted-service/pull/1508 , https://github.com/openshift/assisted-service/pull/1515
- [x] tweak design & popover wording
- [x] Unleash in OCM - uhc-portal 2370

Just temporarily placed on the BM Discovery step until we have profiles.

PR #556, opened 2021-04-21
MGMT-4671: Use preflight cluster requirements

mareklibra:
Depends on:
- [x] Add CNV operator #536

### DIFF
--- a/src/components/AddHosts/AddHosts.tsx
+++ b/src/components/AddHosts/AddHosts.tsx
@@ -12,6 +12,7 @@ import { useDispatch } from 'react-redux';
 import { getErrorMessage, handleApiError, installHosts } from '../../api';
 import { addAlert } from '../../features/alerts/alertsSlice';
 import { updateCluster } from '../../features/clusters/currentClusterSlice';
+import { ClusterPreflightRequirementsContextProvider } from '../clusterConfiguration/ClusterPreflightRequirementsContext';
 import { hasKnownHost } from '../hosts/utils';
 import { ToolbarButton, ToolbarSecondaryGroup } from '../ui';
 import Alerts from '../ui/Alerts';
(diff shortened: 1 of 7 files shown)


src/components/AddHosts/AddHosts.tsx
```diff
@@ -42,41 +43,43 @@ const AddHosts: React.FC = () => {
...
-            </ToolbarButton>
-            <ToolbarSecondaryGroup>
-              <EventsModalButton
-                id="cluster-events-button"
-                entityKind="cluster"
-                cluster={cluster}
-                title="Cluster Events"
-                variant={ButtonVariant.link}
-                style={{ textAlign: 'right' }}
+    <ClusterPreflightRequirementsContextProvider cluster={cluster}>
```
jtomasek, line 46: Seems like `AddHosts` is missing `AlertsContextProvider` so alerts would not work

src/components/clusterConfiguration/ClusterPreflightRequirementsContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+export type ClusterPreflightRequirementsContextType = {
+  preflightRequirements?: PreflightHardwareRequirements;
+};
+
+export const ClusterPreflightRequirementsContext = React.createContext<
+  ClusterPreflightRequirementsContextType
+>({});
+
+export const ClusterPreflightRequirementsContextProvider: React.FC<{ cluster: Cluster }> = ({
+  cluster,
```
jtomasek, line 20: Looks like only `cluster.id` is used so maybe that could be enough to pass as a prop instead of the whole cluster.

jtomasek (changes requested):
Just nits.

src/components/AddHosts/AddHosts.tsx
mareklibra: I see `AddHosts` to be used on two places only, both are wrapped by the `AlertsContextProvider` in the hierarchy up, so we should be safe. The OCM does not reference `AddHosts` directly.

src/components/clusterConfiguration/ClusterPreflightRequirementsContext.tsx
mareklibra: done

PR #559, opened 2021-04-25
MGMT-3454 Display write speed requirement in the UI

jkilzi:
image

### DIFF
--- a/src/components/clusterConfiguration/HostInventory.tsx
+++ b/src/components/clusterConfiguration/HostInventory.tsx
@@ -14,7 +14,7 @@ import { useFeature } from '../../features/featureGate';
 import CheckboxField from '../ui/formik/CheckboxField';
 import { isSingleNodeCluster } from '../clusters/utils';
 import DiscoveryInstructions from './DiscoveryInstructions';
-import { PopoverIcon } from '../ui';
+import { ExternalLink, PopoverIcon } from '../ui';
 import { OPERATOR_NAME_CNV } from '../../config';
 import { fileSize } from '../hosts/utils';
 import { DiscoveryImageModalButton } from './discoveryImageModal';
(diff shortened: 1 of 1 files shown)


nirfarkas:
LGTM! WDYT @afial@redhat.com?

jtomasek (approved):

src/components/clusterConfiguration/HostInventory.tsx
```diff
@@ -32,12 +32,21 @@ const HostRequirementsContent = () => {
...
   return (
-    <Text component="p">
-      Three master hosts are required with at least {master?.cpuCores || 4} CPU cores, {masterRam}{' '}
-      of RAM, and {master?.diskSizeGb || 120} GB of filesystem storage each. Two or more additional
-      worker hosts are recommended with at least {worker?.cpuCores || 2} CPU cores, {workerRam} of
-      RAM, and {worker?.diskSizeGb || 120} GB of filesystem storage each.
-    </Text>
+    <List>
+      <ListItem>
+        Supervisors: At least {master?.cpuCores || 4} CPU cores, {masterRam} RAM,{' '}
```
jtomasek, line 37: Shouldn't the rename from 'master' to 'supervisor' be done in a single focused change in cooperation with the backend? This will potentially confuse users for a long time until we remove all uses of 'master'

jtomasek (changes requested):
Shouldn't the rename from 'master' to 'supervisor' be done in a single focused change in cooperation with the backend? This will potentially confuse users for a long time until we remove all uses of 'master'

src/components/clusterConfiguration/HostInventory.tsx
jkilzi: I'm just following the mockup, does it matter that much if we leave this like it is right now?

jkilzi: Reverted that to "Masters"

jkilzi:
I changed back the text to Masters instead of Supervisors following Jiri's comment and today meeting with @nirfarkas

mareklibra (approved):

PR #564, opened 2021-04-28
MGMT-6009: Rephrase Additional requirements for CNV

mareklibra:


mareklibra:
Cc @nirfarkas

jtomasek (changes requested):
IMHO the renaming should go hand in hand with the backend + other UI parts to avoid having "master" and "supervisor" used at the same time which will be very confusing.

nirfarkas:
So I guess we should have the UI & BE change every "Master" to "Supervisor"

mareklibra:
Filing MGMT-6009, let's discuss the issue there first.

jkilzi:
Closing this since it has been implemented in v1.5.28 already as part of [MGMT-5284](https://issues.redhat.com/browse/MGMT-5284)

PR #566, opened 2021-04-29
Fix OCS disk formatting hint wording

jtomasek:


mareklibra (approved):
We should summon an all-hands meeting for that next time.

PR #572, opened 2021-04-30
BZ1954407 Show Cluster Events button in the OCM wizard card

mareklibra:


mareklibra:
---

jtomasek (approved):

jtomasek:
I would probably be better to put the button to the cluster actions at the top right. For standalone we already have it there.

PR #576, opened 2021-05-03
Add SNO disclaimet ro ClusterDetails

jtomasek:
Screenshot 2021-05-03 at 13 23 23


src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -47,6 +47,7 @@ type ClusterDetailsValues = {
   openshiftVersion: string;
   pullSecret: string;
   baseDnsDomain: string;
+  SNODisclaimer: boolean;
```
mareklibra, line 50: We should skip it from the patch and create cluster params

mareklibra (changes requested):

mareklibra (approved):

PR #578, opened 2021-05-03
MGMT-3210 Fix console troubleshooting modal for SNO

jkilzi:
Kooha-2021-05-03-16_15_03

### DIFF
--- a/src/components/clusterDetail/ConsoleModal.tsx
+++ b/src/components/clusterDetail/ConsoleModal.tsx
@@ -12,6 +12,10 @@ import { removeProtocolFromURL, stringToJSON } from '../../api/utils';
 import { ToolbarButton } from '../ui/Toolbar';
 import { InfoCircleIcon } from '@patternfly/react-icons';
 import PrismCode from '../ui/PrismCode';
+import { findHostById, unmarshallInventory } from '../hosts/utils';
+import isCIDR from 'is-cidr';
+import { Address4, Address6 } from 'ip-address';
+import { isSingleNodeCluster } from '../clusters/utils';
 
 type WebConsoleHintProps = {
   cluster: Cluster;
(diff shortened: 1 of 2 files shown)


src/components/hosts/utils.ts
```diff
@@ -156,3 +157,9 @@ export const fileSize: typeof filesize = (...args) =>
     .call(null, ...args)
     .toUpperCase()
     .replace(/I/, 'i');
+
+export const findHostById = (hosts: Host[] = [], id: string) =>
+  hosts?.find((host) => host.id === id);
+
+export const unmarshallInventory = (host: Partial<Host> | undefined) =>
```
mareklibra, line 164: shouldn't be the `host` of optional `string` type?

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -77,34 +81,102 @@ const getHostIPs = (cluster: Cluster): { [key: string]: string } => {
...
+    // We assume at this point the server will
+    // populate this value with either an IPv4 or IPv6 address
+    const isIPv4 = isCIDR.v4(machineNetworkCidr);
+    const network = isIPv4 ? new Address4(machineNetworkCidr) : new Address6(machineNetworkCidr);
+    const hostIds =
+      hostNetworks?.find((hostNetwork) => hostNetwork.cidr === machineNetworkCidr)?.hostIds || [];
+    // According to BE team, if it's a SNO cluster, there should be only one host in the cluster
+    // network with status 'installed'
+    const host = hostIds
+      ?.map((hostId) => findHostById(hosts, hostId))
```
mareklibra, line 103: this can be done via `Array.filter()`

src/components/hosts/utils.ts
mareklibra, line 164: Is `Partial` needed here?

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -77,34 +81,102 @@ const getHostIPs = (cluster: Cluster): { [key: string]: string } => {
   return hostIPAddresses;
 };
 
+const getVirtualIPs = ({
```
mareklibra, line 84: a `Cluster` is passed here, we can destructuralize later and so avoid typescript's `Partial`

src/components/clusterDetail/ConsoleModal.tsx
```diff
@@ -77,34 +81,102 @@ const getHostIPs = (cluster: Cluster): { [key: string]: string } => {
...
+  apiVip,
+  ingressVip,
+  highAvailabilityMode,
+  machineNetworkCidr = '',
+  hosts = [],
+  hostNetworks = [],
+}: Cluster) => {
+  const vipValues = { apiVip, ingressVip };
+  if (isSingleNodeCluster({ highAvailabilityMode }) && machineNetworkCidr?.length > 1) {
```
mareklibra, line 93:
```suggestion
  if (isSingleNodeCluster(cluster) && machineNetworkCidr?.length > 1) {
```

src/components/clusters/utils.ts
```diff
@@ -6,7 +6,8 @@ import {
   OperatorCreateParams,
 } from '../../api/types';
 
-export const isSingleNodeCluster = (cluster: Cluster) => cluster.highAvailabilityMode === 'None';
+export const isSingleNodeCluster = (cluster: Partial<Cluster>) =>
```
mareklibra, line 9: not needed and confusing

mareklibra (changes requested):

src/components/clusterDetail/ConsoleModal.tsx
jkilzi: it is exactly what it is doing. think about this function as a common operation you would like to perform on the hosts list

src/components/hosts/utils.ts
jkilzi: Yes this will allow a consumer to call the function without passing the whole host object (which is already quite large)

src/components/clusterDetail/ConsoleModal.tsx
jkilzi: sorry I don't understand what you mean

mareklibra: when calling that function, a `cluster` object is passed

mareklibra:
```suggestion
const getVirtualIPs = (cluster: Cluster) => {
  const { apiVip, ingressVip, ./* TBD*/... } = cluster;
  
  ....
}
```

src/components/clusters/utils.ts
jtomasek: I agree, I don't thinks this helper is going to be ever used by passing another object than a cluster as a parameter.

jtomasek (changes requested):

src/components/hosts/utils.ts
jkilzi: sorry my bad we don't need it

mareklibra (approved):

PR #581, opened 2021-05-03
BZ-1954074 Fix discovery instruction to mentions machine in SNO

jkilzi:
image

### DIFF
--- a/src/components/clusterConfiguration/DiscoveryInstructions.tsx
+++ b/src/components/clusterConfiguration/DiscoveryInstructions.tsx
@@ -1,13 +1,17 @@
 import React from 'react';
 import { Text } from '@patternfly/react-core';
 
-const DiscoveryInstructions = ({ isMultiNodeCluster = false }: { isMultiNodeCluster: boolean }) => (
+const DiscoveryInstructions = ({
+  isSingleNodeCluster = false,
+}: {
+  isSingleNodeCluster: boolean;
+}) => (
   <>
     <Text component="h3">Instructions</Text>
     <Text component="p">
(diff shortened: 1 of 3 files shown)


src/components/clusterConfiguration/DiscoveryInstructions.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
 import { Text } from '@patternfly/react-core';
 
-const DiscoveryInstructions: React.FC = () => (
+const DiscoveryInstructions = ({ isMultiNodeCluster = false }: { isMultiNodeCluster: boolean }) => (
```
mareklibra, line 4: nit: I wold stay consistent for the sake of potential `git grep` in the future, so please rename to `isSingleNodeCluster`.

mareklibra (changes requested):

src/components/clusterConfiguration/DiscoveryInstructions.tsx
jkilzi: Can do @mareklibra .

Could you please elaborate on which kind of consistency problems do you see and how they relate to `git grep`?

mareklibra: We use `*SingleNode*` in similar cases, so I prefer to flip predicates instead of the naming.

When sustaining the codebase in the future (who knows by whom), it's common practice to use `git grep` for getting oriented or finding the places to fix. In case it is not an additional complication today, let's not make the task more difficult.

mareklibra (approved):

jtomasek (approved):

PR #563, opened 2021-04-27
BZ-1939251 appends the base dns part to the hostname

jkilzi:
image

PR #583, opened 2021-05-04
BZ-1939251 Fixes duplicated dns part

jkilzi:
Is up to the user to set the FQDN for the host name


mareklibra:
Please see my https://bugzilla.redhat.com/show_bug.cgi?id=1939251#c7 comment.
We should make it clear, that the hostname is expected to be FQDN. Wdyt, @nirfarkas

jkilzi:
@mareklibra this PR reverts a bug introduced in [this PR](https://github.com/openshift-assisted/assisted-ui-lib/pull/563)
Please don't hold back the review until @nirfarkas replies, otherwise we will change the behavior the users expect normally.

I agree with you that we should explain better that the hostname **must** be the FQDN of the host, this is implicit for new users (I made this mistake my self) and we should do and effort in making this explicit. However this effort can be reflected in a separate task

mareklibra (approved):
Improvement of the UX is tracked separately.

PR #584, opened 2021-05-04
Cluster polling hook returns data, state and error

rawagner:
Just a small improvement - we dont have to use `useSelector` anymore because `useClusterPolling` returns the data we need.

### DIFF
--- a/src/components/clusters/clusterPolling.ts
+++ b/src/components/clusters/clusterPolling.ts
@@ -1,11 +1,13 @@
 import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
+import { Cluster } from '../../api';
 import { POLLING_INTERVAL } from '../../config';
 import {
   fetchClusterAsync,
   cleanCluster,
   forceReload,
   cancelForceReload,
+  RetrievalErrorType,
 } from '../../features/clusters/currentClusterSlice';
 import { selectCurrentClusterState } from '../../selectors';
 import { ResourceUIState } from '../../types';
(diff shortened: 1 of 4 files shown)


src/components/clusters/clusterPolling.ts
```diff
@@ -38,4 +42,6 @@ export const useClusterPolling = (clusterId: string) => {
       dispatch(cleanCluster());
     };
   }, [dispatch, fetchCluster]);
+
+  return [data, uiState, errorDetail];
```
jtomasek, line 46: Nit: It may be slightly more convenient to return an object, wdyt?
```suggestion
  return {cluster: data, uiState, errorDetail};
```

rawagner: Array allows to name the variables in whatever way consumer likes without extra code

ie

```js
const [foo, bar, baz] = useClusterPolling(clusterId);
```

vs
```js
const {cluster: foo, uiState: bar, errorDetail: baz} = useClusterPolling(clusterId);
```

The benefit becomes obvious when you want to use `useClusterPolling` multiple times in one component so you need to name the constants differently.
We dont have such a case yet so I dont have a strong preference over object vs array. I'll go with whatever you prefer :)

jtomasek: ok, yeah, object lets you pick items slightly easily but it is not needed in this case, so array it is

jtomasek (dismissed):

src/components/clusters/clusterPolling.ts
jkilzi: I agree with @rawagner argument in this case it

mareklibra (approved):

PR #585, opened 2021-05-05
BZ1954106 Fix updates of Formik's initialValues

mareklibra:


jtomasek (changes requested):
Calling `setInitialValues` is the same as using Formik's `resetForm` and resetForm does not need `enableReinitialize`.

We've used `resetForm` before introducing autosave to reset the form once it has been updated by a response from the server. And I agree that it is correct way to ensure the form becomes not dirty and reflects the changes which backend did.

The reason why we removed usage of resetForm is that it introduces a race condition between server responding and user typing. E.g.: User types in an IP, stops typing for 2 seconds which triggers autosave and a request is submitted. User starts typing again, and in the meantime a response from server arrives. At that point we call `resetForm` and that overrides the value which user typed while request has been handled by server.

jkilzi (commented):
lgtm

mareklibra:
The use of `resetForm` or React's state seems to be semantically equal, I am ok to use any of those.
Let's consider the rewrites.

jtomasek:
I would prefer using `resetForm` as it is a standard API of the Formik and it ensures that things will work as expected, compared to a custom solution which works exactly the same

jtomasek:
We should think about how we'll go about bugs which will be filed on the issue I mentioned above. IMHO at that point we'd have to reconsider whether we do or don't want the autosave functionality

mareklibra:
Switched to `resetForm`

src/components/clusterConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -99,6 +103,9 @@ const NetworkConfigurationForm: React.FC<{
 
       const { data } = await patchCluster(cluster.id, params);
       dispatch(updateCluster(data));
+      actions.resetForm({
+        values: getNetworkInitialValues(cluster, defaultNetworkSettings),
```
jtomasek, line 107:
```suggestion
        values: getNetworkInitialValues(data, defaultNetworkSettings),
```

src/components/clusterWizard/HostDiscovery.tsx
```diff
@@ -50,6 +57,7 @@ const HostDiscovery: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
     try {
       const { data } = await patchCluster(cluster.id, params);
       dispatch(updateCluster(data));
+      actions.resetForm({ values: getHostDiscoveryInitialValues(cluster) });
```
jtomasek, line 60:
```suggestion
      actions.resetForm({ values: getHostDiscoveryInitialValues(data) });
```

jtomasek (changes requested):

jtomasek:
I've tested this with Formik 2.2.6 and it works as expected. The collision between reseting the form and user input is present but it is a tradeof of autosave functionality.

We could update Formik to 2.2.6 as part of this change or a subsequent one.

mareklibra:
I will update Formik in #586

jtomasek (approved):

PR #591, opened 2021-05-06
Fix positioning of Events Dialog in Wizard steps

jtomasek:
The button floating did not work well with Stack and Grid used in wizard step
content. The button prevented grid from spanning full width. This change
puts the button into the grid next to wizard step title.

Screenshot 2021-05-06 at 15 39 14


mareklibra (approved):

nirfarkas:
Let's downgrade it to be a button link, it's too prominent and we're having too many CTAs on the screen anyway.
Just an idea (: can we add it as a menu item under the 'Actions" button dropdown?

PR #595, opened 2021-05-11
Mention FQDN in the host edit dialog

mareklibra:


jkilzi:
@mareklibra we discussed this issue the last Thursday with Moran as well. The user doesn't have to enter the FQDN as the hostname.
The sample DNS records that we generate later in the troubleshooting modal are better off without the FQDN.
The rational behind this decision is that the DNS itself is in charge of generating the FQDN for the hostnames-ip pairs it is authoritative upon them.

nirfarkas:
I've created https://issues.redhat.com/browse/MGMT-6470 to address the 'localhost' naming problem. I think this modal should be simplified:
<Header>
Edit host name
<Body>
Input field label - Host name
Input field content - <the current auto-generated host name> highlighted
Main CTA - 'Rename"
Secondary - 'Cancel'

For your question:
IMO no need to have any description text here

PR #599, opened 2021-05-12
MGMT-6046 Align cluster Ready status with telemetry's Ready

mareklibra:


mareklibra:
@ukalifon This is expected to break tests.
Once merge, we will not have a cluster in the `Ready` state, just in `Draft`. The tests will need to reimplement related conditions.

src/config/constants.ts
```diff
@@ -58,7 +58,7 @@ export const HOST_ROLES: HostRole[] = [
 export const CLUSTER_STATUS_LABELS: { [key in Cluster['status']]: string } = {
   'pending-for-input': 'Draft',
   insufficient: 'Draft',
-  ready: 'Ready',
+  ready: 'Draft',
```
mareklibra, line 61: Cc @ukalifon

src/components/clusters/ClusterStatus.tsx
```diff
@@ -18,10 +18,12 @@ import { CLUSTER_STATUS_LABELS } from '../../config/constants';
 
 type ClusterStatusProps = {
   status: Cluster['status'];
+  className?: string;
```
jtomasek, line 21: I don't see the `className` used anywhere?

mareklibra: Good catch. But it is used from the uhc-portal where a value is actually passed.

jtomasek (approved):

src/components/clusters/ClusterStatus.tsx
jtomasek: I've been greping uhc-portal codebase for `<ClusterStatus` and couldn't see it anywhere?

PR #601, opened 2021-05-13
Make HostsTable more generic

rawagner:
This PR changes HostsTable to make it reusable for other environments.
The table now accepts callbacks for all actions so the consumer can use its own way of opening modals.
If no callback for the action is provided, then its not shown (the table becomes read-only)

Example of consuming the table is here https://github.com/rawagner/dynamic-cim/blob/infra_hosts/src/components/Agent/AgentTable.tsx#L24

The `AgentTable` fetches Agent CR from k8s, maps the fields of the CR to format expected by `HostsTable` (we will need some util method to do this in the future). And passes the mapped hosts to `HostsTable`. Since no callbacks for the actions are provided, the table is in read-only mode.

`HostsTable` in OpenShift Console
Screenshot from 2021-05-13 11-29-12

Expanded state
Screenshot from 2021-05-13 11-29-27

### DIFF
--- a/src/components/AddHosts/InventoryAddHost.tsx
+++ b/src/components/AddHosts/InventoryAddHost.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Text, TextContent, Button } from '@patternfly/react-core';
-import HostsTable from '../hosts/HostsTable';
+import ClusterHostsTable from '../hosts/ClusterHostsTable';
 import { DiscoveryImageModalButton } from '../clusterConfiguration/discoveryImageModal';
 import { DiscoveryTroubleshootingModal } from '../clusterConfiguration/DiscoveryTroubleshootingModal';
 import InformationAndAlerts from '../clusterConfiguration/InformationAndAlerts';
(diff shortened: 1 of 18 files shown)


mareklibra (approved):
Verified that the Assisted-UI still works.

src/components/hosts/HostsTableCluster.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+type HostsTableClusterProps = {
+  cluster: Cluster;
+  columns?: (string | ICell)[];
+  hostToHostTableRow?: HostsTableProps['hostToHostTableRow'];
+  skipDisabled?: boolean;
+  setDiscoveryHintModalOpen?: HostsNotShowingLinkProps['setDiscoveryHintModalOpen'];
+};
+
+const HostsTableCluster: React.FC<HostsTableClusterProps & WithTestID> = ({
```
jtomasek, line 84: Nit: `ClusterHostsTable` might be a bit better.

jtomasek (approved):

src/components/hosts/HostsTableCluster.tsx
rawagner: sounds good!
I also wonder  - dont we want to move these generic components to some specific folder so we can have easier time identifying if its being reused in ACM/OCP or not ?

mareklibra: I would not organize that for ACM/OCP. We are a library which is used in other contexts as well (+3).

We have already "naming contexts" on the top-level export. We can put some effort to better identify what to export/hide, recently we blindly reexport almost everything.

mareklibra (approved):

PR #602, opened 2021-05-13
BZ-1928653 Fixes flooding sentry when there is and API error

jkilzi:
This PR modifies the polling mechanism so it doesn't continues fetching
when there is an API error.
Also, the 'try again' action in the `ErrorState` component will trigger a full page reload only when the received HTTP status code is 401 Unauthorized, this is done in order to force OCM to send the users to the login page so they can renew their token.

### DIFF
--- a/src/components/clusters/clusterPolling.ts
+++ b/src/components/clusters/clusterPolling.ts
@@ -30,7 +30,12 @@ export const useClusterPolling = (
 
   React.useEffect(() => {
     if (isReloadScheduled) {
-      if (![ResourceUIState.LOADING, ResourceUIState.RELOADING].includes(uiState)) {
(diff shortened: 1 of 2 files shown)


src/components/clusters/clusterPolling.ts
```diff
@@ -30,7 +30,12 @@ export const useClusterPolling = (
 
   React.useEffect(() => {
     if (isReloadScheduled) {
-      if (![ResourceUIState.LOADING, ResourceUIState.RELOADING].includes(uiState)) {
+      const bannedUIStates = [
```
mareklibra, line 33: nit: this array is basically a constant, can go outside of the effect

mareklibra (approved):

PR #604, opened 2021-05-17
WIP Refactor DiscoveryImageSummary to  DownloadIsoModal

mareklibra:

PR #608, opened 2021-05-20
Add bmc modal

mareklibra:
Depends on:
- [x] #604  DownloadIsoModal


mareklibra:
---

nirfarkas:
@mareklibra Where is this supposed to live?

mareklibra:
@nirfarkas , that will be consumed by the ACM. Recently WIP to include that in the dynamic-cim plugin (part of the OCP). There is no plan to show that dialog in the AI UI.

src/components/bmc/utils.ts
```diff
@@ -0,0 +1,57 @@
+import { AddBmcValues } from './types';
+
+export const getBareMetalHostCredentialsSecret = (
```
rawagner, line 3: shouldnt this live in `dynamic-cim` ?

mareklibra: I think those methods are generic enough to be used by various consumers which will need that k8s resource composition in their flows.

jtomasek (approved):

rawagner (approved):

PR #612, opened 2021-05-25
BZ-1953528 Don't call PATCH API if the form hasn't changed

jkilzi:
Before this PR we were calling the PATCH API on every transition from the "Cluster Details" to the "Host Discovery" (first to second step) even though the form remains pristine.
For the rest of the steps there is no PATCH API call when the form is pristine. This PR prevents calling the PATCH API when the form is pristine during the first transition only.

### DIFF
--- a/src/components/clusterWizard/ClusterDetails.tsx
+++ b/src/components/clusterWizard/ClusterDetails.tsx
@@ -181,6 +181,19 @@ const ClusterDetailsForm: React.FC<ClusterDetailsFormProps> = (props) => {
     }
   };
 
+  const handleOnNext = (
+    dirty: boolean,
+    submitForm: FormikHelpers<unknown>['submitForm'],
+    cluster?: Cluster,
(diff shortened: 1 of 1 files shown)


src/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -271,7 +284,7 @@ const ClusterDetailsForm: React.FC<ClusterDetailsFormProps> = (props) => {
                 (dirty || (cluster && canNextClusterDetails({ cluster })))
               )
             }
-            onNext={submitForm}
+            onNext={handleOnNext(dirty, submitForm, cluster)}
```
jkilzi, line 287: @mareklibra or @jtomasek in case you ask... :) Formik doesn't allow to read the dirty prop inside the `handleSubmit` callback we already have defined (upwise in this component)

mareklibra (approved):

PR #614, opened 2021-05-26
Do not bundle, transpile only

rawagner:


jkilzi:
@rawagner it is failing when it attempts to install peer dependencies, can you please take a look?

rawagner:
standalone AI PR https://github.com/openshift-assisted/assisted-ui/pull/443

jtomasek (approved):

package.json
```diff
@@ -4,17 +4,19 @@
...
     "prebuild": "run-s test",
-    "build": "microbundle-crl --no-compress --format modern,cjs --css-modules false",
-    "start": "microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
+    "build": "tsc && yarn copy",
+    "start": "tsc --watch & yarn watch-css",
+    "copy": "copyfiles \"./src/**/*.css\" \"./dist/\"",
+    "watch-css": "nodemon --watch ./src -e css --exec \"yarn copy\"",
     "start-assisted-ui": "./scripts/start-assisted-ui.sh",
     "start-uhc-portal": "./scripts/start-uhc-portal.sh",
     "sync-to-ui": "./scripts/sync-dist.sh",
```
jkilzi, line 22: @rawagner do we still need this?

jkilzi, line 22: @rawagner do we still need this script (`sync-to-ui`)? if not we can remove it and also update the README since there's no need to install `inotify` anymore  
How would you launch now the dev environment?

Right now the `start-assisted-ui` script does this: `yarn start" "yarn sync-to-ui" "cd ../assisted-ui; yarn start`
IIUC after this PR it should be only about executing `yarn start` here and `yarn start` in assisted-ui, am I right?

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "module": "esnext",
     "lib": ["dom", "esnext"],
```
jkilzi, line 5: I believe we should add `dom.iterable` here, otherwise we won't get annotations for all of [these](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.iterable.d.ts) types

Also be aware that create-react-app [base tsconfig](https://github.com/tsconfig/bases/blob/main/bases/create-react-app.json) adds it

jkilzi (changes requested):

package.json
rawagner: We still need it. The purpose of `sync-to-ui` is syncing the `dist` folder to standalone, uhc or any other project. 

This PR changes only the content of dist folder (its not bundled, but transpiled only), its not doing any changes to how we sync the dist to consumer app.

There is no change when running the dev env.
You will still run `yarn start-assisted-ui` or `yarn start-uhc-portal`

tsconfig.json
rawagner: added

jkilzi (approved):

PR #605, opened 2021-05-18
Infra wizard

rawagner:

PR #624, opened 2021-06-02
MGMT-6445: Introduce ClusterDeploymentWizard and related refactoring

mareklibra:
Depends on:
- [x] #605 InfraEnv

TODO:
- [x] nice to have: refactor `footer` for canNext-validation. But that code is recently not used.
- [x] Rethink for the Edit flow


mareklibra:
Rabased since #605 is merged.

mareklibra:
Rebased

mareklibra:
In the context of #639, we need to rethink the approach of a single `Formik` and creation of k8s resources at once (`ClusterDeployment` and `AgentClusterInstall`). I will split it per step while having the`Edit` flow on mind.

src/components/ClusterDeployment/ClusterDeploymentWizard.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+import { AlertsContextProvider } from '../AlertsContextProvider';
+import { ClusterDeploymentWizardProps, ClusterDeploymentWizardStepsType } from './types';
+import ClusterDeploymentWizardContext from './ClusterDeploymentWizardContext';
+import ClusterDeploymentDetailsStep from './ClusterDeploymentDetailsStep';
+
+const ClusterDeploymentWizard: React.FC<ClusterDeploymentWizardProps> = ({
+  className = '',
+
```
rawagner, line 9: eslint doesnt complain ? Im shocked :)

src/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { WizardFooterGenericProps } from '../ui/WizardFooter';
+import { Alert, AlertGroup, AlertVariant } from '@patternfly/react-core';
+import { CLUSTER_DEPLOYMENT_FIELD_LABELS } from './constants';
+
+type ClusterDeploymentWizardFooterProps = WizardFooterGenericProps & {
+  // additionalActions?: React.ReactNode;
+  errorFields?: string[];
+};
+
+const ValidationSection = ({ errorFields = [] }: { errorFields?: string[] }) => {
```
rawagner, line 14: why the default ?

I think the `ValidationSection` doesnt even have to render `AlertGroup` if there are no errorFields

rawagner, line 14: lets use React.FC, same for `ClusterDeploymentWizardFooter`

```suggestion
const ValidationSection: React.FC<ValidationSectionProps> = ({ errorFields }) => {
```

src/components/ClusterDeployment/ClusterDeploymentWizard.tsx
mareklibra: It's a mystery to be solved by a grand master.

src/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
mareklibra: done. I keep the default but got your point and changed.

mareklibra: done

src/components/ui/WizardFooter.tsx
```diff
@@ -0,0 +1,96 @@
...
+  onNext,
+  onBack,
+  onCancel,
+  isNextDisabled,
+  isBackDisabled,
+  leftExtraActions,
+  isSubmitting,
+  submittingText = 'Saving changes...',
+}) => {
+  const rightExtraActions = isSubmitting ? (
```
jtomasek, line 44: Should it be possible to pass `rightExtraActions` as a prop? If not it could be included directly in the return jsx instead of having it in a variable.

src/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: ClusterDeploymentWizardFooterProps) => {
+  const { alerts } = useAlerts();
+  const alertsSection = alerts.length ? <Alerts /> : undefined;
+  const errorsSection = <ValidationSection errorFields={errorFields} />;
+
+  return (
+    <WizardFooter
+      alerts={alertsSection}
+      errors={errorsSection}
+      leftExtraActions={undefined}
```
jtomasek, line 41: not needed?

jtomasek (changes requested):

src/components/ClusterDeployment/ClusterDeploymentDetails.tsx
```diff
@@ -0,0 +1,44 @@
...
+      </GridItem>
+      <GridItem span={12} lg={10} xl={9} xl2={7}>
+        <ClusterDetailsFormFields
+          toggleRedHatDnsService={toggleRedHatDnsService}
+          versions={ocpVersions}
+          defaultPullSecret={defaultPullSecret}
+          canEditPullSecret={!cluster || !cluster.pullSecretSet}
+          isSNOGroupDisabled={true}
+          forceOpenshiftVersion={undefined}
+          {...values}
```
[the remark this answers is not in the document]
mareklibra: Right, leftover from the former refactoring where the two use-cases did not share a single type for Formik's values. It's not the recent case, so fixing.

src/components/ui/WizardFooter.tsx
mareklibra: Done. Both recent usecases share the same rightExtraCtions, so there is no need to extract it.

src/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
mareklibra: Removed

src/components/clusterWizard/ClusterDetailsForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+  usedClusterNames: string[];
+
+  navigation: React.ReactNode;
+
+  moveNext: () => void;
+  handleClusterCreate: (params: ClusterCreateParams) => Promise<void>;
+  handleClusterUpdate: (clusterId: Cluster['id'], params: ClusterUpdateParams) => Promise<void>;
+};
+
+const ClusterDetailsForm: React.FC<ClusterDetailsFormProps> = (props) => {
```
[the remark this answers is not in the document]
mareklibra: I hope I do not miss something. The `ClusterDetetailsForm` is used by the `ClusterDetails` (Create cluster wizard flow). Honestly, there is no more need for that split and we could leave content of the `ClusterDetailsForm` insinde of the `ClusterDetails` but it happend to be that way because of the second round of refactorings ... Anyway, the benefit of that split can be in having 2 smaller components instead of a single bigger one.

src/components/ui/WizardFooter.tsx
```diff
@@ -0,0 +1,96 @@
...
+  submittingText?: string;
+};
+
+type WizardFooterProps = WizardFooterGenericProps & {
+  alerts?: React.ReactNode;
+  errors?: React.ReactNode;
+  leftExtraActions?: React.ReactNode;
+};
+
+const WizardFooter: React.FC<WizardFooterProps> = ({
```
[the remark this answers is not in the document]
mareklibra: The `ClusterWizardFooter` is a generic component used by the ClusterWizardFooter (from the Create cluster wizard) and the ClusterDeploymentWizardFooter.

mareklibra:
Rebased

src/components/clusterWizard/ClusterDetailsForm.tsx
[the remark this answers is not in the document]
jtomasek: Right, I agree that splitting is fine. My bad, GitHub did not load the diff of `ClusterDetails.tsx` and I thought that it is being removed.

jtomasek (approved):

PR #632, opened 2021-06-08
Fix typescript for createAsyncThunk

mareklibra:


rawagner:
I think a better would be to change the `reduxjs/toolkit` version from `^1.3.5` to `1.5.x` (uhc is using 1.5.1, standalone 1.4.0). 1.6+ is what broke the types.

mareklibra:
I just tried that fix with `1.5.1` ( https://github.com/openshift-assisted/assisted-ui/pull/444 ) and it seems to work all together.

Since we already have those patches here, we can be ready for the potential upgrade in the uhc-portal.

jkilzi (approved):
lgtm

PR #636, opened 2021-06-08
fix: errorDetail.code prop is Number in OCM but string in standalone app

jkilzi:


mareklibra (dismissed):

mareklibra:
I can remember we had status code checking on multiple places, it would be good to locate all of tehm

mareklibra (approved):

PR #638, opened 2021-06-08
MGMT-3713 Allows the user to generate a minimal ISO

jkilzi:
image

### DIFF
--- a/src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
+++ b/src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
@@ -1,0 +1,66 @@
+import React from 'react';
+import { Flex, FlexItem } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import { PopoverIcon } from '../ui';
+import { RadioField } from '../ui/formik/RadioField';
+
+const GROUP_NAME = 'imageType';
+
+const DiscoveryImageTypeControlGroupLabel = ({
+  text,
+  popoverContent,
+}: {
+  text: string;
+  popoverContent: string;
+}) => (
+  <>
+    {text}
+    <PopoverIcon
+      component={'a'}
+      variant={'plain'}
+      IconComponent={HelpIcon}
+      bodyContent={popoverContent}
+    />
+  </>
+);
+
+const DiscoveryImageTypeControlGroup = () => (
+  <Flex justifyContent={{ default: 'justifyContentFlexStart' }}>
+    <FlexItem>
+      <RadioField
+        name={GROUP_NAME}
+        id={'full-iso'}
+        value={'full-iso'}
+        label={
+          <DiscoveryImageTypeControlGroupLabel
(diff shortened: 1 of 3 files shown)


jkilzi:
@nirfarkas FYI, without blue question marks

src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+  text,
+  popoverContent,
+}: {
+  text: string;
+  popoverContent: string;
+}) => (
+  <>
+    {text}
+    <PopoverIcon
+      component={'a'}
```
mareklibra, line 19: Do not the `PopoverIcon` defaults serve our need even hear?

src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+      variant={'plain'}
+      IconComponent={HelpIcon}
```
mareklibra, line 21: I think we should unify on either `HelpIcon` or the `OutlinedQuestionCircleIcon` (no preference from my side).

src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+            popoverContent={
+              'Use when you plan to boot your host by attaching the ISO with virtual media. ' +
+              'Minimal ISO contains a smaller image size, and only part of the image will ' +
+              'download on the first boot.'
+            }
+          />
+        }
+      />
+    </SplitItem>
+    <SplitItem style={{ marginRight: '8.2rem' }} />
```
mareklibra, line 46: Can we use classes?

src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
jkilzi: @mareklibra no, unfortunately. The default PopoverIcon rendered the button in blue.

src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
jkilzi: I agree. PF used HelpIcon in their examples. Also, I chose it because it made it look closer to the mock.

src/components/clusterConfiguration/DiscoveryImageTypeControlGroup.tsx
jkilzi: The best PF class I could find, look like this:
image

mareklibra: Looks good to my eyes. I would with it unless @nirfarkas thinks differently.

mareklibra (dismissed):

mareklibra (approved):

PR #639, opened 2021-06-09
MGMT-6447 Add Networking step to the ClusterDeployment wizard

mareklibra:
Depends on:
- [x] #624 Introduce ClusterDeploymentWizard and related refactoring

Required by:
- https://github.com/rawagner/dynamic-cim/pull/17

TODO:
- [x] Allocate virtual IPs via DHCP server - recently not supported by BE, we need to disable it
- [x] Cluster defaults - can we query them? Deferred to `MGMT-7063`
- [x] Additional actions (so far just the first one Edit Host) + `can*` functions (so far just `() => true`) - will be implemented later, TODOs added
- [x] add at least Role selection action
- [x] validations in the left-side navigation - deferred to `MGMT-7062`
- [x] Fix passing of the OCP version
- [x]  fix wiping of data when Advanced is selected/unselected
- [x] disable fields which we can not `Edit` in the k8s flow
- [x] review calculation of `aiCluster.hostNetworks` - use `connectivityMajorityGroups`
- [x] set `agentClusterInstall?.spec?.networking?.machineNetwork?.[0]?.cidr`  from hostSubnet
- [x] spruce up CSS - esp. row spacing
- [ ] form-based validations and transitions (i.e. subnet is required)

### DIFF
--- a/src/components/clusterConfiguration/NetworkConfigurationFormFields.tsx
+++ b/src/components/clusterConfiguration/NetworkConfigurationFormFields.tsx
@@ -1,18 +1,16 @@
 import React from 'react';
 import { Form } from 'formik';
 import { TextContent, Text } from '@patternfly/react-core';
-import { HostSubnets } from '../../types/clusters';
-import NetworkConfiguration from './NetworkConfiguration';
+import NetworkConfiguration, { NetworkConfigurationProps } from './NetworkConfiguration';
 import ClusterSshKeyFields from './ClusterSshKeyFields';
-import { Cluster } from '../../api';
 
-const NetworkConfigurationFormFields: React.FC<{
-  cluster: Cluster;
-  hostSubnets: HostSubnets;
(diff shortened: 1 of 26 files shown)


mareklibra:
**Work-in-progress** screenshot:

src/components/clusterConfiguration/NetworkConfigurationFormFields.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Form } from 'formik';
```
jtomasek, line 2: I think you want import 'Form' from patternfly instead of formik. That will fix the styling because PF Form component has pf-c-form class which adds spacing between form groups.

mareklibra: Thank you!

mareklibra:
Rebased

jtomasek (approved):

PR #652, opened 2021-06-21
BZ-1931017 NTP sync should be resolved at Networking step

jkilzi:
[Discussion](https://bugzilla.redhat.com/show_bug.cgi?id=1931017)
[Mocks](https://issues.redhat.com/browse/MGMT-4786)

https://user-images.githubusercontent.com/77008341/122877867-c1a2bb80-d33f-11eb-889d-99303d95b527.mp4

### DIFF
--- a/src/components/hosts/HostStatus.tsx
+++ b/src/components/hosts/HostStatus.tsx
@@ -1,5 +1,13 @@
-import React from 'react';
-import { Popover, Button, ButtonVariant, Text, TextContent } from '@patternfly/react-core';
+import React, { PropsWithChildren } from 'react';
+import {
+  Popover,
+  Button,
+  Text,
+  TextContent,
+  FlexItem,
+  Flex,
+  ButtonProps,
+} from '@patternfly/react-core';
 import {
   global_danger_color_100 as dangerColor,
   global_warning_color_100 as warningColor,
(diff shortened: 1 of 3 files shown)


src/components/hosts/HostStatus.tsx
```diff
@@ -184,34 +252,47 @@ const HostStatus: React.FC<HostStatusProps> = ({
...
+              {titleWithProgress}
+            </WithHostStatusPopover>
+          }
+          falsy={<FlexItem className={'pf-u-mb-0'}>{titleWithProgress}</FlexItem>}
+        />
+
+        <RenderIf condition={Boolean(sublabel)}>
+          <FlexItem
+            className={'pf-u-font-size-xs'}
+            style={{ marginTop: 'calc(-1 * var(--pf-l-flex--spacer--xs))' }}
```
jkilzi, line 281: Dear reviewers,
This is the only place I had to tweak the style, honestly, I see no benefit in creating a CSS class only for this case.
IMHO, should this "tweak" become something widely used in our codebase, only then would be worth creating a CSS class for it.

mareklibra: It's a matter of code style. So far we have created a class for everything, avoiding direct styling. My preference is to keep that approach.

src/components/ui/RenderIf.tsx
```diff
@@ -6,3 +6,9 @@ interface RenderIfProps {
 
 export const RenderIf = ({ condition = false, children }: PropsWithChildren<RenderIfProps>) =>
   condition ? <>{children}</> : null;
+
+export const RenderIfElse = (props: {
```
mareklibra, line 10: nit: Honestly, I do not see much benefit of those components. But I am even ok with keeping them.

mareklibra: I would just not use it for more complex subtrees because of performance reasons (both branches are evaluated every time).

src/components/hosts/HostStatus.tsx
jkilzi: What's the reason to use classes for styling everything? The whole idea behind the CSS classes is to allow to reuse of style rules. In this case, there's nothing to be reused.

src/components/ui/RenderIf.tsx
jtomasek: No strong opinion here.

src/components/hosts/HostStatus.tsx
jtomasek: Honestly I'd get rid of the negative margin altogether and keep the styling intact. Alternatively we'd have to tweak the line heights for sublabel and label but IMHO it is not needed.

jkilzi: So what do we want? 
Keep using CSS classes even for non-reusable rules? How many times are you gonna write a CSS class that does `margin` this `padding` that just to override it in a specific place?

Right now we have something that is very close to the mock, is it worth throwing that away just because there's an explicit coding convention around the CSS?

jkilzi: Currently looks like this:
image

Without the margin-top override looks like this:
image

jtomasek: In general I am not opposed to adding style to a component directly.
I like to follow these rules:
1/ use vanilla PF as much as possible
2/ define classes for reused overrides to avoid duplication
3/ small tweaks using 'style' in single place are fine as it simplifies eventual removing, we don't have to open another file to see the tweaks applied, no need to invent className name

jkilzi: @jtomasek then I think we are squared here, I followed the same principle.

jtomasek (approved):

PR #660, opened 2021-06-23
chore: do not fail if dist folder doesn't exist yet

jkilzi:
`rsync` fails if the `dist` folder doesn't exist, this happens normally after setting up a local fork of the project. This PR creates an empty `dist` folder only if it doesn't exist already.

### DIFF
--- a/scripts/sync-dist.sh
+++ b/scripts/sync-dist.sh
@@ -9,24 +9,31 @@ NPM_PROJECT=openshift-assisted-ui-lib
 
 function sync {
   if [ -d "${ASSISTED_UI_ROOT}" ]; then
-    rsync -avz --delete ${BUILD_OUTPUT} ${ASSISTED_UI_ROOT}/node_modules/${NPM_PROJECT}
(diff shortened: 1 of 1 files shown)


scripts/sync-dist.sh
```diff
@@ -9,24 +9,31 @@ NPM_PROJECT=openshift-assisted-ui-lib
...
+    echo Sync to "${UHC_PORTAL}"/node_modules/${NPM_PROJECT} skipped
   fi
 }
 
-sync
-cp ./package.json ${ASSISTED_UI_ROOT}/node_modules/${NPM_PROJECT}/package.json || true
-cp ./package.json ${UHC_PORTAL}/node_modules/${NPM_PROJECT}/package.json || true
+function main {
+  if [ ! -d "${BUILD_OUTPUT}" ]; then
+    mkdir "${BUILD_OUTPUT}"
```
mareklibra, line 26: nit: When doing that, the `-p` param would fit here as well

mareklibra (approved):

PR #666, opened 2021-06-30
Capital letter for Hostname in edit host

danielerez:
Changed 'Requested hostname' to 'Requested Hostname' for keeping consistency.


y0n1:
shouldn't it be "host name"?

danielerez:
> shouldn't it be "host name"?

I think a single word makes sense in this context, but just wanted to align it with "Discovered Hostname":
https://github.com/openshift-assisted/assisted-ui-lib/blob/6d16c3f1ed3b29bf4d6aa0f894f74f7d6923d3f9/src/components/hosts/EditHostForm.tsx#L110
@jtomasek WDYT

jkilzi (approved):

PR #667, opened 2021-07-01
Allow passing custom class to HostsTable

jtomasek:


rawagner (approved):

src/components/hosts/HostsTable.tsx
```diff
@@ -16,6 +16,7 @@ import {
   RowWrapper,
 } from '@patternfly/react-table';
 import { ExtraParamsType } from '@patternfly/react-table/dist/js/components/Table/base';
+import classnames from 'classnames';
```
jkilzi, line 19: @jtomasek FYI PF has a utility function called `css` that uses this lib and does exactly the same :)

PR #669, opened 2021-07-02
Ease typescript in DownloadISOProps

mareklibra:


jkilzi:
@mareklibra why was this needed?

mareklibra:
@jkilzi In the ACM flow, those fields are not mandatory.

jkilzi (approved):

PR #678, opened 2021-07-12
Add cim/components module

jtomasek:


src/cim/components/common/ValidatedConditionAlert.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+const ValidatedConditionAlert = ({ condition }: ValidationAlertProps) => {
+  const { status, message } = condition;
+  if (status === 'True') return null;
+  if (status === 'Unknown') {
+    return <Alert variant={AlertVariant.info} title={message} isInline />;
+  }
+  // const [title, ...messages] = message.split(/[:,]/);
+  const [title, ...messages] = message.split(/:|\.,/);
```
mareklibra, line 16: A highway to hell.

src/cim/components/common/ValidatedConditionAlert.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Alert variant={AlertVariant.warning} title={title} isInline>
+      {messages.length && (
```
mareklibra, line 19: what about not rendering the Alert at all in case of no-messages?

src/cim/components/helpers/conditions.ts
```diff
@@ -0,0 +1,11 @@
+import { AgentK8sResource } from '../../types/k8s/agent';
+import { AgentClusterInstallK8sResource } from '../../types/k8s/agent-cluster-install';
+import { StatusCondition } from '../../types/k8s/common';
+
+export const getClusterValidatedCondition = (resource: AgentClusterInstallK8sResource) =>
+  resource.status?.conditions.find((c) => c.type === 'Validated') as StatusCondition<'Validated'>;
```
mareklibra, line 6: nit: to be on the safe side: `conditions?.find()`

mareklibra (commented):
Just a few of small comments. I know it's mostly copy&paste from one project to another, but let's improve it.

src/cim/components/common/ValidatedConditionAlert.tsx
jtomasek: I agree, I am not sure if there is a better way to parse the messages right now. There is an open Jira requesting support for validations_info data structure from the backend.

src/cim/components/common/ValidatedConditionAlert.tsx
jtomasek: I believe that the alert should not be rendered only in ase when the condition status is 'True' which is handled in the component.

src/cim/components/helpers/conditions.ts
jtomasek: done

mareklibra (dismissed):

mareklibra:
This needs rebase after https://github.com/openshift-assisted/assisted-ui-lib/pull/689

mareklibra (approved):

PR #699, opened 2021-07-26
Add Failed installation alert and actions to failed cluster

jtomasek:


src/common/components/ui/EventsList.tsx
```diff
@@ -88,6 +88,7 @@ const EventsList: React.FC<EventsListProps> = ({ events }) => {
 
   return (
     <Table
+      className="events-table"
```
rawagner, line 91: what is this className ?

jtomasek: I am using the class to override console styles which break the tables. https://github.com/rawagner/dynamic-cim/pull/35/files#diff-50f6b8ebc22c8459a2e7f252ad99d7ce3082b6490f7fefe437f1c0a1e16e9d62R3

rawagner: than it should be passed as parameter. Its needed only in context of OCP, not ACM

rawagner (approved):

PR #700, opened 2021-07-27
Update Create Cluster wizard for ACM integration

rawagner:
In ACM, we integrate only the form parts of wizard steps so I had to split the components.
I've also changed the props to not accept rest API objects, but k8s objects.

dynamic CIM related PR https://github.com/rawagner/dynamic-cim/pull/36


src/cim/components/helpers/toAssisted.ts
```diff
@@ -102,3 +103,13 @@ export const getAICluster = ({
 
   return aiCluster;
 };
+
+export const getOCPVersions = (
```
jtomasek, line 107: This function would better fit into it's own module (`versions` maybe?). `toAssisted` is a module for utils converting k8s resources into REST API based resources.

jtomasek (changes requested):

src/cim/components/helpers/toAssisted.ts
rawagner: ok, moved to `versions`

jtomasek (approved):

PR #703, opened 2021-07-28
NO-ISSUE: Updates swagger.yaml

jkilzi:
*Merge last!*

This PR syncs everything in the swagger.yaml with the one from assisted-service.
Please be aware that the BE API declares now a type called **InfraEnv**. We also had such a type in the CIM folder which wasn't being consumed by anything (please double-check me). In order to pass CI, I had to rename our type to **CIMInfraEnv**, hope this will be good enough.

### DIFF
--- a/src/cim/types/k8s/index.ts
+++ b/src/cim/types/k8s/index.ts
@@ -1,4 +1,4 @@
-export * from './infra-env';
+export * from './infra-env-k8s-resource';
 export * from './cluster-deployment';
 export * from './agent';
 export * from './agent-cluster-install';
--- a/src/cim/types/k8s/infra-env-k8s-resource.ts
+++ b/src/cim/types/k8s/infra-env-k8s-resource.ts
@@ -1,6 +1,6 @@
 import { K8sResourceCommon, Selector } from 'console-sdk-ai-lib';
 
-export type InfraEnv = K8sResourceCommon & {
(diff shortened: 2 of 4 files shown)


src/cim/types/k8s/cim-infra-env.ts
```diff
@@ -1,6 +1,6 @@
 import { K8sResourceCommon, Selector } from 'console-sdk-ai-lib';
 
-export type InfraEnv = K8sResourceCommon & {
+export type CIMInfraEnv = K8sResourceCommon & {
```
jkilzi, line 3: @jtomasek, @mareklibra , @rawagner Please review this change ^

rawagner: A better name would be
```suggestion
export type InfraEnvK8sResource = K8sResourceCommon & {
```

which aligns with name of other k8s resources

rawagner (changes requested):

src/cim/types/k8s/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from './infra-env';
+export * from './cim-infra-env';
```
jkilzi, line 1:
```suggestion
export * from './infra-env-k8s-resource;
```

jgyselov (approved):

rawagner (approved):

PR #713, opened 2021-08-04
BZ-1971759: Disk size calculation includes the read-only bytes from DVD-ROM storage

jkilzi:
Signed-off-by: jkilzi <jkilzi@redhat.com>

### DIFF
--- a/src/common/components/hosts/hardwareInfo.ts
+++ b/src/common/components/hosts/hardwareInfo.ts
@@ -1,6 +1,6 @@
 import Humanize from 'humanize-plus';
 import { Disk, Inventory } from '../../api';
-import { DASH } from '../constants';
+import { DASH, OpticalDiskDriveType } from '../constants';
 import { HumanizedSortable } from '../ui/table/utils';
 import { fileSize } from './utils';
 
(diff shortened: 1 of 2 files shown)


mareklibra (approved):

src/common/components/hosts/hardwareInfo.ts
```diff
@@ -21,8 +21,9 @@ export type SimpleHardwareInfo = {
 export const getMemoryCapacity = (inventory: Inventory): number =>
   inventory.memory?.usableBytes || 0;
 export const getDiskCapacity = (inventory: Inventory): number =>
-  inventory.disks?.reduce((diskSize: number, disk: Disk) => diskSize + (disk.sizeBytes || 0), 0) ||
-  0;
+  inventory.disks
+    ?.filter((disk) => disk.driveType !== 'ODD')
```
mareklibra, line 25: nit: A new constant would be better here. That way, we a future maintainer can find that faster and reliably.

jkilzi: Done, 10x

mareklibra (approved):

PR #714, opened 2021-08-04
MGMT-7381: Make cluster events toolbar fixed

jgyselov:
https://issues.redhat.com/browse/MGMT-7381

https://user-images.githubusercontent.com/87187179/128347644-13831dfa-78f5-4c96-887c-4acc43c2e259.mp4

### DIFF
--- a/src/common/components/ui/ClusterEventsList.tsx
+++ b/src/common/components/ui/ClusterEventsList.tsx
@@ -10,7 +10,7 @@ import ClusterEventsToolbar, { getInitialClusterEventsFilters } from './ClusterE
 export type ClusterEventsListProps = {
   events: EventList;
   cluster: Cluster;
-  className: string;
+  className?: string;
 };
 
 const filterEvents = (
--- a/src/common/components/ui/ClusterEventsList.tsx
+++ b/src/common/components/ui/ClusterEventsList.tsx
@@ -10,6 +10,7 @@ import ClusterEventsToolbar, { getInitialClusterEventsFilters } from './ClusterE
 export type ClusterEventsListProps = {
   events: EventList;
   cluster: Cluster;
+  className: string;
 };
 
 const filterEvents = (
@@ -33,7 +34,7 @@ const filterEvents = (
     );
 };
 
-const ClusterEventsList: React.FC<ClusterEventsListProps> = ({ events, cluster }) => {
+const ClusterEventsList: React.FC<ClusterEventsListProps> = ({ events, cluster, className }) => {
   const [filters, setFilters] = React.useState<ClusterEventsFiltersType>(() =>
     getInitialClusterEventsFilters(cluster),
   );
@@ -69,7 +70,7 @@ const ClusterEventsList: React.FC<ClusterEventsListProps> = ({ events, cluster }
           }
         />
       ) : (
-        <EventsList events={filteredEvents} />
+        <EventsList events={filteredEvents} className={className} />
       )}
     </>
   );
--- a/src/common/components/ui/ClusterEventsToolbar.css
+++ b/src/common/components/ui/ClusterEventsToolbar.css
@@ -1,3 +1,12 @@
 #checkbox-deselect-all-hosts-action {
   visibility: hidden;
 }
+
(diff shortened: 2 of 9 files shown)


src/common/components/ui/ClusterEventsToolbar.css
```diff
@@ -1,3 +1,12 @@
 #checkbox-deselect-all-hosts-action {
   visibility: hidden;
 }
+
+#clusters-events-toolbar {
```
rawagner, line 5: I realize we do not follow BEM closely in this lib but we should.
Please change the selector to class name.

Also once you're at it, can you change the id selector above to class name selector too ? This is the only file with id selectors so it would be nice to get rid of them all :)

jgyselov: It appears the first one is never used. Should I remove it?

rawagner:
Im wondering if the better solution wouldnt be to have the `EventsList` in scrollable instead of the whole modal content. That way you probably wouldn't have to do all the magic on the `ClusterEventsToolbar` css.

src/common/components/ui/EventsModal.css
```diff
@@ -0,0 +1,8 @@
+.events-modal .pf-c-modal-box__body {
```
rawagner, line 1: instead of this selector, it would be better to wrap the `EventListFetch` in div with className, something like
```js
<div className="events-modal__body">
  <EventListFetch
    hostId={hostId}
    cluster={cluster}
    entityKind={entityKind}
    onFetchEvents={onFetchEvents}
  />
</div>
```

src/common/components/ui/EventsModal.css
```diff
@@ -0,0 +1,8 @@
...
+  padding-top:0;
+  display: flex;
+  flex-direction: column;
+}
+.events-modal .events-list-wrapper {
```
rawagner, line 6: I'd rather see that the `events-list-wrapper` style will be passed as className prop to `EventListFetch` and ends up applied to `EventsList` table component.

And according to BEM it would be named something like `events-modal__event-list`

jgyselov: The table needs a wrapper (I have renamed it to `events-modal__event-list`) otherwise it doesn't scroll. But everything else I was able to implement, I hope.

src/common/components/ui/ClusterEventsList.tsx
```diff
@@ -10,6 +10,7 @@ import ClusterEventsToolbar, { getInitialClusterEventsFilters } from './ClusterE
 export type ClusterEventsListProps = {
   events: EventList;
   cluster: Cluster;
+  className: string;
```
rawagner, line 13: should be optional
```suggestion
  className?: string;
```

src/common/components/ui/EventsList.tsx
```diff
@@ -43,9 +43,10 @@ const getLabelIcon = (severity: Event['severity']) => {
 
 export type EventsListProps = {
   events: EventList;
+  className: string;
```
rawagner, line 46: sam here, optional 
```suggestion
  className?: string;
```

src/common/components/ui/eventsModal.tsx
```diff
@@ -70,6 +72,7 @@ export const EventsModal: React.FC<EventsModalProps> = ({
     <Modal
       title={title}
       isOpen={isOpen}
+      hasNoBodyWrapper={true}
```
rawagner, line 75:
```suggestion
      hasNoBodyWrapper
```

src/common/types/events.ts
```diff
@@ -26,4 +26,5 @@ export type EventListFetchProps = EventFetchProps & {
     onSuccess: (data: EventList) => void,
     onError: (message: string) => void,
   ) => Promise<void>;
+  className: string;
```
rawagner, line 29:
```suggestion
  className?: string;
```

rawagner (commented):
just some nits, otherwise looks great!

rawagner (approved):

PR #716, opened 2021-08-08
MGMT-6645: Long cluster name + base domain breaks cluster list

DanielleChason:
https://issues.redhat.com/browse/MGMT-6645

This is the way the table (with max cluster name and max base domain) looked before:
MGMT-6645 before
The last columns and the kebab menu are missing. 

And this is the way it looks now, after the changes:
MGMT-6645 after


jkilzi:
@DanielleChason we'll need to apply the same fix to the HostsTable

jkilzi (approved):

src/ocm/components/clusters/ClustersTable.tsx
```diff
@@ -28,17 +31,31 @@ interface ClustersTableProps {
   deleteCluster: (id: string) => void;
 }
 
-const columnConfig = {
+type TablePropsCellType = TableProps['cells'][0];
```
jkilzi, line 34: Maybe we should take this type to a commonplace, so it can be reused later by any other component that consumes the Table from PF.

PR #717, opened 2021-08-08
BZ-1990011: Adv. Networking checkbox is enabled despite defaults aren't changed

jkilzi:
[BZ-1990011](https://bugzilla.redhat.com/show_bug.cgi?id=1990011)

### DIFF
--- a/src/common/components/clusterConfiguration/utils.ts
+++ b/src/common/components/clusterConfiguration/utils.ts
@@ -59,7 +59,7 @@ export const isAdvNetworkConf = (cluster: Cluster, defaultNetworkSettings: Clust
   cluster.clusterNetworkCidr !== defaultNetworkSettings.clusterNetworkCidr ||
   cluster.clusterNetworkHostPrefix !== defaultNetworkSettings.clusterNetworkHostPrefix ||
   cluster.serviceNetworkCidr !== defaultNetworkSettings.serviceNetworkCidr ||
-  cluster.networkType !== 'OpenShiftSDN';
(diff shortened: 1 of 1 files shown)


src/common/components/clusterConfiguration/utils.ts
```diff
@@ -59,7 +59,7 @@ export const isAdvNetworkConf = (cluster: Cluster, defaultNetworkSettings: Clust
   cluster.clusterNetworkCidr !== defaultNetworkSettings.clusterNetworkCidr ||
   cluster.clusterNetworkHostPrefix !== defaultNetworkSettings.clusterNetworkHostPrefix ||
   cluster.serviceNetworkCidr !== defaultNetworkSettings.serviceNetworkCidr ||
-  cluster.networkType !== 'OpenShiftSDN';
+  (Boolean(cluster.networkType) && cluster.networkType !== 'OpenShiftSDN');
```
mareklibra, line 62: nit: `(!!cluster.networkType && cluster.networkType !== 'OpenShiftSDN')`

mareklibra (approved):

src/common/components/clusterConfiguration/utils.ts
jkilzi: I think we both agree that `!!x` and `Boolean(x)` are the same. BTW, nowadays in terms of performance, they are also the same.
Perhaps we can agree that we disagree on this, but don't you think using `Boolean(x)` is more explicit? (specially helpful for newcomers)

mareklibra: I think we both agree that it's better to agree on those little things which can easily generate a disagreement.

PR #726, opened 2021-08-11
BZ-1940475: The first wizard screen opens when a user returns to a cluster in "Ready" state

jgyselov:
https://bugzilla.redhat.com/show_bug.cgi?id=1940475
We can use `Cluster.status` to determine on which step should the wizard open - `cluster-detail`, `host-discovery`, or `review`.


jtomasek:
Note that the original logic was introduced to avoid entering first wizard step (cluster details) after creating the cluster. Please verify that when a new cluster is created, user is redirected to `clusters/:id` route and the 'host-discovery' step is the active one.

jgyselov:
> Note that the original logic was introduced to avoid entering first wizard step (cluster details) after creating the cluster. Please verify that when a new cluster is created, user is redirected to `clusters/:id` route and the 'host-discovery' step is the active one.

@jtomasek It works exactly like that.

jkilzi (approved):

PR #731, opened 2021-08-16
BZ-1992575: In case of media disconnection, the UI gets stuck on first page

jgyselov:
[BZ-1992575](https://bugzilla.redhat.com/show_bug.cgi?id=1992575)


src/common/components/hosts/HostStatus.tsx
```diff
@@ -138,6 +138,8 @@ const HostStatusPopoverContent: React.FC<HostStatusPopoverContentProps> = (props
           {statusDetails}
           <br />
           {toSentence(statusInfo)}
+          <br />
+          {statusInfo.includes('ISO') && 'Please generate a new minimal ISO and reboot the hosts.'}
```
jtomasek, line 142: A source of the problem is failing host validation. If the validationInfo message is not broad enough the UI codebase has HOST_VALIDATION_FAILURE_HINTS which lets us add additional description to the failing validation.

jkilzi (changes requested):

src/common/components/hosts/HostStatus.tsx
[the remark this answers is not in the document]
jgyselov: I didn't realize we had validations for packet loss. It sounds like a better idea than what we have now, let me see what I can do.

src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -286,7 +286,8 @@ export const getWizardStepClusterStatus = (
     const allHostsReady = (cluster?.hosts || []).every((host) =>
       allowedStatuses.includes(getWizardStepHostStatus(host, wizardStepId)),
     );
-    return allHostsReady &&
+
+    return (wizardStepId === 'cluster-details' || allHostsReady) &&
```
jtomasek, line 290: We should keep this function as simple as possible. All the wizard step variations and specifics should be covered inside `wizardStepsValidationsMap`. This specific problem we have - being unable to move to the next step from cluster-details step - is caused by incorrect `clusterDetailsStepValidationsMap.hosts.allowedStatuses`. No host validation failure should prevent us from moving from cluster-details step.

The easiest solution should be to add 'insufficient' to the `clusterDetailsStepValidationsMap.hosts.allowedStatuses`

jtomasek (changes requested):

src/ocm/components/clusterWizard/wizardTransition.ts
jtomasek: Or even easier, setting `clusterDetailsStepValidationsMap.hosts.allowedStatuses` to `[]` should have the same effect

jgyselov: But the host status is `error`. Should we add that too/instead? @jtomasek

jkilzi: lgtm

jgyselov:
@jkilzi @jtomasek I tried to implement your suggestions. Please take a look when you have the time.

src/common/components/hosts/HostStatus.tsx
```diff
@@ -138,6 +152,15 @@ const HostStatusPopoverContent: React.FC<HostStatusPopoverContentProps> = (props
           {statusDetails}
           <br />
           {toSentence(statusInfo)}
+          <br />
+          {hints
+            .filter((hint) => hint)
```
jkilzi, line 157: What is this line filtering exacly?

jgyselov: Empty strings so that we don't get a bunch of empty lines in case there are multiple failed validations.

src/common/components/hosts/HostStatus.tsx
```diff
@@ -106,10 +110,20 @@ type HostStatusPopoverContentProps = ValidationInfoActionProps & {
...
+  const hints: string[] = [];
+
+  if (validationsInfo?.network) {
+    validationsInfo.network.forEach((element) => {
+      if (element.status == 'failure') {
+        hints.push(HOST_VALIDATION_FAILURE_HINTS[element.id]);
+      }
+    });
+  }
+
```
jkilzi, line 126: We don't need this, the `<HostValidationGroups />` does this as well.

src/common/components/hosts/HostStatus.tsx
```diff
@@ -138,6 +152,15 @@ const HostStatusPopoverContent: React.FC<HostStatusPopoverContentProps> = (props
...
           {toSentence(statusInfo)}
+          <br />
+          {hints
+            .filter((hint) => hint)
+            .map((hint, index) => (
+              <span key={index}>
+                {hint}
+                <br />
+              </span>
+            ))}
```
jkilzi, line 163: neither this...

src/common/config/constants.ts
```diff
@@ -171,7 +171,8 @@ export const HOST_VALIDATION_FAILURE_HINTS: { [key in HostValidationId]: string
   'compatible-with-cluster-platform': '',
   'has-default-route': '',
   'sufficient-network-latency-requirement-for-role': '',
-  'sufficient-packet-loss-requirement-for-role': '',
+  'sufficient-packet-loss-requirement-for-role':
+    'Please generate a new minimal ISO and reboot the hosts.',
```
jkilzi, line 175: The message is coming from the BE, so they need to change it.

src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -283,9 +276,12 @@ export const getWizardStepClusterStatus = (
     const { groups, validationIds } = wizardStepsValidationsMap[wizardStepId].cluster;
     const { softValidationIds } = wizardStepsValidationsMap[wizardStepId];
     const { allowedStatuses } = wizardStepsValidationsMap[wizardStepId].host;
-    const allHostsReady = (cluster?.hosts || []).every((host) =>
-      allowedStatuses.includes(getWizardStepHostStatus(host, wizardStepId)),
+    const allHostsReady = (cluster?.hosts || []).every(
+      (host) =>
+        allowedStatuses.length == 0 ||
```
jkilzi, line 281:
```suggestion
        allowedStatuses.length === 0 ||
```

jkilzi (changes requested):

jkilzi (dismissed):

src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -86,7 +79,7 @@ const hostDiscoveryStepValidationsMap: WizardStepValidationMap = {
   },
   host: {
     allowedStatuses: ['known', 'disabled'],
-    groups: ['hardware'],
+    groups: ['hardware', 'network'],
```
jtomasek, line 82: I am not sure what is the intention here, the network validations should not be checked in host-discovery step. Is there some specific validation we want to check from the networking group? If so add it to the `validationIds`.

jtomasek (changes requested):

src/ocm/components/clusterWizard/wizardTransition.ts
jkilzi: IIRC the `connected` validation belongs to the network group

jtomasek: Right, that's why it is explicitly specified in `validationIds`. `groups` array is used to specify that we want to check all validations in these groups instead of having to name all of them in `validationIds` array. So in this case the 'network' group should not be included here.

jgyselov: Removed.

jkilzi (approved):

jkilzi (approved):

jtomasek (approved):

PR #733, opened 2021-08-16
MGMT-7102 Add VMWare to the wizard UI

jkilzi:
[MGMT-7102](https://issues.redhat.com/browse/MGMT-7102)

#### Here is a description of the workflow:
1. When users arrive at the hosts-discovery step. 
1. They start adding hosts, (in parallel, we poll the `GET /v2/supported-platforms` API - only in this step) 
1. Once an allowed amount of hosts had been detected, the API will return a value similar to `["baremetal", "vsphere"]` if **all** the hosts belong to the same platform (right now only 'vsphere' is available). Otherwise, it will return just `["baremetal"]`. 
1. If all the hosts belong to the same platform the UI must enable the platform integration switch. 
1. Turing the switch on changes the `cluster.platform.type` to `vsphere` and turning it off to `baremetal`. 
1. Later on other hosts might be added to the cluster, but before the users hit the install button. 
1. In that case a host validation error (`compatible-with-cluster-platform`) will fail to prevent the user from installing the cluster. 
1. The users can choose between turning off the platform integration or removing the hosts that don't belong to the platform.

image


package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "prebuild": "run-s test",
     "build": "microbundle-crl --no-compress --format modern,cjs --css-modules false",
-    "start": "microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
+    "start": "microbundle-crl watch --no-compress --format modern --sourcemap --css-modules=false",
```
jkilzi, line 17:
```suggestion
    "start": "microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
```

mareklibra (approved):

jtomasek (approved):

PR #734, opened 2021-08-16
Remove dist folder before new build

rawagner:


package.json
```diff
@@ -13,8 +13,8 @@
   },
   "scripts": {
     "prebuild": "run-s test",
-    "build": "microbundle-crl --no-compress --format modern,cjs --css-modules false",
-    "start": "microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
+    "build": "rm -rf dist && microbundle-crl --no-compress --format modern,cjs --css-modules false",
+    "start": "rm -rf dist && microbundle-crl watch --no-compress --format modern,cjs --css-modules false",
```
jkilzi, line 17: This is unrelated to your PR but if we are already discussing this stuff, then...

```suggestion
    "start": "rm -rf dist && microbundle-crl watch --no-compress --format modern --css-modules false",
```

In watch mode do we really need to create a cjs bundle?

jkilzi (commented):
FYI

package.json
rawagner: +1
cjs removed

rawagner:
> Why do we need this?

generally we should clean dist folder before building to make sure we do not have some leftovers for previous builds there

> Isn't sync-dist.sh taking care of that?

it does not, if Im looking correclty. Also you may not be using `sync-dist.sh`, just running `yarn build`

rawagner:
@jkilzi updated

jkilzi (approved):
LGTM

PR #741, opened 2021-08-18
Fix scrollbar in ClusterDeploymentWizard

mareklibra:


src/common/components/clusterWizard/ClusterWizardStep.css
```diff
@@ -12,7 +12,8 @@
 }
 
 @media (min-width: 768px) {
```
jkilzi, line 14: @jgyselov can we achieve this with pure PF classes?

jgyselov: @jkilzi Do you mean using something like `@media (min-width: $pf-breakpoint--md)`? Because that's not possible.

jkilzi: Actually, what I mean is to use some PF class/es instead of creating our own overrides

jgyselov: I don't think so, unless PF is hiding a `pf-global--overflow-inherit-on-md` class somewhere. Or if it attached classes to elements on breakpoints... But it doesn't do that. @jkilzi

jkilzi (approved):

PR #745, opened 2021-08-23
chore: Runs clean script on prebuild only + restores cjs in watch mode

jkilzi:


package.json
```diff
@@ -13,9 +13,9 @@
   },
   "scripts": {
     "clean": "rm -rf dist",
```
rawagner, line 15:
```suggestion
    "clean": "rm -rf dist/*",
```
this change should be enough - seems to work for me

jkilzi: [sync-to-dist creates the dist folder if it doesn't exist yet](https://github.com/openshift-assisted/assisted-ui-lib/blob/d8fa6a4f7610b9dda24fe5a032c7537dcb401636/scripts/sync-dist.sh#L25) is that the goal behind preserving the dist folder?

rawagner: the issue is with concurrency here.
1. `sync-to-dist` will check the `dist` folder if it exists and creates one if necessary
2. runs the rsync

meanwhile `yarn start` will delete the `dist folder` between step 1 and 2 which will make step 2. (rsync) fail because `dist` does not exist

jkilzi: > 1. sync-to-dist will check the dist folder if it exists and creates one if necessary

btw, after cloning the repo you don' have the dist/ folder, that why I've added this step to the script

> meanwhile `yarn start` will delete the `dist folder` ...

Then let's not do that. Instead, we could do the `yarn clean` thing inside the sync script, WDYT?

jkilzi: it could be wonderful if we could use some rollup-battle-tested-plugin for this, just a thinking...

PR #776, opened 2021-09-06
Update host selection step

rawagner:


src/cim/components/ClusterDeployment/ClusterDeploymentHostSelectionStep.tsx
```diff
@@ -42,78 +50,90 @@ const getInitialValues = ({
...
+  ]);
+
+  const cdName = clusterDeployment?.metadata?.name;
+  const cdNamespace = clusterDeployment?.metadata?.namespace;
+
+  const availableAgents = React.useMemo(
     () =>
-      (unfilteredMatchingAgents || []).filter((agent) => {
+      (agents || []).filter((agent) => {
         const [status] = getAgentStatus(agent);
```
mareklibra, line 120: We should filter-out agents based on `clusterDeploymentName` as well. The labels are set within UI flow only.

mareklibra: https://github.com/rawagner/dynamic-cim/pull/51/files#diff-234124f695ab67686e1b6bdfdc04f939f8f90dcd18a76d6af459e516b28c0838L256

mareklibra:
Issues:
- [ ] agents of different clusterDeploymentName should be filtere-out
- [ ] locations are not prefilled (Agents have location set, but just `No location` is listed
- [ ] the specific location-label should be omitted from label selection
- [ ] A location must be selected before agents are queried or the table shown (UX requirement, maybe worth of review recently)

src/cim/components/ClusterDeployment/ClusterDeploymentHostSelectionStep.tsx
rawagner: why cannot we reassign agents from one cluster to another if they are not reserved ?

rawagner:
> Issues:
> 
> * [ ]  agents of different clusterDeploymentName should be filtere-out

why cannot we reassign agents from one cluster to another if they are not reserved ?

> * [ ]  A location must be selected before agents are queried or the table shown (UX requirement, maybe worth of review recently)

I did this on purpose, the UX is weird and unncesessary needing to select location before seeing the agents. Also the help text on location selector says something like `leave empty so see agents in all locations`

rawagner:
I was looking at annotations instead of labels for agent's location. Should be fixed now

 - [x]  locations are not prefilled (Agents have location set, but just `No location` is listed
 - [x]  the specific location-label should be omitted from label selection

rawagner:
> > Issues:
> > 
> > * [ ]   agents of different clusterDeploymentName should be filtere-out
> 
> why cannot we reassign agents from one cluster to another if they are not reserved ?
> 
> >I can not properly say without having the late-binding available but those "reserved" labels are just an UI shortcut so far. Please consider the non-UI flow where Agents "get reserved" via `clusterDeploymentName`

the problem I have is that there is no 'reservation' mechanism in backend thus the reservation does not exist outside UI flows. What reservation system should UI consider for non-UI flows ? Something we just say that once `clusterDeploymentName` is set it is reserved ? who will know about it ? It also does not mean that the agent will be picked up by the cluster as labels do not have to match. I just do not like having a UI logic for non-UI flow which does not have any support in backend.

> 
> > * [ ]   A location must be selected before agents are queried or the table shown (UX requirement, maybe worth of review recently)
> 
> I did this on purpose, the UX is weird and unncesessary needing to select location before seeing the agents. Also the help text on location selector says something like `leave empty so see agents in all locations`
> 
>> In the context of ACM, I agree - we have to read all agents at any case. The original thought was to narrow the search result when we do not have to query everything (OCP or better ACM resource handling).

+1, also the help popup says we do not have to select anything and if we want to adhere to that we still have to query everything in OCP and ACM

mareklibra (approved):

cben:
if you removed `react-autosuggest` dep, you can probably drop this too:
```
    "@types/react-autosuggest": "^10.1.5",
```

PR #784, opened 2021-09-10
Integrate Hosts selection to ACM create wizard

rawagner:


src/cim/components/ClusterDeployment/ACMClusterDeploymentHostsStep.tsx
```diff
@@ -0,0 +1,66 @@
...
+          validationSchema={validationSchema}
+          innerRef={formRef}
+          onSubmit={noop}
+        >
+          <ClusterDeploymentHostsSelection
+            agents={agents}
+            clusterDeployment={clusterDeployment}
+            agentClusterInstall={agentClusterInstall}
+            onValuesChanged={onValuesChanged}
+            hostActions={{}} // TODO
```
mareklibra, line 51: Can we add `hostActions` property and pass it here? Let's pass temporary empty object from the ACM instead.

mareklibra (approved):

PR #786, opened 2021-09-10
Details screen filters out matching agents for cluster

rawagner:


rawagner:
included in https://github.com/openshift-assisted/assisted-ui-lib/pull/787

PR #803, opened 2021-09-16
Update approve host modal content

rawagner:
Screenshot from 2021-09-16 10-35-04


mareklibra:
I think this will need designer's touch but good for now.

mareklibra (approved):

PR #805, opened 2021-09-16
Move PreviewBadge under common

mareklibra:
TODO:
- [x] check assisted-ui app
- [x] check OCM


mareklibra:
Neither of OCM and AI UI app seem to be affected by this change. We can merge.

rawagner (approved):

PR #807, opened 2021-09-17
Pass default pull secret to formik initial values

rawagner:


mareklibra:
@rawagner Can you please link the PR where this is needed?

mareklibra (approved):

rawagner:
I dont have a PR open, I will open one agains ACM with all changes for the new ai lib. 

The reason for this change is that we have integration with ACM credentials provider in create cluster wizard. The ACM credentials provider contains pull secret so Im passing it as initial value for formik here

PR #809, opened 2021-09-17
Add cluster scale up functionality to CIM

jtomasek:
- Introduce Modal and Form for adding hosts to the cluster
- Replicate Basic/Advanced functionality from Host selection wizard step
- Refactor Host selection code to allow for reusability in scale up process
- Enhance NumberInputField

Screenshot from 2021-09-17 12-23-35
Screenshot from 2021-09-17 13-51-07

### DIFF
--- a/src/cim/components/Agent/AgentsSelectionHostCountAlerts.tsx
+++ b/src/cim/components/Agent/AgentsSelectionHostCountAlerts.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { Alert, AlertGroup, AlertVariant } from '@patternfly/react-core';
+
+const getHostCountWarningText = (hostsSelected: number) => {
+  switch (hostsSelected) {
+    case 0:
+      return 'No host is selected.';
+    case 1:
+      return 'Only 1 host is selected.';
+  }
+  return `Only ${hostsSelected} hosts are selected.`;
+};
+
+type AgentsSelectionHostCountAlertsProps = {
+  matchingAgentsCount: number;
+  selectedAgentsCount: number;
+  targetHostCount: number;
+};
+
+const AgentsSelectionHostCountAlerts = ({
+  matchingAgentsCount,
+  selectedAgentsCount,
+  targetHostCount,
+}: AgentsSelectionHostCountAlertsProps) => {
+  return (
+    <AlertGroup>
+      {selectedAgentsCount === targetHostCount && (
+        <Alert
+          variant={AlertVariant.success}
+          title={`${selectedAgentsCount} ${
+            selectedAgentsCount === 1 ? 'host' : 'hosts'
+          } selected out of ${matchingAgentsCount} matching.`}
+          isInline
+        />
+      )}
+      {selectedAgentsCount < targetHostCount && (
+        <Alert
+          variant={AlertVariant.warning}
+          title={getHostCountWarningText(selectedAgentsCount)}
+          isInline
+        />
+      )}
+    </AlertGroup>
+  );
+};
+
+export default AgentsSelectionHostCountAlerts;
--- a/src/cim/components/Agent/AgentsSelectionUtils.ts
+++ b/src/cim/components/Agent/AgentsSelectionUtils.ts
@@ -1,0 +1,37 @@
+import React from 'react';
+import { useFormikContext } from 'formik';
+import * as _ from 'lodash';
+import { AgentK8sResource } from '../../types/k8s/agent';
+import {
+  ClusterDeploymentHostsSelectionValues,
+  ScaleUpFormValues,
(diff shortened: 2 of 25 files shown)


src/cim/components/Agent/AgentsSelectionHostCountAlerts.tsx
```diff
@@ -0,0 +1,45 @@
...
+  matchingAgentsCount,
+  selectedAgentsCount,
+  targetHostCount,
+}: AgentsSelectionHostCountAlertsProps) => {
+  return (
+    <AlertGroup>
+      {selectedAgentsCount === targetHostCount && (
+        <Alert
+          variant={AlertVariant.success}
+          title={`${selectedAgentsCount} hosts selected out of ${matchingAgentsCount} matching.`}
```
rawagner, line 30: needs a singular variant if `selectedAgentsCount === 1`

src/cim/components/Agent/AgentsSelectionUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  }, [availableAgents, locations, autoSelectedHostIds]);
+
+  React.useEffect(() => {
+    const ids = matchingAgents.map((a) => a.metadata?.uid).splice(0, hostCount);
+    if (!_.isEqual(ids, autoSelectedHostIds)) {
+      setTimeout(() => setFieldValue('autoSelectedHostIds', ids, true));
+    }
+  }, [matchingAgents, setFieldValue, autoSelectedHostIds, hostCount]);
+
+  return { matchingAgents, selectedAgents, hostCount };
```
rawagner, line 36: lets return array to stay consistent with other hooks

src/cim/components/ClusterDeployment/ClusterDeploymentHostsSelectionAdvanced.tsx
```diff
@@ -1,119 +1,53 @@
...
-        <LocationsSelector agents={availableAgents} />
-      </GridItem>
-      <GridItem>
-        <LabelsSelector agents={matchingAgents} />
-      </GridItem>
-      <GridItem>
-        <AgentsSelectionTable matchingAgents={matchingAgents} />
-      </GridItem>
+      <Grid hasGutter>
+        <GridItem span={12} lg={10} xl={9} xl2={7}>
```
rawagner, line 45: this might make fields too small in ACM create wizard, when YAML editor is open. Would be good to check.

src/cim/components/ClusterDeployment/ClusterDeploymentHostsSelectionBasic.tsx
```diff
@@ -1,124 +1,74 @@
...
-      <GridItem>
-        <CheckboxField
-          idPostfix="mastersasworkers"
-          name="useMastersAsWorkers"
-          label="Run workloads on control plane (master) hosts"
-          isDisabled={hostCount < 5}
-        />
-      </GridItem>
+      <Grid hasGutter>
+        <GridItem span={12} lg={10} xl={9} xl2={7}>
```
rawagner, line 35: same here - needs to be verified in ACM/YAML editor open scenario

src/cim/components/Agent/AgentsSelectionUtils.ts
jtomasek: Hmm, returning an array forces me to explicitly specify the return value of `useAgentsAutoSelection` as `: [AgentK8sResource[], AgentK8sResource[], number]` without it typescript infers the return value as `(AgentK8sResource[] | number)[]` which does not work with the usage. Also in this case it is not expected the returned keys to need to be renamed by the consuming application so I feel using an object is better in this case.
Formik uses objects as well btw.

src/cim/components/ClusterDeployment/ClusterDeploymentHostsSelectionAdvanced.tsx
jtomasek: This component is not used in any of the views with yaml editor. I've fixed the vertical spacing of the fields.

src/cim/components/ClusterDeployment/ClusterDeploymentHostsSelectionBasic.tsx
jtomasek: I've fixed the vertical spacing and removed the largest breakpoint to avoid locations field to be too small in case when yaml editor is open

src/cim/components/Agent/AgentsSelectionHostCountAlerts.tsx
jtomasek: done

rawagner (approved):

PR #831, opened 2021-09-22
Allign peerDeps with ACM

mareklibra:


mareklibra:
None of those is actually used in the ACM flows.
Potential upgrade on the OCM side will be covered by tests.

rawagner (approved):

PR #835, opened 2021-09-22
Show peding icon for initialization if cluster does not have monitore…

rawagner:
…dOperators


jtomasek (dismissed):

jtomasek:
IIUC This will need further enhancement to handle the icon change once the cluster installation finishes.

rawagner:
> IIUC This will need further enhancement to handle the icon change once the cluster installation finishes.

I think it should be fine now

jtomasek (approved):

PR #837, opened 2021-09-22
Fix min hw text in case isSNOCluster is undefined

rawagner:


src/common/components/hosts/HostRequirements.tsx
```diff
@@ -31,7 +31,7 @@ const HostRequirements: React.FC<HostRequirementsProps> = ({
   return (
     <TextContent>
       <List>
-        {(isSNOCluster === false || undefined) && (
+        {(isSNOCluster === false || isSNOCluster === undefined) && (
```
mareklibra, line 34:
```suggestion
        {!isSNOCluster && (
```

mareklibra (approved):

PR #842, opened 2021-09-23
MGMT-7722 - V2 main page

jgyselov:
https://issues.redhat.com/browse/MGMT-7722

- [x] GET clusters
- [x] DELETE cluster
- [ ] GET component_versions


src/ocm/components/clusters/Clusters.tsx
```diff
@@ -44,7 +47,20 @@ const Clusters: React.FC<ClustersProps> = ({ history }) => {
   const deleteClusterAsync = React.useCallback(
     async (clusterId) => {
       try {
+        const infraEnvId = LocalStorageBackedCache.getItem(clusterId);
```
jkilzi, line 50: This is not enough. If the infraEnvId is not in the cache then we'll need to fetch it form the API.
There is this scenario where the cluster might have been created in one browser/computer but when the users want to delete it they use a different browser. In this case there's going to be a cache miss for sure.

jkilzi (changes requested):

src/ocm/api/addHostsClusters.ts
```diff
@@ -10,3 +10,9 @@ export const installHosts = (clusterId: Cluster['id']): AxiosPromise<Cluster> =>
 
 export const installHost = (clusterId: Cluster['id'], hostId: Host['id']): AxiosPromise<Host> =>
   client.post(`/v1/clusters/${clusterId}/hosts/${hostId}/actions/install`);
+
+export const getHosts = (infraEnvId: InfraEnv['id']): AxiosPromise<HostList> =>
+  client.get(`/v2/infra-envs/${infraEnvId}/hosts`);
+
+export const deleteHost = (infraEnvId: InfraEnv['id'], hostId: Host['id']): AxiosPromise<void> =>
+  client.delete(`/v2/infra-envs/${infraEnvId}/hosts/${hostId}`);
```
jkilzi, line 18: I'd rather put all these functions in a module called `HostsService`

jkilzi (changes requested):

jkilzi:
I'm closing this PR since we currently have most of this stuff already implemented in the `v2-api-migration` branch

PR #858, opened 2021-10-01
Rename Master to Control plane in HW requirements

jtomasek:


jkilzi (dismissed):

jkilzi:
@jtomasek why do we need this **now**? Is it a must-have?
I'm asking because right now in OCM we still show "master" host roles in the host tables. Merging this PR may trigger more work in order to keep the UI consistent everywhere with the same terminology.
I know we want to rename the roles from master to control-plane anyway, but there is no specific task that scopes only the renaming part.
@nirfarkas FYI...

jtomasek:
> @jtomasek why do we need this **now**? Is it a must-have? I'm asking because right now in OCM we still show "master" host roles in the host tables. Merging this PR may trigger more work in order to keep the UI consistent everywhere with the same terminology. I know we want to rename the roles from master to control-plane anyway, but there is no specific task that scopes only the renaming part. @nirfarkas FYI...

This fixes an issue raised on the ACM side https://github.com/open-cluster-management/backlog/issues/16737
I agree that there are places where we still use 'master' wording but IMHO it is fine to start changing it in places where it does not introduce collisions. The HW requirements is one of them. cc @jelkosz

PR #893, opened 2021-10-12
Improve agent progress calculation

jtomasek:
- Highlight bootstrap node when installing cluster
- Improve agent progress calculation by using correct progress stages


src/cim/components/helpers/toAssisted.ts
```diff
@@ -48,23 +43,14 @@ export const getAIHosts = (
         createdAt: agent.metadata?.creationTimestamp,
         validationsInfo: JSON.stringify({ hardware: [] }),
         inventory: JSON.stringify(inventory),
-        progress: {
-          currentStage,
-          progressInfo,
-          stageStartedAt: stageStartTime,
-          stageUpdatedAt: stageUpdateTime,
+        progress: agentProgress && {
```
mareklibra, line 46: This can result in `false` boolean value. Is it allowed value for the `progress`? What about using `undefined` when `!agentProgress`?

jtomasek: hmm, host.progress  `agentProgress` is `HostProgress | undefined`. I don't see how the result could be `false`

PR #891, opened 2021-10-12
Integration tests infrastructure

jkilzi:
(Very large PR ahead...)

# Integration tests infrastructure
### Changelog:
- Uses `jest` + `msw` for the test environment and server mocking
- Updates `react-scripts` to 4.0.3
- Updates `eslint` + `prettier` and all of its friends to meet the demands of react-scripts (most of the tooling for testing and code quality is influenced by this dependency)
- Updates the `eslint presets` due to deprecated rules, now they resemble the setup provided by CRA.
- `Github actions`: now each script runs in each own step (demo below)
- Uses `microbundle` instead of `microbundle-crl` (bye-bye `window['Fragment'] = React.Fragment` workaround)

### Pipeline demo

https://user-images.githubusercontent.com/77008341/136815821-d887ede1-d1a8-4013-8158-a91afce55c9f.mp4

### Resources
- [Testing Library](https://testing-library.com/docs/)
- [Redux/Writing tests](https://redux.js.org/usage/writing-tests)
- [Mock Service Worker](https://mswjs.io/docs/)
- [Stop mocking fetch](https://kentcdodds.com/blog/stop-mocking-fetch)
- [Epic React App](https://github.com/kentcdodds/bookshelf)

(... but it pays off because you can mock server responses now)

PR #910, opened 2021-10-19
Feature/integration tests

DanielleChason:
### **Files I changed:**

**IntegraionTestsUtils.tsx:**
I changes the wrapper to <BrowserRouter>, which is a router that keeps the UI in sync with the URL.

 **ClusterList.test.tsx:**
In this document there is one test suit with four test cases:
1. In this case there is no network failure and no clusters, and the application should render the EmptyState component. I added clicking on “creating new cluster” button and testing navigation to the right page. 
2. In this case there is a network failure and the application should render the ErrorState component.
3. In this case there is a cluster table with one cluster:

image

My test checks if the right fields of the table and the cluster appears with the correct data, and if clicking on the cluster’s name is taking you to the cluster’s page. 
4. In this case we’re testing if the refresh button is working as expected. 

**ClustersList.json**
This is a json file which includes all the data of the “localhost” cluster (case 3). 

**Changes in Clusters.tsx, ClusterLIstToolbar.tsx and ClustersTable.tsx:**
In those files I added “data-testid” attributes, in order to make the testing process more clean and accurate. 

Danielle Chason


jkilzi:
Closing in favor of #891

PR #936, opened 2021-10-31
Adds useStateSafely hook

jkilzi:
This new hook is a drop-in replacement for `React.useState`. 

Should be used when:
- The internal state of a component is altered by delegating the operation to a nested component.
- The internal state changes inside an async operation.

In general, this hook prevents changing the state of an unmounted component.

### DIFF
--- a/src/cim/components/helpers/versions.ts
+++ b/src/cim/components/helpers/versions.ts
@@ -12,7 +12,7 @@ const getVersion = (releaseImage = '') => {
 // eslint-disable-next-line
 const getSupportLevelFromChannel = (channel?: string): OpenshiftVersion['supportLevel'] => {
   if (!channel) {
-    return 'custom';
+    return 'maintenance';
   }
 
   if (channel.startsWith('fast')) {


src/common/hooks/useStateSafely.ts
```diff
@@ -0,0 +1,21 @@
+import type { Dispatch, SetStateAction } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
+
+export default function useStateSafely<S>(initialState: S): [S, Dispatch<SetStateAction<S>>] {
```
rawagner, line 4: We usually use arrow functions, why mixing the style with `function` keyword ?

src/common/components/hosts/DiskRole.tsx
```diff
@@ -81,9 +82,9 @@ const DiskRoleDropdown: React.FC<DiskRoleDropdownProps> = ({
         // TODO(mlibra): Improve for the case onDiskRole === undefined
         setOpen(false);
       };
-      asyncFunc();
+      void asyncFunc();
```
rawagner, line 85: what is the reason for adding `void` here ?

src/common/hooks/useStateSafely.ts
```diff
@@ -0,0 +1,21 @@
+import type { Dispatch, SetStateAction } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
```
rawagner, line 2: why do we have separate imports for functions and types ?

src/common/components/hosts/DiskRole.tsx
jkilzi: > This operator allows evaluating expressions that produce a value into places where an expression that evaluates to undefined is desired.
Since we ignore the returned promise, IMHO it makes sense to use it. The code gets executed as usual, the promise still fulfills or rejects anyway.

src/common/hooks/useStateSafely.ts
jkilzi: A: in order to make it clear what types the module is consuming and which functions it uses from the React package.

src/common/hooks/useStateSafely.ts
jkilzi: A: This module exports a single default function. Using a lambda here (as in many other places in our codebase) would demand declaring it in one statement and then adding an additional statement for the exported member.

src/common/components/hosts/DiskRole.tsx
mareklibra, line 85: why is this needed?

jkilzi: https://github.com/openshift-assisted/assisted-ui-lib/pull/936#discussion_r740035500

mareklibra: I missed the response above. IMHO, it brings noise in the code.
But I would be ok with merging it.

jkilzi: You can think of it as an explicit way of saying: "just call this (async) function, I (explicitly) don't care about the outcome"

src/common/hooks/useStateSafely.ts
mareklibra: IMHO, it would make sense when applied consistently throughout the code, which is not recent state.

jkilzi: IMHO, it's a good chance to start adopting it.

src/common/components/hosts/DiskRole.tsx
rawagner: so without `void` it means something different than `"just call this (async) function, I (explicitly) don't care about the outcome"` ?

src/common/hooks/useStateSafely.ts
rawagner: I disagree with adopting this, unless we have automatic prettier/eslint rule for it

src/common/components/hosts/DiskRole.tsx
jkilzi: If you and I have the same dev environment setup, you should see an annoying linting message saying that you're ignoring the value returned by the promise and then the alternative will be to silent the linter for this line or just use the void operator as I did.

src/common/hooks/useStateSafely.ts
rawagner: Ok, Im fine with it

src/common/components/hosts/DiskRole.tsx
rawagner: linter setup is defined by `.eslintrc` in root folder, isnt it ? Why do you use different setup ? The linter validation should be the same as CLI config otherwise we start adding random stuff (like this `void`) and later somebody can remove it because our CLI linter config changes and it starts to complain etc. Endless loop.

jkilzi: I totally agree, with you. My setup consumes the eslint configuration from `.eslintrc` located at the root folder. I'm wondering whether you see this message about the returned promise being ignored, as I do, on this line or not?

rawagner: I dont.
Even running `yarn test:lint` does not show any warnings

jkilzi: image

rawagner: Is this really an eslint rule ? I could not find anything like that.
Isnt it just some smartness build in IntelliJ ?

src/common/hooks/useStateSafely.ts
jkilzi: Can anybody mention why this is a bad practice? So far this discussion is only a matter of "taste".
- Line 1 gets completely erased after the code is transpiled.
- It is not harming the project in any form.
- It helps readers differentiate between types and functions exported by the package.

*Honestly, this is not mandatory. I don't expect you to start following this convention if you don't like it.*

Anyway, [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) should help (despite I don't think it should be using it).

src/common/components/hosts/DiskRole.tsx
jkilzi: Yep seems it's coming from inteliJ. Is that a bad thing?

rawagner: I guess its fine. I was mostly worried that you have different eslint config. 
If you dont mind putting void on other places in the lib in the future (as people who do not use InteliJ wont do it), instead of turning that inspection info off, then Im good :)

jkilzi: > If you don't mind putting void on other places in the lib in the future

I'll even do that with pleasure :)

mareklibra: I still think it's just about IDE setup. Such syntactic sugar just makes the code/language talkative, harder to read.

jkilzi: Well, this is JavaScript @mareklibra. May I ask what part exactly do you find hard to read, maybe I can explain it to you?

mareklibra: My point is not in that. I was just referring to style - to avoid whatever is not needed.

batzionb: [There's an interesting discussion in github about this ](https://stackoverflow.com/questions/41278900/intellij-idea-warning-promise-returned-is-ignored-with-aysnc-await)
I'm assuming Joni saw this...
The solution Joni mentioned is one of the answers.
I thought the solution of adding .catch might also be a good option, since error handling is always a good idea...

rawagner: `try catch` makes more sense to me than adding `void`.

jkilzi: @rawagner you cannot use a try-catch there, `asyncFunc` is not being called inside an async block.
What you can do is actually something like this: `asyncFunc().catch(/*...*/);`

PR #938, opened 2021-11-01
Correct endpoints for log downloading

jgyselov:


src/common/components/clusterDetail/KubeconfigDownload.tsx
```diff
@@ -27,16 +26,8 @@ const KubeconfigDownload: React.FC<KubeconfigDownloadProps> = ({
...
-          const { data } = await ClustersAPI.getPresignedForClusterCredentials({
-            clusterId,
-            fileName,
-          });
-          saveAs(data.url);
-        } else {
-          const response = await ClustersAPI.downloadClusterCredentials(clusterId, fileName);
-          saveAs(response.data, fileName);
-        }
+        const response = await ClustersAPI.downloadClusterCredentials(clusterId, fileName);
```
jkilzi, line 29: It is calling GET /v2/clusters/{cluster_id}/downloads/credentials

jkilzi: this one was using prevously GET /v1/clusters/{cluster_id}/downloads/files-presigned
should be migrated to GET /v2/clusters/{cluster_id}/downloads/files-presigned

PR #940, opened 2021-11-03
MGMT-7919 - Filter out OCP versions with maintenance support level in the create cluster screen

jgyselov:
https://issues.redhat.com/browse/MGMT-7919

### DIFF
--- a/src/common/api/types.ts
+++ b/src/common/api/types.ts
@@ -1723,7 +1723,7 @@ export interface OpenshiftVersion {
   /**
    * Level of support of the version.
    */
-  supportLevel: 'beta' | 'production' | 'custom';
+  supportLevel: 'beta' | 'production' | 'custom' | 'maintenance';
   /**
    * Indication that the version is the recommended one.
    */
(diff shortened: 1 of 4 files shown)


src/common/api/types.ts
```diff
@@ -1723,7 +1723,7 @@ export interface OpenshiftVersion {
   /**
    * Level of support of the version.
    */
-  supportLevel: 'beta' | 'production' | 'custom';
+  supportLevel: 'beta' | 'production' | 'custom' | 'maintenance';
```
jkilzi, line 1726: is this coming from swagger? I don't see an updated swagger.yaml file here?

jkilzi: this file should not be modified manually

jkilzi: seems like `custom` is not part of this interface anymore
image

jgyselov: No, it's added manually. Should I update the swagger.yaml instead? @jkilzi

jkilzi: yes please

jkilzi (approved):

PR #941, opened 2021-11-03
MGMT-8117: Integrate the AI flow into ACM Create Cluster Wizard

mareklibra:
The user can add hosts within the Cluster Create Wizard without prior
preparation of an infrastructure environment.


mareklibra:
Still WIP, it needs more work

src/cim/components/ClusterDeployment/ClusterDeploymentHostsDiscovery.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const isSNOCluster = getIsSNOCluster(agentClusterInstall);
+
+  return (
+    <Grid hasGutter>
+      <GridItem>
+        <TextContent>
+          <DiscoveryInstructions isSingleNodeCluster={isSNOCluster} />
+        </TextContent>
+      </GridItem>
+      {!!onCreateBMH && (
```
jtomasek, line 61: When should this button be available? It is currently missing when editing the cluster to add hosts. IIUC it should be available whenever this is AI flow, right?

src/cim/components/ClusterDeployment/ClusterDeploymentWizard.tsx
```diff
@@ -33,18 +56,48 @@ const ClusterDeploymentWizard: React.FC<ClusterDeploymentWizardProps> = ({
...
-            agentClusterInstall={agentClusterInstall}
-            onClose={onClose}
-            onSaveHostsSelection={onSaveHostsSelection}
-            agents={agents}
-            aiConfigMap={aiConfigMap}
-          />
-        ) : (
-          <LoadingState />
-        );
+        if (agentClusterInstall?.metadata?.name) {
```
jtomasek, line 59: Instead of defining conditional logic under hosts-selection step, could we define a new `host-discovery` step and use different steps for AI and CIM flow? IMHO it would greatly simplify the wizard here.

Ideally the wizard would just return components for steps and any specific logic should be handled inside the component (loading state which we have here originally as well)

src/cim/components/ClusterDeployment/ClusterDeploymentWizard.tsx
```diff
@@ -33,18 +56,48 @@ const ClusterDeploymentWizard: React.FC<ClusterDeploymentWizardProps> = ({
...
+                agents={agents}
+                aiConfigMap={aiConfigMap}
+              />
+            );
+          } else {
+            if (infraEnv) {
+              return (
+                <ClusterDeploymentHostsDiscoveryStep
+                  agentClusterInstall={agentClusterInstall}
+                  agents={aiFlowFilteredAgents}
```
jtomasek, line 76: Do we need to filter agents by the label? Can't we just filter for agents which are bound to the cluster?

mareklibra: It should be the same. But you are right, we can do the filtering one component level down.

src/cim/components/ClusterDeployment/ClusterDeploymentHostsDiscovery.tsx
mareklibra: The whole BMH flow needs to be reviewed. I.e. it seems that BMHs are skipped in the Agent's table in the Cluster wizard at all. Let's do that consistently in a follow-up.

src/cim/components/ClusterDeployment/ClusterDeploymentWizard.tsx
mareklibra: Sure. To do that, we will need to change logic around `next` in the `ClusterDeploymentDetailsStep` and `onBack` in the `ClusterDeploymentNetworkingStep`

src/cim/components/ClusterDeployment/ClusterDeploymentWizard.tsx
mareklibra: done

src/cim/components/ClusterDeployment/ClusterDeploymentWizard.tsx
mareklibra: done

PR #943, opened 2021-11-04
useStateSafely

jkilzi:
This new hook is a drop-in replacement for `React.useState`.

Should be used when:

- The internal state of a component is altered by delegating the operation to a nested component.
- The internal state changes inside an async operation.

In general, this hook prevents changing the state of an unmounted component.


batzionb (approved):

rawagner:
thanks @jkilzi !

PR #968, opened 2021-11-18
MGMT-7942: feature support level

batzionb:
https://issues.redhat.com/browse/MGMT-7942
1. show warnings for unsupported and dev/tech preview features in the "Review and create" step
2. show badges next to tech/preview features 
3. show inline comment for unsupported vms in cluster networking

### DIFF
--- a/src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
+++ b/src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
@@ -6,7 +6,6 @@ import { OPENSHIFT_LIFE_CYCLE_DATES_LINK } from '../../config';
 import { ClusterCreateParams } from '../../api';
 import { OpenshiftVersionOptionType } from '../../types';
 import { isSNOSupportedVersion, SelectField } from '../ui';
-import { FeatureSupportLevelContext } from '../../contexts';
 
 const OpenShiftLifeCycleDatesLink = () => (
   <a href={OPENSHIFT_LIFE_CYCLE_DATES_LINK} target="_blank" rel="noopener noreferrer">
@@ -38,7 +37,7 @@ const OpenShiftVersionSelect: React.FC<OpenShiftVersionSelectProps> = ({ version
     values: { highAvailabilityMode, openshiftVersion },
     setFieldValue,
   } = useFormikContext<ClusterCreateParams>();
-  const supportLevelData = React.useContext(FeatureSupportLevelContext);
+
   const selectedVersion = versions.find((v) => v.value === openshiftVersion) || versions[0];
   React.useEffect(() => {
     if (highAvailabilityMode === 'None' && !isSNOSupportedVersion(selectedVersion)) {
@@ -46,10 +45,7 @@ const OpenShiftVersionSelect: React.FC<OpenShiftVersionSelectProps> = ({ version
       setFieldValue('openshiftVersion', firstSupportedVersionValue);
     }
   }, [highAvailabilityMode, selectedVersion, setFieldValue, versions]);
-  React.useEffect(() => {
-    supportLevelData.openshiftVersion = openshiftVersion;
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [openshiftVersion]);
+
   const selectOptions = React.useMemo(
     () =>
       versions
--- a/src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
+++ b/src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
@@ -6,6 +6,7 @@ import { OPENSHIFT_LIFE_CYCLE_DATES_LINK } from '../../config';
 import { ClusterCreateParams } from '../../api';
 import { OpenshiftVersionOptionType } from '../../types';
 import { isSNOSupportedVersion, SelectField } from '../ui';
+import { FeatureSupportLevelContext } from '../../contexts';
 
 const OpenShiftLifeCycleDatesLink = () => (
   <a href={OPENSHIFT_LIFE_CYCLE_DATES_LINK} target="_blank" rel="noopener noreferrer">
@@ -37,7 +38,7 @@ const OpenShiftVersionSelect: React.FC<OpenShiftVersionSelectProps> = ({ version
     values: { highAvailabilityMode, openshiftVersion },
     setFieldValue,
   } = useFormikContext<ClusterCreateParams>();
-
+  const supportLevelData = React.useContext(FeatureSupportLevelContext);
   const selectedVersion = versions.find((v) => v.value === openshiftVersion) || versions[0];
   React.useEffect(() => {
     if (highAvailabilityMode === 'None' && !isSNOSupportedVersion(selectedVersion)) {
@@ -45,7 +46,10 @@ const OpenShiftVersionSelect: React.FC<OpenShiftVersionSelectProps> = ({ version
       setFieldValue('openshiftVersion', firstSupportedVersionValue);
     }
   }, [highAvailabilityMode, selectedVersion, setFieldValue, versions]);
-
+  React.useEffect(() => {
+    supportLevelData.openshiftVersion = openshiftVersion;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [openshiftVersion]);
   const selectOptions = React.useMemo(
     () =>
       versions
--- a/src/common/components/ui/FeatureSupportLevelBadge.tsx
+++ b/src/common/components/ui/FeatureSupportLevelBadge.tsx
@@ -13,6 +13,7 @@ import {
 
 export type SupportLevelBadgeProps = {
   featureId: FeatureId;
+  openshiftVersion?: string;
 } & DeveloperPreviewProps;
 
 const popoverTexts = {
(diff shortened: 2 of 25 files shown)


src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -37,15 +38,17 @@ const OpenShiftVersionSelect: React.FC<OpenShiftVersionSelectProps> = ({ version
...
   React.useEffect(() => {
     if (highAvailabilityMode === 'None' && !isSNOSupportedVersion(selectedVersion)) {
       const firstSupportedVersionValue = versions.find(isSNOSupportedVersion)?.value;
       setFieldValue('openshiftVersion', firstSupportedVersionValue);
     }
   }, [highAvailabilityMode, selectedVersion, setFieldValue, versions]);
-
+  React.useEffect(() => {
+    supportLevelService.setOpenshiftVersion(openshiftVersion);
+  }, [supportLevelService, openshiftVersion]);
```
jkilzi, line 51: The `supportLevelService` should refer be always the same instance after the context is initialized. AFAIU we are not creating a different instance during the app's lifecycle, it is a singleton and it remains such until the app ends, correct?

```suggestion
  }, [openshiftVersion]);
```

src/common/components/ui/FeatureSupportLevelBadge.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <p>
+        {getPopoverText(supportLevelTitle)}
+        <br />
+        <ExternalLink href={SUPPORT_LEVEL_LINKS[supportLevel]}>Learn more</ExternalLink>
+      </p>
+    </>
+  );
+  return (
+    <Popover bodyContent={bodyContent}>
+      <Component position={position} className={className} {...props}></Component>
```
jkilzi, line 53:
```suggestion
      <Component position={position} className={className} {...props} />
```

src/ocm/components/clusterConfiguration/ReviewFeatureSupportLevel.tsx
```diff
@@ -0,0 +1,104 @@
...
+  let supportLevel: PreviewSupportLevel;
+  //Show only one preview support level list, first priority to developer preview features
+  if (previewSupportLevels['dev-preview'].length) {
+    supportLevel = 'dev-preview';
+  } else if (previewSupportLevels['tech-preview'].length) {
+    supportLevel = 'tech-preview';
+  } else {
+    return null;
+  }
+  const featureList = previewSupportLevels[supportLevel].map((featureId: FeatureId) => (
```
jkilzi, line 59: `supportLevel` can be `null` --> `previewSupportLevels[supportLevel]` can be `undefined` --> `previewSupportLevels[supportLevel].map(...)` throws: `Uncaught TypeError: Cannot read properties of undefined (reading 'map')`
```suggestion
  const featureList = previewSupportLevels[supportLevel].map((featureId: FeatureId) => (
```

jkilzi (changes requested):

src/ocm/contexts/FeatureSupportLevelContextProvider.tsx
```diff
@@ -0,0 +1,137 @@
...
+import FeatureSupportLevelContext, {
+  SupportLevelDataInterface,
+  EmptySupportLevelData,
+} from '../../common/contexts/FeatureSupportLevelContext';
+import { ClusterFeatureUsage, FeatureId, SupportLevel, SupportLevelMap } from '../../common/types';
+import { getErrorMessage, handleApiError } from '../api/utils';
+import { FeatureSupportLevels, useAlerts } from '../../common';
+import { FeatureSupportLevelsAPI } from '../services/api';
+
+class FeatureSupportLevelData implements SupportLevelDataInterface {
```
jkilzi, line 12:
```suggestion
class FeatureSupportLevelData implements SupportLevelDataInterface {
```
It's long enough to be worth having its own file, wdyt?

src/ocm/components/clusterConfiguration/ReviewFeatureSupportLevel.tsx
batzionb: The support levels that are not 'tech-preview' and 'dev-preview' are filtered in line 45:
` for (const [featureId, supportLevel] of Object.entries(supportLevelMap)) {
           if (!isPreviewSupportLevel(supportLevel)) {
`

src/ocm/contexts/FeatureSupportLevelContextProvider.tsx
batzionb: Thanks for the reference, I'll try to use it in a separate PR

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -37,15 +38,18 @@ const OpenShiftVersionSelect: React.FC<OpenShiftVersionSelectProps> = ({ version
...
   const selectedVersion = versions.find((v) => v.value === openshiftVersion) || versions[0];
   React.useEffect(() => {
     if (highAvailabilityMode === 'None' && !isSNOSupportedVersion(selectedVersion)) {
       const firstSupportedVersionValue = versions.find(isSNOSupportedVersion)?.value;
       setFieldValue('openshiftVersion', firstSupportedVersionValue);
     }
   }, [highAvailabilityMode, selectedVersion, setFieldValue, versions]);
-
+  React.useEffect(() => {
+    supportLevelData.openshiftVersion = openshiftVersion;
```
rawagner, line 50: why are you changing the context value directly ? I think it needs to be done via callback & state to be reactive

src/ocm/contexts/FeatureSupportLevelContextProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+}>;
+
+export const FeatureSupportLevelProvider = ({
+  children,
+  loadingUi,
+  clusterFeatureUsage,
+  openshiftVersion,
+}: SupportLevelProviderProps) => {
+  const alerts = useAlerts();
+  const fetcher = () => FeatureSupportLevelsAPI.list().then((res) => res.data);
```
rawagner, line 24: this is a new instance on every rerender. Is it not going to cause problems in `useSWR` hook ? I think it would be safer to move it ouside the component

batzionb: It definitely isn't called on every rerender, useSWR makes sure of that. it does get called however on every onMount. I disabled it for rerunning every focus. I wanted to prevent that by upgrading useSWR to 1.0.1 to use useSWRImmutable that also takes it from the cache, but I saw it was complicated since OCM and CIM will need to be updated. There is this option however: 
1. to add another context in the router of the application that calls this API and stores just the value of the global vsupport level versions and this context will use it
2. not use useSWR at all... So it won't revalidate on mount

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
batzionb: This is a Work Around for the default openshift version not being available in NewClusterPage. If the openshift_versions were in a global context it wouldn't need to be updated in ClusterDetails at all.
The reason it is interactive, is that FeatureSupportLevelBadge rerender will have this updated value

src/ocm/contexts/FeatureSupportLevelData.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  public get isFullySupported() {
+    return this._isFullySupported;
+  }
+
+  public get clusterSupportLevelMap(): SupportLevelMap {
+    return this.clusterSupportLevels;
+  }
+
+  getVersionSupportLevel(featureId: FeatureId): SupportLevel | undefined {
```
jtomasek, line 45: wouldn't it be better to call this function `getFeatureSupportLevel`?

batzionb: good point. This is confusing. I'm differentiating between the feature support level in the context of a specific version, this is used for the badges, and the feature support level in the context of a specific cluster, this is used for review and create and for the inline comment of the cluster networking managed with vms. I'll think of a better name

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
jtomasek: Why do we need to store the currently selected version in the FeatureSupportLevelData? If the version is selected, it is stored in the cluster. I feel like we're adding a lot of logic to manage the state of currently selected version inside the class instead of passing it as a parameter to the getters. If we removed that, we wouldn't need to do these effects and maybe event the class would not be necessary and we'd have just a set of functions for retrieving the support information provided by the context provider:

`getFeatureSupportLevel(featureId, openshiftVersion)`, `isVersionFullySupported(openshiftVersion)`, ...

batzionb: The idea of this implementation is that the badge is plug and play without properties of the cluster, so you don't need to drill down the cluster feature usage and cluster openshift version to every component that needs to know the support level of a specific version.
You are right that the drawback is that the openshift version is maintained in two places. 
This can be fixed if the cluster was stored in a global place and the FeatureSupportLevelProvider can access it. (It actually is currently stored in a Redux state but my understanding is that we don't want to use this)
In general this implementation isn't ideal because the data required for this context isn't available globally, so I needed to work around it to achieve the easy usage of the badge.
I'm considering refactoring this in a separate PR

PR #974, opened 2021-11-22
MGMT-8221: End-of-support message for all versions

DanielleChason:
https://issues.redhat.com/browse/MGMT-8221

image

If a selected openshift version has less than 30 days until it'll be unsupported, a warning is presented to the user. 
Because there's no API currently all the versions + their last supported date is hard coded in a document called "openshiftVersionsData". 
All the data is taken from **Openshift platform life cycle** (https://access.redhat.com/support/policy/updates/openshift#dates).


src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -13,6 +13,9 @@ const OpenShiftLifeCycleDatesLink = () => (
   </a>
 );
 
+const lastSupportedVersions = ['4.7'];
+const lastSupportDate = 'August 24, 2022';
+
```
jkilzi, line 18: Please move this into `src/common/config/constants.ts`

jkilzi:
@jelkosz I think we should add some grace period, so this message starts appearing like X months/weeks/days before the end date occurs. wdyt?

jelkosz:
> @jelkosz I think we should add some grace period, so this message starts appearing like X months/weeks/days before the end date occurs. wdyt?

this message should start appearing only once there is a newer version available. So I would say the grace period should not really be "how long before the date occurred" but "is there a newer version we can recommend instead". WDYT?

jkilzi:
> > @jelkosz I think we should add some grace period, so this message starts appearing like X months/weeks/days before the end date occurs. wdyt?
> 
> this message should start appearing only once there is a newer version available. So I would say the grace period should not really be "how long before the date occurred" but "is there a newer version we can recommend instead". WDYT?

In that case, I think we should change the message because we are not recommending what version should be best to use (the higher or the lower in range?). And if this is the case, then we should pass information about the recommended version to this component (@DanielleChason 
 FYI)

DanielleChason:
> 

Is the recommended version the last one?

jkilzi:
> > 
> 
> Is the recommended version the last one?

Yes

DanielleChason:
I pushed the new changes to that PR.

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -4,13 +4,31 @@ import { global_warning_color_100 as warningColor } from '@patternfly/react-toke
...
 const OpenShiftLifeCycleDatesLink = () => (
   <a href={OPENSHIFT_LIFE_CYCLE_DATES_LINK} target="_blank" rel="noopener noreferrer">
     {'Learn more'} <ExternalLinkAltIcon />
   </a>
 );
 
+const diffInDaysBetweenDates = (versionDateString: string): number => {
+  const versionDate = new Date(versionDateString);
+  const today = Date.now();
+  const msInOneDay = 60 * 60 * 24 * 1000;
```
jkilzi, line 18: we have this constant already defined

jkilzi (approved):

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -4,13 +4,31 @@ import { global_warning_color_100 as warningColor } from '@patternfly/react-toke
...
+  const utcVersion = Date.UTC(
+    versionDate.getFullYear(),
+    versionDate.getMonth(),
+    versionDate.getDate(),
+  );
+  return Math.floor((utcVersion - today) / msInOneDay);
```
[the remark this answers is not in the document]
DanielleChason: So should I wait for that branch to be merged, and then use it in my code?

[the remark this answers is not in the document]
jkilzi: yes this version is not going to production until Feb-2 anyway

jkilzi (changes requested):
Please try to reuse the functionality in [this PR](https://github.com/openshift-assisted/assisted-ui-lib/pull/1081)

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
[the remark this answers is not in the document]
DanielleChason: OK

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -4,13 +4,20 @@ import { global_warning_color_100 as warningColor } from '@patternfly/react-toke
 import { OPENSHIFT_LIFE_CYCLE_DATES_LINK } from '../../config';
 import { OpenshiftVersionOptionType } from '../../types';
 import { SelectField } from '../ui';
+import openshiftVersionData from './openshiftVersionsData.json';
+import { DiffInDaysBetweenDates } from '../../sevices/DateAndTime';
```
jkilzi, line 8:
```suggestion
import { diffInDaysBetweenDates } from '../../sevices/DateAndTime';
```

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -4,13 +4,20 @@ import { global_warning_color_100 as warningColor } from '@patternfly/react-toke
...
 
 const OpenShiftLifeCycleDatesLink = () => (
   <a href={OPENSHIFT_LIFE_CYCLE_DATES_LINK} target="_blank" rel="noopener noreferrer">
     {'Learn more'} <ExternalLinkAltIcon />
   </a>
 );
 
+const isSelectedVersionInDataFile = (selectedVersionValue: string): boolean => {
+  if (selectedVersionValue in openshiftVersionData['versions']) return true;
+  return false;
```
jkilzi, line 18:
```suggestion
  return (selectedVersionValue in openshiftVersionData['versions']);
```

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -23,6 +30,18 @@ const getOpenshiftVersionHelperText = (versions: OpenshiftVersionOptionType[]) =
         &nbsp;Please note that this version is not production ready. <OpenShiftLifeCycleDatesLink />
       </>
     );
+  } else if (
+    isSelectedVersionInDataFile(selectedVersionValue) &&
+    DiffInDaysBetweenDates(openshiftVersionData['versions'][selectedVersionValue]) <= 30
```
jkilzi, line 35:
```suggestion
    diffInDaysBetweenDates(openshiftVersionData['versions'][selectedVersionValue]) <= 30
```

src/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx
```diff
@@ -23,6 +30,18 @@ const getOpenshiftVersionHelperText = (versions: OpenshiftVersionOptionType[]) =
         &nbsp;Please note that this version is not production ready. <OpenShiftLifeCycleDatesLink />
       </>
     );
+  } else if (
+    isSelectedVersionInDataFile(selectedVersionValue) &&
```
jkilzi, line 34: Why not just? Do we really need a function for doing this?
```suggestion
    selectedVersionValue in openshiftVersionData['versions'] &&
```

DanielleChason: You're right, we don't

jkilzi (approved):

PR #984, opened 2021-11-29
MGMT-7447: UI Step 2: Consume cluster network data from the new fields

jkilzi:
- The selector for the new fields fallback to the old fields for backwards compatibility
- The old fields should be removed once they are deprecated in the REST API  
  - [MGMT-7527](https://issues.redhat.com/browse/MGMT-7527) sets the context for this task

### DIFF
--- a/src/common/components/clusterConfiguration/networkConfigurationValidation.ts
+++ b/src/common/components/clusterConfiguration/networkConfigurationValidation.ts
@@ -9,20 +9,28 @@ import {
 } from '../ui';
 
 import { getSubnetFromMachineNetworkCidr } from './utils';
+import {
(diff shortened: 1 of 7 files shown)


batzionb (approved):
The name 'selectors' and using the term 'select' is confusing because it sounds like the state is consumed from a redux store

jkilzi:
> The name 'selectors' and using the term 'select' is confusing because it sounds like the state is consumed from a redux store

The term `selector` is not only bound to redux.
Please take a look at [this resource](https://redux.js.org/usage/deriving-data-selectors) to learn more.

batzionb (approved):

PR #985, opened 2021-11-29
BZ-2000142 - Sorting by Host in Assisted Cluster does not work as expected

jgyselov:
https://bugzilla.redhat.com/show_bug.cgi?id=2000142
The displayed host count is either from the `enabledHostCount` or `totalHostCount` field. Sorting value is now consistent with this.

Before:
image
After:
image

### DIFF
--- a/src/ocm/selectors/clusters.tsx
+++ b/src/ocm/selectors/clusters.tsx
@@ -8,7 +8,8 @@ import {
   Cluster,
   DASH,
   getDateTimeCell,
-  getReadyHostCount,
+  getEnabledHostCount,
+  getTotalHostCount,
   HostsCount,
   HumanizedSortable,
   ResourceUIState,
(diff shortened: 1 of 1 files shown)


src/ocm/selectors/clusters.tsx
```diff
@@ -62,7 +63,8 @@ const clusterToClusterTableRow = (cluster: Cluster): IRow => {
       {
         title: <HostsCount cluster={cluster} />,
         props: { 'data-testid': `cluster-hosts-count-${name}` },
-        sortableValue: getReadyHostCount(cluster),
+        sortableValue:
+          'enabledHostCount' in cluster ? getEnabledHostCount(cluster) : getTotalHostCount(cluster),
```
jkilzi, line 67: @jgyselov but it would still jump if we do this.
Assume first `enabledHostCount` is not in `cluster` then we sort by the `totalHostCount`.
Later, `enabledHostCount` becomes available in `cluster`, then we sort by a different number?
Do we have some guarantee the order would behave as expected?

jgyselov: @jkilzi [This](https://github.com/openshift-assisted/assisted-ui-lib/blob/9693a46d0cc97d37a3519bc217290e516e9010df/src/common/components/hosts/HostsCount.tsx#L39) is how we get the value that is displayed in the table. I didn't want to make changes to that, so I made the `sortableValue` parameter consistent with it.

jgyselov: If we wanted to make it all less obfuscated, we could always display only the `totalHostCount` and sort by it as well.

jkilzi (approved):

PR #988, opened 2021-11-30
MGMT-7891 - UI shows wrong machine cidr in available subnets

jgyselov:
https://issues.redhat.com/browse/MGMT-7891
Initial value for subnet select field is now `machineNetworkCidr` instead of `machineNetworks.cidr`.


src/ocm/selectors/clusterSelectors.ts
```diff
@@ -4,7 +4,7 @@ import { Cluster } from '../../common/api/types';
 export const selectMachineNetworkCIDR = ({
   machineNetworks,
   machineNetworkCidr,
-}: Partial<Cluster>) => _.head(machineNetworks)?.cidr ?? machineNetworkCidr;
+}: Partial<Cluster>) => machineNetworkCidr ?? _.head(machineNetworks)?.cidr;
```
jkilzi, line 7: I introduced this change, this is the way it should work. First attempt to grab the first machine network CIDR, but if it fails use the old `machineNetworkCidr` field
```suggestion
}: Partial<Cluster>) => _.head(machineNetworks)?.cidr ?? machineNetworkCidr;
```

jkilzi (changes requested):

src/ocm/selectors/clusterSelectors.ts
jkilzi: w/o the mock you shouldn't get the same result

PR #997, opened 2021-12-06
BZ-1986596 - SSH Key does not show up in active 'Generate Discovery ISO' modal after Generating Discovery ISO, followed by Edit ISO Configuration

jgyselov:
https://bugzilla.redhat.com/show_bug.cgi?id=1986596

Default values for the `DiscoveryImageConfigForm` are taken from the infraEnv. Loading or Error components will get displayed in corresponding states.

Also, let's not set the `pullSecret` to an empty string, it causes too many issues. :slightly_smiling_face:

https://user-images.githubusercontent.com/87187179/144851877-3540b0ba-1aec-4f7b-a4e2-1c06b9302ad8.mp4

### DIFF
--- a/src/ocm/services/DiscoveryImageFormService.ts
+++ b/src/ocm/services/DiscoveryImageFormService.ts
@@ -20,7 +20,7 @@ const DiscoveryImageFormService = {
       httpsProxy: formValues.httpsProxy,
       noProxy: formValues.noProxy,
       // TODO(mlibra): Does the user need to change pull-secret?
-      pullSecret: ocmPullSecret,
+      pullSecret: ocmPullSecret || undefined,
       sshPublicKey: formValues.sshPublicKey,
     };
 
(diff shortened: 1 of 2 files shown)


src/ocm/services/DiscoveryImageFormService.ts
```diff
@@ -20,7 +20,7 @@ const DiscoveryImageFormService = {
       httpsProxy: formValues.httpsProxy,
       noProxy: formValues.noProxy,
       // TODO(mlibra): Does the user need to change pull-secret?
-      pullSecret: ocmPullSecret,
+      pullSecret: ocmPullSecret === '' ? undefined : ocmPullSecret,
```
mareklibra, line 23:
```suggestion
      pullSecret: ocmPullSecret || undefined,
```

Would it work here?

jgyselov: It does. Thanks!

mareklibra (approved):

PR #1002, opened 2021-12-07
Add host auto-approve for Edit AI flow wizard

mareklibra:


src/cim/components/ClusterDeployment/helpers.ts
```diff
@@ -92,6 +92,9 @@ export const isCIMFlow = (clusterDeployment?: ClusterDeploymentK8sResource) =>
     INFRAENV_GENERATED_AI_FLOW
   ];
 
+export const isAgentOfInfraEnv = (infraEnv: InfraEnvK8sResource, agent: AgentK8sResource) =>
+  agent.metadata?.labels?.[INFRAENV_AGENTINSTALL_LABEL_KEY] === infraEnv.metadata?.name;
```
rawagner, line 96: we should also check that the namespace is the same, to be sure

mareklibra: The agent is created in the namespace of the InfraEnv and just the infraEnv's name is stored among labels. 
I can add it for future reusability of the utility function.

mareklibra: done

rawagner (approved):

PR #1004, opened 2021-12-08
Use cluster validationsInfo in CIM

jtomasek:
Refactoring of cluster validations section components and wizardTransitions such that it can be used by both CIM and OCM.  Adds cluster validations to wizard steps.

Screenshot 2021-12-08 at 10 50 37


mareklibra (approved):

nirfarkas:
Are we keeping the "Install" button enabled when validations fail?

PR #1006, opened 2021-12-08
Use tooltip instead of popover "?" for BMC on BM info

mareklibra:


src/cim/components/modals/AddHostModal.tsx
```diff
@@ -1,28 +1,23 @@
...
+  Modal,
+  ModalBoxBody,
+  ModalVariant,
+  Radio,
+  Tooltip,
+} from '@patternfly/react-core';
 import {
   DownloadIso,
-  PopoverIcon,
+  // PopoverIcon,
```
jtomasek, line 14: Can be removed

jtomasek (approved):

jtomasek (approved):

PR #1015, opened 2021-12-13
MGMT-8502: add test IDs to the support level badges

batzionb:
added test ids to:
support level badges
review and create support level section
cluster managed networking with vms alert

### DIFF
--- a/src/common/components/ui/PreviewBadge.tsx
+++ b/src/common/components/ui/PreviewBadge.tsx
@@ -3,6 +3,14 @@ import { Label, Popover } from '@patternfly/react-core';
 import { InfoCircleIcon } from '@patternfly/react-icons';
 import { TECH_SUPPORT_LEVEL_LINK } from '../../config/constants';
 import ExternalLink from './ExternalLink';
+import { WithTestID } from '../../types';
+
+export type DeveloperPreviewProps = {
+  position?: PreviewBadgePosition;
+  className?: string;
+} & WithTestID;
+
+export type TechnologyPreviewProps = DeveloperPreviewProps;
 
 export enum PreviewBadgePosition {
   default,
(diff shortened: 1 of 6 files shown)


src/common/components/ui/PreviewBadge.tsx
```diff
@@ -53,14 +55,15 @@ const PreviewBadge: React.FC<PreviewBadgeProps> = ({
...
+        data-testid={`${idPrefix}-support-level`}
       >
         {text}
       </Label>
     </Popover>
   );
 };
 
-export type DeveloperPreviewProps = Pick<PreviewBadgeProps, 'className' | 'position'>;
+export type DeveloperPreviewProps = Pick<PreviewBadgeProps, 'className' | 'position' | 'idPrefix'>;
```
jkilzi, line 66: We already have a type called `WithTestID`, which you can use in order to extend the props of some component with a `testId` member ([example](https://github.com/openshift-assisted/assisted-ui-lib/blob/07da563e63701afbeae15ac430d1ca33a85c86b2/src/common/components/ui/DetailList.tsx#L47))

jkilzi (changes requested):

src/common/components/ui/PreviewBadge.tsx
batzionb: This doesn't get a test id it gets an id prefix the testId itself is generated in the PreviewBadge component

jkilzi: I still don't understand why the badges need this prop? is it for automation? if it is then you should use a different name for that prop (maybe `testId`) since `idPrefix` is used in other places in order to put id attributes on the underlying components ([example](https://github.com/openshift-assisted/assisted-ui-lib/blob/5e6951917760e4f5c40049806302c1b45110943b/src/ocm/components/AddHosts/InventoryAddHost.tsx#L30)).

jkilzi (changes requested):

jkilzi (approved):

PR #1018, opened 2021-12-14
Add static ip option for non-bm platforms

rawagner:
Screenshot from 2021-12-14 10-09-24

Screenshot from 2021-12-14 10-09-07

### DIFF
--- a/src/cim/components/InfraEnv/InfraEnvFormPage.tsx
+++ b/src/cim/components/InfraEnv/InfraEnvFormPage.tsx
@@ -18,6 +18,7 @@ import {
   TitleSizes,
 } from '@patternfly/react-core';
 import { Formik, FormikProps, useFormikContext } from 'formik';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 import {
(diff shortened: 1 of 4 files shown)


jtomasek (approved):

nirfarkas:
I think we need to change the text.
In the popover:
To use static network configuration, follow the steps listed in the documentation.
<link>
in the inline alert to:
To use static network configuration, follow the steps listed in the documentation.
<link>

The text in the popover isn't describing accurately what the user needs to do (create a CR modify another and provide NMstate?)
The text in the inline alert should be identical.

jelkosz:
> I think we need to change the text.
> In the popover:
> To use static network configuration, follow the steps listed in the documentation.
> in the inline alert to: To use static network configuration, follow the steps listed in the documentation. 

sounds good

> The text in the popover isn't describing accurately what the user needs to do (create a CR modify another and provide NMstate?)

I would say that that is a bit too much info given that the documentation describes it

nirfarkas:
So the text for both (popover and inline notification is:
To use static network configuration, follow the steps listed in the documentation.

PR #1026, opened 2021-12-15
Show static ip info on ISO download page too

rawagner:
Fix iso-download/iso-config logic
Update static network help text

### DIFF
--- a/src/common/components/clusterConfiguration/DownloadIso.tsx
+++ b/src/common/components/clusterConfiguration/DownloadIso.tsx
@@ -12,13 +12,17 @@ import {
   Title,
   ModalBoxBody,
   ModalBoxFooter,
+  Stack,
+  StackItem,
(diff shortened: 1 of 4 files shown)


src/common/components/clusterConfiguration/DownloadIso.tsx
```diff
@@ -30,45 +34,61 @@ const DownloadIso: React.FC<DownloadISOProps> = ({
...
+                    {wgetCommand}
+                  </ClipboardCopy>
+                }
+              />
+            </DetailList>
+            <Alert
+              variant="info"
+              isInline
+              title={
+                'Never share your downloaded ISO with anyone else. ' +
```
mareklibra, line 86: Nit: can we use just a single string?

mareklibra (approved):

PR #1028, opened 2021-12-16
MGMT8041: Update minimum hw requirements alerts/popovers

rawagner:
Screenshot from 2021-12-16 13-19-10

Screenshot from 2021-12-16 13-15-39


rawagner:
@ShiranHi FYI in the JIRA you mentioned to keep the alert expanded if there are no hosts but PF does not have that option

ShiranHi:
@rawagner regarding the expanded inf-bar: can we have it open by default, but keep it closed once users collapsed it? do we have any other solution we can have here?

nirfarkas:
@ShiranHi why not use here the same pattern we're using in AI?
We have a link button that takes the minimum vertical space and clicking on it is showing the info in a modal.

ShiranHi:
@nirfarkas we created an info bar because we wanted to make it more prominent than just one small link on top of the table view. It is more accessible this way to read while users see the table right below it.

mareklibra (dismissed):

rawagner:
@mareklibra Lets put this on hold, seems like @nirfarkas and @ShiranHi did not yet agree if the design is ok.

ShiranHi:
@rawagner, @nirfarkas and I just talked about this and agreed that in this case, it is ok to have it as an info-bar in CIM and as a link in AI, so let's keep it as implemented.

src/cim/components/InfraEnv/AgentAlerts.tsx
```diff
@@ -18,22 +26,38 @@ const AgentAlerts: React.FC<AgentAlertsProps> = ({ infraEnv, bareMetalHosts, doc
...
+      )}
+      {aiConfigMap && (
+        <Alert
+          title="Minimum hardware requirements"
+          variant={AlertVariant.info}
+          isInline
+          isExpandable
+          className="cim-resource-alerts"
+        >
+          <HostRequirements {...getHWRequirements(aiConfigMap)} />
```
jtomasek, line 56: The list in the alert is left-padded too much as seen on the screenshot. That is caused by presence of `<TextContent>` inside `HostRequirements`'s render function.

Since `HostRequirements` component is currently used only on one place, I'd suggest renaming it to `HostRequirementsList` and take the `TextContent` out and use it like this in `MinimalHWRequirements` component:
```
<TextContent>
  <HostRequirementsList {...getHWRequirements(aiConfigMap)} isSNOCluster={isSNOCluster} />
</TextContent>
```

rawagner: good catch. Fixed
Screenshot from 2022-02-08 15-32-21

jtomasek (approved):

PR #1035, opened 2021-12-21
MGMT-8500 - OCP versions are sorted old -> new rather than the other way around

jgyselov:
https://issues.redhat.com/browse/MGMT-8500
Before:
image

After:
image


nirfarkas:
Super!

batzionb:
This might not work when there'll be version 10, since it's not alphanumerically sorted

batzionb (approved):

PR #1070, opened 2022-01-11
Fix canUnbindHost for BMH

mareklibra:


src/cim/components/Agent/tableUtils.tsx
```diff
@@ -302,7 +302,7 @@ export const useAgentsTable = (
         canUnbindHost: canUnbindHost
           ? (host: Host) => {
               const agent = agents.find((a) => a.metadata?.uid === host.id);
-              return agent ? canUnbindHost(agent) : ['false'];
+              return agent ? canUnbindHost(agent) : [false, '' /* Only agent can be unbound */];
```
jtomasek, line 305: This should not be needed since agent is optional in `canUnbindHost` and when there is no agent, it is handled there already: https://github.com/stolostron/console/pull/1095/files#diff-e336279de3b51f940b63be20165303c34785a1a90746a6bcdfea7746feaa4a19R94

jtomasek, line 305:
```suggestion
              return canUnbindHost(agent);
```

src/common/components/hosts/tableUtils.tsx
```diff
@@ -486,20 +486,23 @@ export const hostActionResolver = ({
...
-            description={canUnbindHostResult?.[1]}
-            title="Remove from the cluster"
-          />
-        ),
-        id: `button-unbind-host-${hostname}`,
-        onClick: () => !isDisabled && onUnbindHost && onUnbindHost(host),
-        disabled: isDisabled,
-      });
+      const isEnabled = canUnbindHostResult?.[0];
+      const disabledReason = canUnbindHostResult?.[1];
```
jtomasek, line 490: `canUnbindHost` should always return the `[boolean, string]` tuple, right?
```
const [isEnabled, disabledReason] = canUnbindHostResult(host);
```

jtomasek (changes requested):

src/cim/components/Agent/tableUtils.tsx
mareklibra: You are right in the recent flow. It's just keeping it safe for the future. Anyway, we need to handle `agent === undefined` somehow and this is one approach.

mareklibra: You are right for that flow. We will need to handle `agent === undefined` anyway, so we move that logic into canUnbindHost

src/common/components/hosts/tableUtils.tsx
mareklibra: right, changed

src/cim/components/Agent/tableUtils.tsx
mareklibra: Already done.

jtomasek (approved):

PR #1082, opened 2022-01-18
MGMT-8068: Add edit capabilities for infra env pullsecret and ssh keys

rawagner:
Screenshot from 2022-01-19 09-54-58

Screenshot from 2022-01-19 09-55-11

Screenshot from 2022-01-19 09-55-22

@ShiranHi 

According to https://marvelapp.com/prototype/haechg7/screen/84021478 we should show a warning if agent is booted with old ISO. But that cannot be detected from UI.


ShiranHi:
@rawagner looks good to me 👍

mareklibra (approved):

PR #1087, opened 2022-01-24
Add NTP sources configuration for CIM

rawagner:
I've made some changes so @ShiranHi let me know if you agree or not

1. sources are separated via comma, not via new line. This is how we do it on OCM so I think it would be good to align
2. I've added a help text under TextArea to make clear what format we expect (we do the same on OCM)
3. TextArea is not aligned with checkbox content, we do not do it on other places

ntp1

ntp4

net5


ShiranHi:
@rawagner thank you, happy to see it! Looks good to me 👍

jtomasek (approved):

PR #1088, opened 2022-01-25
NO-ISSUE: resolved dependency on ocm in common

batzionb:


src/common/components/featureSupportLevels/utils.ts
```diff
@@ -47,11 +43,7 @@ export const getLimitedFeatureSupportLevels = (
     }
     return ret;
   } catch (err) {
-    captureException(
-      err,
-      `Failed to get cluster ${cluster.id} feature support levels`,
-      Sentry.Severity.Warning,
-    );
+    console.error(`Failed to get cluster ${cluster.id} feature support levels`);
```
jkilzi, line 46: Idea: we could pass the `captureException` as an additional arg to this function. Later any consumer can decide "how to capture the exception". In OCM we report it to Sentry in a disconnected environment they might decide to do something else that doesn't require contacting an external service. WDYT?

jkilzi (commented):
I left some ideas, let me know what you think...

src/common/components/featureSupportLevels/utils.ts
batzionb: this function is called from a component in common, that's called from another component in common, etc. I think the only way to solve this is with the errorhandler context

jkilzi: That's not correct
image

batzionb:
opened a [new PR](https://github.com/openshift-assisted/assisted-ui-lib/pull/1098) instead of this one

PR #1091, opened 2022-01-27
Introduce ACMFeatureSupportLevelProvider

mareklibra:
So far with hardcoded values.


src/common/components/clusterConfiguration/SNOControlGroup.tsx
```diff
@@ -4,25 +4,29 @@ import { ClusterDetailsValues } from '../../../common/components/clusterWizard/t
 import { SingleNodeCheckbox } from '../ui';
 import { OpenshiftVersionOptionType } from '../../types';
 import SNODisclaimer from './SNODisclaimer';
-import { FeatureSupportLevelContext } from '../featureSupportLevels';
+import { useFeatureSupportLevel } from '../featureSupportLevels';
 
 type SNOControlGroupProps = {
   versions: OpenshiftVersionOptionType[];
   highAvailabilityMode: ClusterDetailsValues['highAvailabilityMode'];
+  isDisabled?: boolean;
```
batzionb, line 12: The idea was to use a global context to determine if a feature is diabled or not instead of props drilling. Is there no way to implement the same logic that determines the isDisabled property value in the FeatureSupportLevelContext provider?

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const supportLevelData: FeatureSupportLevelsMap = React.useMemo<FeatureSupportLevelsMap>(() => {
+    const featureSupportLevelsMap: FeatureSupportLevelsMap = {};
+
+    ocpVersions.forEach((version) => {
+      // TODO(mlibra): get this dynamically
+      const featureIdToSupportLevel: FeatureIdToSupportLevel = {
+        REQUESTED_HOSTNAME: 'supported',
+        SNO: 'supported',
+        CNV: 'supported',
+        NETWORK_TYPE_SELECTION: 'supported',
```
batzionb, line 33: NETWORK_TYPE_SELECTION is supported always in ACM? what about in SNO?

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
```diff
@@ -0,0 +1,99 @@
...
+    [getVersionSupportLevelsMap],
+  );
+
+  const isFeatureSupportedCallback: FeatureSupportLevelData['isFeatureSupported'] = React.useCallback(
+    () => true,
+    [],
+  );
+
+  const getDisabledReasonCallback: FeatureSupportLevelData['getFeatureDisabledReason'] = React.useCallback(() => {
+    return undefined; // we have recently everything enabled in the ACM
```
batzionb, line 68: SNO is always enabled in ACM? What about with OCP 4.8?

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
```diff
@@ -0,0 +1,99 @@
...
+        SNO: 'supported',
+        CNV: 'supported',
+        NETWORK_TYPE_SELECTION: 'supported',
+      };
+      featureSupportLevelsMap[version.version] = featureIdToSupportLevel;
+    });
+    return featureSupportLevelsMap;
+  }, [ocpVersions]);
+
+  const getVersionSupportLevelsMap: FeatureSupportLevelData['getVersionSupportLevelsMap'] = React.useCallback(
```
batzionb, line 40: you can simply return undefined here

src/common/components/clusterConfiguration/SNOControlGroup.tsx
mareklibra: This time it's not about the feature but the flow. We have it enabled in the Create Cluster but not Edit flow. We can leverage FeatureSupportLevel for that but it somehow does not fit in my thinking.

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
mareklibra: Right. It's weird since it is a TODO. So far there is everything on the "production" support level.
@jelkosz Do we have an ACM 2.5 task for that? Or plan to change it?

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
mareklibra: Right. I could simplify whole component but I consider it as a skeleton, expected to be changed once we properly retrieve the support map.

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
jelkosz: it should not be the case - Im sure I have seen the banner for sno for older version in acm. But checking now it does not work.
@jtomasek any idea what happened to it?

batzionb: It's since I removed the hard coded behavior of this banner, it needs to be implemented in the featuresupportlevel provider

src/common/components/clusterConfiguration/SNOControlGroup.tsx
batzionb: I know it's not ideal, it should be in a separate context, but for now it does eliminate the props drilling and is convenient to use

batzionb: We want to implement a separate context FeatureStateContext - does that fit more in your opinion? Question is if for now it can be in FeatureSupportLevel context

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
batzionb: makes sense

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
mareklibra: We will need to carefully review all features. I.e. I can see disk encryption in the ACM UI, which is not persisted. Let's do that in a separate task.

src/cim/components/featureSupportLevels/ACMFeatureSupportLevelProvider.tsx
batzionb: You might need to add to this the logic of SNO being disabled in version < 4.9, but as you said it can be done in a separate task

batzionb:
LGTM

rawagner (approved):

src/common/components/clusterConfiguration/SNOControlGroup.tsx
jtomasek: I agree that disabling SNO due to editting the cluster is not really related to the SNO feature support so it should not be driven by the featuresupport context/hook.

jtomasek: So the SNO can be disabled for 2 separate reasons (editting the cluster and feature not being supported in selected version)

batzionb: yes, that's why in OCM we return in getDisabledReason a different message for each case and it shows in a popover. Regarding the disabling in FeatureSupportLevel we want to refactor this

PR #1092, opened 2022-01-27
MGMT-9061 - Download kubeconfig button doesn't exist

jgyselov:
https://issues.redhat.com/browse/MGMT-9061

The constant used to indicate installation hasn't started/finished yet is different than the `TIME_ZERO` constant. This was causing the calculated number of days until the kubeconfig file is deleted to be negative which prevented the button from rendering.


jkilzi (approved):

jkilzi:
https://github.com/openshift/assisted-service/pull/3261 fixes the timestamp to its previous value.
We'll need to revert this PR in order to make it work with the latest BE change

PR #1097, opened 2022-01-28
MGMT-9067: Implement the UI for auto-select control plane nodes

jgyselov:
https://issues.redhat.com/browse/MGMT-9067

- [x] No toggle switch will be displayed
- [x] When 1-3 hosts are discovered we should show them with the suggested role ("Control plane, worker") that is not editable
- [x] Hovering the host role should show a tooltip: "To make the role editable, 4 or more hosts need to be discovered."
- [x] When the 4th  host is discovered, we should enable the edit option for the hosts' role and add the dropdown caret
- [x] We should not limit the user with the role selections (the user can select any combination of roles and they will be validated

### DIFF
--- a/src/ocm/components/hosts/HostsDiscoveryTable.tsx
+++ b/src/ocm/components/hosts/HostsDiscoveryTable.tsx
@@ -6,6 +6,7 @@ import {
   getSchedulableMasters,
   Cluster,
   Host,
+  isSingleNodeCluster,
 } from '../../../common';
 import { HostsTableModals, useHostsTable } from './use-hosts-table';
 import {
@@ -24,6 +25,9 @@ import { ExpandComponentProps } from '../../../common/components/hosts/AITable';
 import { AdditionalNTPSourcesDialogToggle } from './AdditionaNTPSourceDialogToggle';
 import { onDiskRoleType } from '../../../common/components/hosts/DiskRole';
 import { Stack, StackItem } from '@patternfly/react-core';
+import { HostsService } from '../../services';
+import { updateHost } from '../../reducers/clusters';
+import { useDispatch } from 'react-redux';
 
 const getExpandComponent = (onDiskRole: onDiskRoleType, canEditDisks: (host: Host) => boolean) => ({
   obj: host,
@@ -60,10 +64,17 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
     ...modalProps
   } = useHostsTable(cluster);
 
+  const dispatch = useDispatch();
+
   const content = React.useMemo(
     () => [
       hostnameColumn(onEditHost, undefined, actionChecks.canEditHostname),
-      roleColumn(actionChecks.canEditRole, onEditRole, getSchedulableMasters(cluster)),
+      roleColumn(
+        actionChecks.canEditRole,
+        onEditRole,
+        getSchedulableMasters(cluster),
+        !isSingleNodeCluster(cluster),
+      ),
       hardwareStatusColumn(onEditHost),
       discoveredAtColumn,
       cpuCoresColumn,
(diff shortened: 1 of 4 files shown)


src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -60,10 +64,17 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
...
+
   const content = React.useMemo(
     () => [
       hostnameColumn(onEditHost, undefined, actionChecks.canEditHostname),
-      roleColumn(actionChecks.canEditRole, onEditRole, getSchedulableMasters(cluster)),
+      roleColumn(
+        actionChecks.canEditRole,
+        onEditRole,
+        getSchedulableMasters(cluster),
+        !isSingleNodeCluster(cluster),
```
jkilzi, line 76: Please use `isSNO` from `src/ocm/clusterSelectors` instead

src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -74,6 +85,20 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
...
+  React.useEffect(() => {
+    const forceRole = async () => {
+      cluster.hosts?.forEach(async (host) => {
+        if (host.role !== 'master') {
+          const { data } = await HostsService.updateRole(cluster.id, host.id, 'master');
+          dispatch(updateHost(data));
+        }
+      });
+    };
+    if (cluster.hosts && cluster.hosts?.length <= 3) {
```
jkilzi, line 97: Isn't the question mark in `cluster.hosts?` redundant?

src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -74,6 +85,20 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
     [onEditHost, actionChecks.canEditHostname, actionChecks.canEditRole, onEditRole, cluster],
   );
 
+  React.useEffect(() => {
+    const forceRole = async () => {
+      cluster.hosts?.forEach(async (host) => {
+        if (host.role !== 'master') {
+          const { data } = await HostsService.updateRole(cluster.id, host.id, 'master');
```
jkilzi, line 92: What would happen if any of these requests would throw an error?

src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -74,6 +85,20 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
...
+          dispatch(updateHost(data));
+        }
```
jkilzi, line 94: Would be better written and allow for better error handling if it would be written with [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)

```suggestion
      for await (const host of cluster.hosts) {
        if (host.role !== 'master') {
          try {
            const { data } = await HostsService.updateRole(cluster.id, host.id, 'master');
            dispatch(updateHost(data));
          } catch (error) {
            // here comes the error handling part...
          }
        }
```

jkilzi (changes requested):

src/ocm/components/hosts/HostsDiscoveryTable.tsx
jkilzi: ?

src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -74,6 +89,35 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
     [onEditHost, actionChecks.canEditHostname, actionChecks.canEditRole, onEditRole, cluster],
   );
 
+  React.useEffect(() => {
+    const forceRole = async () => {
+      if (cluster.hosts && cluster.hosts.length <= 3) {
+        try {
+          const promises = [];
+          for await (const host of cluster.hosts) {
```
jkilzi, line 97:
```suggestion
          for (const host of cluster.hosts) {
```
my bad no need for `await` here

jkilzi (approved):

PR #1098, opened 2022-01-30
NO-ISSUE: resolve dependency of common on OCM featureSupportlevel utils

batzionb:
resolves issue of dependency on OCM sentry error handling using a callback

### DIFF
--- a/src/common/components/featureSupportLevels/utils.ts
+++ b/src/common/components/featureSupportLevels/utils.ts
@@ -1,15 +1,15 @@
-import { ClusterFeatureUsage, FeatureId, FeatureIdToSupportLevel } from '../../../common/types';
-import { Cluster } from '../../../common/api/types';
-import { FeatureSupportLevelData } from '../../../common/components/featureSupportLevels/FeatureSupportLevelContext';
-import { captureException } from '../../sentry';
-import * as Sentry from '@sentry/browser';
-import { stringToJSON } from '../../../common/api/utils';
-
 // TODO: move to common since networkConfiguration is dependant on it.
+
+import { Cluster, stringToJSON } from '../../api';
+import { ErrorHandler } from '../../types/errorHandling';
+import { ClusterFeatureUsage, FeatureId, FeatureIdToSupportLevel } from '../../types';
+import { FeatureSupportLevelData } from './FeatureSupportLevelContext';
+
 // will be able to implement it once there's a common error handler
 export const getLimitedFeatureSupportLevels = (
   cluster: Cluster,
   featureSupportLevelData: FeatureSupportLevelData,
+  errorHandler?: ErrorHandler,
 ): FeatureIdToSupportLevel | undefined => {
   try {
     if (!cluster.openshiftVersion) {
@@ -47,11 +47,9 @@ export const getLimitedFeatureSupportLevels = (
     }
     return ret;
   } catch (err) {
-    captureException(
-      err,
-      `Failed to get cluster ${cluster.id} feature support levels`,
-      Sentry.Severity.Warning,
-    );
+    if (errorHandler) {
+      errorHandler(err, `Failed to get cluster ${cluster.id} feature support levels`);
+    }
     return undefined;
   }
 };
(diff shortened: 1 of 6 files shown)


src/common/components/featureSupportLevels/utils.ts
```diff
@@ -47,11 +47,7 @@ export const getLimitedFeatureSupportLevels = (
     }
     return ret;
   } catch (err) {
-    captureException(
-      err,
-      `Failed to get cluster ${cluster.id} feature support levels`,
-      Sentry.Severity.Warning,
-    );
+    errorHandler(err, `Failed to get cluster ${cluster.id} feature support levels`);
```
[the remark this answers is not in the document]
jkilzi: IMHO logging things to the console and passing no `errorHandler` is pretty much the same. Both options are noops, maybe the latter would be useful in local development, but the first could impact the app's performance.
If console.logging is the actual consumer's intention, I'd let them do that explicitly rather than doing that for them implicitly.

src/ocm/sentry.ts
```diff
@@ -16,3 +17,13 @@ export const captureException = (
...
+
+export const sentryErrorHandler: ErrorHandler = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  error: any,
+  message?: string,
+  severity = 'error',
+) => {
+  const sentrySeverity = severity === 'error' ? Sentry.Severity.Error : Sentry.Severity.Warning;
+  return captureException(error, message, sentrySeverity);
+};
```
jkilzi, line 29: This is just duplicating what `captureException` already does. What is this good for?

src/ocm/components/featureSupportLevels/ReviewClusterFeatureSupportLevels.tsx
```diff
@@ -113,7 +114,7 @@ export const ClusterFeatureSupportLevelsDetailItem: React.FC<{ cluster: Cluster
   const featureSupportLevelData = useFeatureSupportLevel();
 
   const clusterFeatureSupportLevels = React.useMemo(() => {
-    return getLimitedFeatureSupportLevels(cluster, featureSupportLevelData);
+    return getLimitedFeatureSupportLevels(cluster, featureSupportLevelData, sentryErrorHandler);
```
jkilzi, line 117: Why not just call `captureException`?

```suggestion
    return getLimitedFeatureSupportLevels(cluster, featureSupportLevelData, captureException);
```

src/ocm/sentry.ts
jkilzi: Maybe `captureException` is not a good name and you'd like it to be called `sentryErrorHandling` instead? (If so, I'm with you on this)

jkilzi (changes requested):

src/ocm/sentry.ts
batzionb: it's because of the mismatch in types. I'll find a different way to resolve this

jkilzi: Sorry, I don't see what's the mismatch? Both functions seem to have the same signature

batzionb: the captureException uses the sentry error type, which can't be imported in common

src/common/components/featureSupportLevels/utils.ts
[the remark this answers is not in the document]
batzionb: no need to write IMHO in reviews of my code ;)

batzionb:
I thought about this in general, the sentry error handling is also not good, since for every polling of the cluster it'll report an error to santry, which anyhow gets too many errors, and it's complex to make this not happen. 
Maybe we should avoid the error handling all together. The chances featureUsage is wrong is low, and if it is there will be a bug and we'll fix the bug. 
I think better to remove the dependency and not handle this error

src/ocm/sentry.ts
jkilzi: Then why not refactor `captureException` so it can receive a generic type for the severity argument?

jkilzi:
> I thought about this in general, the sentry error handling is also not good, since for every polling of the cluster it'll report an error to santry, which anyhow gets too many errors, and it's complex to make this not happen. Maybe we should avoid the error handling all together. The chances featureUsage is wrong is low, and if it is there will be a bug and we'll fix the bug. I think better to remove the dependency and not handle this error

There is no problem with having the same issue reported multiple times. It's actually worse if we'll have no visibility over such issues.

jkilzi (approved):

PR #1114, opened 2022-02-01
MGMT-9137: Disk Encryption - radio buttons will appear below "enable on" switch buttons

DanielleChason:
https://issues.redhat.com/browse/MGMT-9137

image

### DIFF
--- a/src/common/components/clusterConfiguration/DiskEncryptionFields/DiskEncryptionMode.tsx
+++ b/src/common/components/clusterConfiguration/DiskEncryptionFields/DiskEncryptionMode.tsx
@@ -65,8 +65,6 @@ export interface DiskEncryptionModeProps extends DiskEncryptionValues {
 }
 
 export const DiskEncryptionMode: React.FC<DiskEncryptionModeProps> = ({
-  enableDiskEncryptionOnMasters,
-  enableDiskEncryptionOnWorkers,
   diskEncryptionMode,
   isDisabled,
 }) => {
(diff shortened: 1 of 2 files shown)


src/common/components/clusterConfiguration/DiskEncryptionFields/DiskEncryptionMode.tsx
```diff
@@ -94,19 +92,18 @@ export const DiskEncryptionMode: React.FC<DiskEncryptionModeProps> = ({
...
-            <StackItem>
-              <TextContent>
-                <Text component={TextVariants.h6}>Tang servers</Text>
-              </TextContent>
-            </StackItem>
-            &nbsp;
-            <TangServers isDisabled={isDisabled} />
-          </Stack>
-        )}
+      {diskEncryptionMode == 'tang' && (
```
jkilzi, line 95:
```suggestion
      {diskEncryptionMode === 'tang' && (
```

jkilzi (approved):

nirfarkas:
Maybe change the wording to:
"To use this encryption, enable TPMv2 encryption in the BIOS of each selected host."

jkilzi:
> Maybe change the wording of the inline warning to: "To use this encryption, enable TPMv2 encryption in the BIOS of each selected host." and we need to add "on" for the workers' label so it will be: "Enable encryption of installation disks on workers"

I would suggest also saying "To use this encryption method,..." instead of "To use this encryption,..."

PR #1117, opened 2022-02-02
MGMT-8342 : Update approval popover

rawagner:
Screenshot from 2022-02-02 12-46-43

@ShiranHi Can you take a look ?
I dropped the MAC address field as we dropped it in mass approve modal too.


ShiranHi:
@rawagner the text is geat, @nirfarkas please let us know if you have any input on this. 
Re. the popover: can we make it not that wide, I think it is too hard to read such a long text line, wdyt? 
Could be like this: https://marvelapp.com/prototype/haechg7/screen/83677573

rawagner:
@ShiranHi updated. Im not sure if we want to put the `approving.` on another line ? Seems bit unnecessary to me.

ShiranHi:
@rawagner looks good to me

nirfarkas:
Why do we need the host name in the popover?

nirfarkas:
FYI Allison

rawagner:
> Why do we need the host name in the popover?

not sure, i would prefer to drop it

ShiranHi:
> > Why do we need the host name in the popover?

To make sure the users understand that popover is about this host, and also sometimes the hostname has changed (and then we show the original one)

wolfeallison:
Hi! It feels pretty repetitive in the header and the first sentence of the body. What do you think of removing the first sentence? So it would be: 
"**Approve host to add it to the infrastructure.** 

Make sure that you expect and recognize the host before approving. 
hostname: localhost

Approve host"

rawagner:
> > > Why do we need the host name in the popover?
> 
> To make sure the users understand that popover is about this host, and also sometimes the hostname has changed (and then we show the original one)

uh, i dont do that right now. The hostname that is shown in popup is the changed one. Do we really want to show original ? wouldnt that be confusing ? What if somebody else changed the hostname and now Im deleting..suddenly the hostname is different than the one shown in the table

ShiranHi:
@rawagner I am sorry it was not clear, maybe it should be "Original hostname"? and maybe it should appear only when the name has changed?

jtomasek (approved):

rawagner:
/hold

needs some more discussion with UXD

mareklibra:
Reverting.
@rawagner please file a new PR if you want this or an update in master. I am sorry for the mess, too late notice.

PR #1119, opened 2022-02-02
MGMT-8999: Add hosts tab is displayed for an SNO cluster created via OCM

jgyselov:
https://issues.redhat.com/browse/MGMT-8999


src/ocm/components/AddHosts/utils.ts
```diff
@@ -1,27 +1,22 @@
...
     ) {
-      isAllowed = true;
-    }
-
-    if (cluster.aiCluster && cluster.aiCluster['high_availability_mode'] === 'None') {
-      isAllowed = false;
+      return true;
+    } else if (
+      (cluster.state === 'ready' && cluster.product?.id == 'OCP-AssistedInstaller') ||
+      cluster.cloud_provider?.id == 'baremetal'
```
jkilzi, line 15:
```suggestion
      cluster.cloud_provider?.id === 'baremetal'
```

jgyselov: thx

src/ocm/components/AddHosts/utils.ts
```diff
@@ -1,27 +1,22 @@
...
     ) {
-      isAllowed = true;
-    }
-
-    if (cluster.aiCluster && cluster.aiCluster['high_availability_mode'] === 'None') {
-      isAllowed = false;
+      return true;
+    } else if (
+      (cluster.state === 'ready' && cluster.product?.id == 'OCP-AssistedInstaller') ||
+      cluster.cloud_provider?.id === 'baremetal'
```
jkilzi, line 15: cluster.aiCluster might be defined && the entire condition holds && isSNO(cluster.aiCluster) === true -> BOOM!

jkilzi: I think we should return false immediately if it is SNO, regardless of the status

jgyselov: So like switch the order? First we check if `cluster.aiCluster` is defined, then if it's not SNO and the status last?  @jkilzi

jkilzi:
```javascript
if (cluster.aiCluster) {
      if (!isSNO(cluster.aiCluster)) {
        return cluster.aiCluster.status === "installed";
      } else {
        return false;
      }
    }
```

jgyselov: LGTM, if we want to nest the conditions. :shrug:

jgyselov: The logic stays the same -- all three conditions must be true for the function to be true.

src/ocm/components/AddHosts/types.ts
```diff
@@ -37,5 +37,16 @@ export type OcmClusterType = {
...
 
+  metrics?: {
+    nodes: [
+      {
+        compute?: number;
+        infra?: number;
+        master?: number;
+        total?: number;
+      },
+    ];
```
jkilzi, line 48: OMG, this is not always an array!!!

image

jkilzi (approved):

PR #1133, opened 2022-02-07
NO-ISSUE: Reset tang servers' fields in different scenarios

DanielleChason:
The different scenarios can be seen in the video (https://youtu.be/sfgRzMkisJA). 
For example, if the user fills the url and/or thumbprint fields of the tang server and then regrets and cancel the encryption, those fields will be reset to their initial values. Same thing will happen if he switches from SNO to a regular cluster. 
As long as on of the "enable on masters" and "enable on workers" is on, the tang servers' fields will maintain their current values. 

Relates to #1114

### DIFF
--- a/src/cim/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
+++ b/src/cim/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
@@ -47,7 +47,7 @@ const ValidationSection = ({
           title="Provided cluster configuration is not valid"
           isInline
         >
-          The following fields are not valid:{' '}
+          The following fields are invalid or missing:{' '}
           {errorFields
             .map((field: string) => CLUSTER_DEPLOYMENT_FIELD_LABELS[field] || field)
             .join(', ')}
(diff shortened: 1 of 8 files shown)


jkilzi:
@DanielleChason we need:
- [ ] A JIRA bug in order to track this issue
- [ ] Please update this PR's title, it must include the issue number
- [ ] Link to the PR in the bug and vice-versa

jkilzi:
@DanielleChason we need a tracking JIRA ticket for this PR

jkilzi (dismissed):
TangServers has still 2 issues:
1. The first time an URL/thumbprint is entered a warning is displayed in the console: "Warning: A component is changing an uncontrolled input of type url to be controlled..."
2. While removing a tang server from the list: "Warning: Can't perform a React state update on an unmounted component."

We need an additional JIRA ticket in order to track those issues, the impact is minor since the users won't see these errors in the production bundle. @DanielleChason please open a new bug for those issues.

jkilzi (approved):

PR #1149, opened 2022-02-15
MGMT-9284: Kubeconfig file is always named "kubeconfig"

jgyselov:
https://issues.redhat.com/browse/MGMT-9284

### DIFF
--- a/src/common/components/clusterDetail/KubeconfigDownload.tsx
+++ b/src/common/components/clusterDetail/KubeconfigDownload.tsx
@@ -7,12 +7,19 @@ import { Cluster } from '../../api/types';
 import { ocmClient } from '../../../ocm/api/axiosClient';
 import { getErrorMessage, handleApiError } from '../../../ocm/api/utils';
 import ClustersAPI from '../../../ocm/services/apis/ClustersAPI';
+import { AxiosResponseHeaders } from 'axios';
 
 type KubeconfigDownloadProps = {
   clusterId: Cluster['id'];
   status: Cluster['status'];
   id?: string;
   handleDownload?: () => void;
+};
+
+const getKubeconfigFileName = (headers: AxiosResponseHeaders) => {
+  const fileNameMatch =
+    headers['content-disposition'] && headers['content-disposition'].match(/filename=".*"/);
+  return fileNameMatch ? fileNameMatch[0].slice(10, -1) : 'kubeconfig';
 };
 
 const KubeconfigDownload: React.FC<KubeconfigDownloadProps> = ({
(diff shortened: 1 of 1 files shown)


src/common/components/clusterDetail/KubeconfigDownload.tsx
```diff
@@ -34,7 +34,11 @@ const KubeconfigDownload: React.FC<KubeconfigDownloadProps> = ({
           saveAs(data.url);
         } else {
           const response = await ClustersAPI.downloadClusterCredentials(clusterId, 'kubeconfig');
-          saveAs(response.data, 'kubeconfig');
+
+          const fileNameMatch = response.headers['content-disposition'].match(/filename=".*"/);
+          const fileName = fileNameMatch ? fileNameMatch[0].slice(10, -1) : '';
```
jkilzi, line 39: Some remarks:
1. Please extract to a reusable function (perhaps inside `src/ocm/services`?)
2. Add some error handling, maybe fallback to 'kubeconfig' if necessary for the file name.
3. `response.headers['content-disposition']` === `undefined` when "content-disposition" is not there and, then calling `.match(...)` isn't it problematic?
4. Are we covered inside OCM?

jkilzi (changes requested):

src/common/components/clusterDetail/KubeconfigDownload.tsx
jgyselov: > 1. Please extract to a reusable function (perhaps inside `src/ocm/services`?)

Done
> 2. Add some error handling, maybe fallback to 'kubeconfig' if necessary for the file name.

I had an empty string as fallback but I do agree that 'kubeconfig' is better.
> 3. `response.headers['content-disposition']` === `undefined` when "content-disposition" is not there and, then calling `.match(...)` isn't it problematic?

Resolved
> 4. Are we covered inside OCM?

It works in the OCM, no changes necessary.

jkilzi: Cool! But, there are circular dependencies...

jkilzi (dismissed):

src/common/components/clusterDetail/KubeconfigDownload.tsx
jgyselov: Yes. I cannot import the `ClustersService` file. Instead of creating a new file just for this one function, I put it inside of `KubefigDownload.tsx`. I hope that makes sense.

jkilzi: What if you put the whole download logic inside `ClustersService` and just add to it a `downloadKubeConfig` method. WDYT? (No `React.useCallback` needed with this approach)

jkilzi: Also, in line 60, that arrow function wrapping the `download function` is defeating the `React.useCallback` purpose

jgyselov: Should `saveAs(...)` be a part of this function or should the function return the data/url and optionally a file name? @jkilzi

jkilzi: Your choice

jkilzi (approved):

PR #1162, opened 2022-02-23
MGMT-8599: Add info that Cluster details cannot be edited

rawagner:
Screenshot from 2022-02-23 14-47-11

MGMT-8599 suggests to use tooltip for disabled fields, but all fields are disabled so @nirfarkas agreed to just use the alert. IIRC there was some text proposal but I dont remember. Is the current one ok ? @nirfarkas


mareklibra (approved):

nirfarkas:
Can we tweak the text to:
"This option is not editable after the draft cluster was created."

PR #1175, opened 2022-02-28
MGMT-8285: Implement new UI for dual stack

jgyselov:
https://issues.redhat.com/browse/MGMT-8285

### Switching to dual-stack:

- OVN and disabled automatic DHCP allocation is forced with dual-stack.
- "Advanced networking" fields are automatically shown and the users must configure them.

https://user-images.githubusercontent.com/87187179/160807693-22a0d5aa-d0ec-43a1-8551-f70627bcc0b6.mp4

### Switching to IPv4 ("single-stack"):

- When switching to IPv4, the UI forces SDN and automatic DHCP allocation and removes all second -network fields, which leaves the users with just their IPv4 subnets (if they've configured dual-stack successfully).

https://user-images.githubusercontent.com/87187179/160807717-9612075c-e3a4-4caf-8e1d-301b1272d0ce.mp4

---

### Initial (single-stack) networking form:
image

### Initial (single-stack) networking form with checked advanced networking:
image

### Succesfully configured dual-stack form:
image
image

---

### Validations:
Using IPv6 subnets is not allowed if the users choose IPv4 ("single-stack").

image

### Validations for dual-stack:

- first subnet must be IPv4
- second subnet must be IPv6

Applies for all network types (machine-, cluster-, service-).

image
image


nirfarkas:
Would love to see screenshots and/or video

jgyselov:
### Networking restrictions with IPv6 detected
https://user-images.githubusercontent.com/87187179/156576031-ec93a90f-0035-4a22-81bc-e25a6d779c1b.mp4

If an IPv6 subnet is detected in either `machine`, `cluster`, or `service` network fields, the UI enforces `OVNKubernetes` network type and disables DHCP allocation.

Screenshot from 2022-03-03 14-43-58
Screenshot from 2022-03-03 14-47-14

jgyselov:
image
Review and create page shows stack type and every cidr from the `machineNetworks` object.

nirfarkas:
Looks great, thanks!
I wonder what's the "Allocate VIPs..." checkbox behaviour when switching between IPV4 and 6 and back.
Let's talk about it on Monday.

nirfarkas:
awolfe@redhat.com
Can you please review it?

jgyselov:
> I wonder what's the "Allocate VIPs..." checkbox behaviour when switching between IPV4 and 6 and back.

@nirfarkas  This can be seen in [the video](https://github.com/openshift-assisted/assisted-ui-lib/pull/1175#issuecomment-1058058876). The current behavior is as follows:

- If IPv6 is detected, the UI forces OVN and disables SDN and DHCP allocation.
- Once everything is IPv4 again, the UI enables SDN and DHCP allocation but it doesn't change any of these fields.
- If the users want automatic DHCP allocation, they will first need to remove any IPv4 subnets and select SDN in the advanced networking section. The tooltips should guide them.

jgyselov:
1. The typo has been fixed.
2. We could change the alert title to "Support limited". But since it's not related to dual-stack in any way, I wonder if it wouldn't be better to open a separate ticket in Jira for that.

@wolfeallison

parvathyvr:
Hi @jgyselov I have made 2 small updates:
1)Changed the tooltip text. Marvel link-https://marvelapp.com/prototype/15c16hhj/screen/85460425
After discussing with @nirfarkas I have updated the tooltip text to 
"A cluster with OVN networking type cannot use ‘allocate virtual IPs via DHCP server’ " thinking it will give better clarity in a screen reader scenario.

2) I have changed the text in the modal when user is changing from dual stack to single stack. Marvel link https://marvelapp.com/prototype/15c16hhj/screen/85495571 to match it with the PF4 standards.
@wolfeallison Please take a look and let me know your comments on these changes.

batzionb:
small comment on look and feel - it seems crowded that each cidr and and host prefix have the long help text below. can it be under the aggregated fields?

src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
```diff
@@ -0,0 +1,185 @@
...
+      )}
+
+      {children}
+
+      {!isUserManagedNetworking &&
+        !!clusterFeatureSupportLevels &&
+        clusterFeatureSupportLevels['CLUSTER_MANAGED_NETWORKING_WITH_VMS'] === 'unsupported' &&
+        vmsAlert}
+
+      {!(isMultiNodeCluster && isUserManagedNetworking) && (
```
batzionb, line 150: should this be enabled in SNO? SNO is always user managed

src/ocm/components/clusterConfiguration/networkConfiguration/StackTypeControl.tsx
```diff
@@ -0,0 +1,130 @@
...
+      setFieldValue(
+        'serviceNetworks',
+        [...values.serviceNetworks, { cidr: '', clusterId: clusterId }],
+        false,
+      );
+    }
+  };
+
+  const setStackType = () => {
+    if (values.stackType === 'singleStack') {
```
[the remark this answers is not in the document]
jgyselov: Actually, I was counting on `values.stackType` not being updated and I was setting it later in the `setDualStack` and `setSingleStack` functions. I've changed it now so that the value is taken from `e.target.value`.

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -235,13 +244,11 @@ export const dnsNameValidationSchema = Yup.string()
     message: 'Value "${value}" is not valid DNS name. Example: basedomain.example.com', // eslint-disable-line no-template-curly-in-string
   });
 
-export const hostPrefixValidationSchema = ({
-  clusterNetworkCidr,
-}: Partial<NetworkConfigurationValues>) => {
+export const hostPrefixValidationSchema = ({ cidr }: Partial<ClusterNetwork>) => {
```
batzionb, line 247: revert the name change and the CIM change

src/ocm/components/clusterConfiguration/networkConfiguration/StackTypeControl.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+  };
+
+  return (
+    <>
+      <FormGroup
+        label="Networking stack type"
+        fieldId={getFieldId('stackType', 'radio')}
+        isInline
+        onChange={setStackType}
```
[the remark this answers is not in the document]
batzionb: @rawagner explained to me that it works because onChange propogates to the parent if no child stops the propogation. So this is actually a cool way to implement a radio form group

src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
```diff
@@ -0,0 +1,133 @@
...
+      fieldId="machine-networks"
+      isRequired
+    >
+      <FieldArray name="machineNetworks">
+        {({ push, remove }) => (
+          <Stack>
+            {values.machineNetworks?.map((machineNetwork, index) => {
+              return (
+                <StackItem key={index}>
+                  <Tooltip
```
batzionb, line 63: The remove button tooltip is duplicated three times - perhaps create a component WithRemoveButton that gets the child component of the Tooltip and returns a new component with the remove button tooltip?
Also - for more readability (it took me time to understand that the tooltip is where the remove button is...) you can implement the remove button like this:
 const [showRemoveButton, setShowRemoveButton] = React.useState(false);
  const updateShowRemoveButton = (value: boolean) => {
    if (!enableRemoveHost) {
      setShowRemoveButton(false);
    } else {
      setShowRemoveButton(value);
    }
  };
  return (
    <ComponentWithRemoveButton
      onMouseEnter={() => updateShowRemoveButton(true)}
      onMouseLeave={() => updateShowRemoveButton(false)}
    >
    This is how I implemented it in static ip.
    If we have the WithRemoveButton component I'll be able to reuse it

batzionb:
Regarding validations - Shouldn't they be just red text instead of inline info alerts?

nirfarkas:
@jgyselov Can you please provide screenshots for the error messages?

src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
```diff
@@ -0,0 +1,133 @@
...
+                      }
+                      isDisabled={!hostSubnets.length}
+                      isRequired={isRequired}
+                    />
+                  </Tooltip>
+                </StackItem>
+              );
+            })}
+
+            {values.stackType === 'singleStack' && (
```
batzionb, line 111: consider creating a AddButton component to avoid this code duplication

src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
jgyselov: I'm adding a new component for this but I'm not sure I understand the code you wrote here. Are you suggesting that I should replace the `Tooltip` with it? That would certainly work just as well when it comes to showing and hiding the button. But the tooltip also takes care of that as well as the positioning (which would be much more difficult to do on our own).

I'm hoping that by creating a separate component for this we can get away with using the _Tooltip black magic_ without confusing too many people. @batzionb

jgyselov: I'm adding a new component for this but I'm not sure I understand the code you wrote here. Are you suggesting that I should replace the `Tooltip` with it? That would certainly work just as well when it comes to showing and hiding the button. But the tooltip also takes care of that as well as the positioning (which would be much more difficult to do on our own).

I'm hoping that by creating a separate component for this we can get away with using the _Tooltip black magic_ without confusing too many people. @batzionb

batzionb: sure, you can use the tooltip for it

jgyselov:
> Can you please provide screenshots for the error messages?

> Regarding validations - Shouldn't they be just red text instead of inline info alerts?

@nirfarkas Screenshots of error messages are in the "main" comment - in the Validations section. Is it correct that they are in the colour blue?

nirfarkas:
@jgyselov I think they need to be either warnings or errors. WDYT @parvathyvr  ?

parvathyvr:
> @jgyselov I think they need to be either warnings or errors. WDYT @parvathyvr ?

Hi @nirfarkas @jgyselov ! looking at the validation screenshots,i see that we are kind of warning the user about how the field values should be. So,i think we could use the yellow inline warning alerts for these messages.

jgyselov:
In **dual stack**, we expect the users to configure all the advanced fields, so I think disabling the checkbox like this here would make sense. 
Screenshot from 2022-03-22 10-45-59
@parvathyvr WDYT?

jgyselov:
> In **dual stack**, we expect the users to configure all the advanced fields, so I think disabling the checkbox like this here would make sense. Screenshot from 2022-03-22 10-45-59

@nirfarkas @wolfeallison Please, check this out. I would like to know what you think about disabling the button and the tooltip content. :slightly_smiling_face:

batzionb (approved):

batzionb (approved):

batzionb (dismissed):

batzionb (dismissed):

batzionb (dismissed):

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -66,11 +62,33 @@ export const getSubnetFromMachineNetworkCidr = (machineNetworkCidr?: string) =>
...
+    ),
+  );
+
+export const isAdvNetworkConf = (
+  cluster: Cluster,
+  defaultNetworkValues: Partial<NetworkConfigurationValues>,
+) => {
+  return !(
+    Boolean(cluster.networkType) &&
+    cluster.networkType === 'OpenShiftSDN' &&
```
celdrake, line 84: The assumption I have now, is that if we are using OVN as the default, then it needs to be changed here as well.

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -466,3 +475,63 @@ export const locationValidationSchema = Yup.string()
...
+      (values) => values[1].cidr && Address6.isValid(values[1].cidr),
+    );
+
+export const isSingleStack = (
+  networks1?: (MachineNetwork | ClusterNetwork | ServiceNetwork)[],
+  networks2?: (MachineNetwork | ClusterNetwork | ServiceNetwork)[],
+  networks3?: (MachineNetwork | ClusterNetwork | ServiceNetwork)[],
+) =>
+  !!(
+    (networks1?.every((network) => network.cidr && Address4.isValid(network.cidr)) &&
```
celdrake, line 520: I would extract this logic to a reusable function, as I see the same code is used later in a Component as well.

So you can define either one function and pass a parameter to check for V4 or V6, or create two separate functions for that. (I couldn't think of a better name for the function,)

```
const networkOnlyHasIP4Addresses = (networkList)  => ...;
const networkOnlyHasIP6Addresses = (networkList)  => ...;
```

Then the main code is easier to read:

```
const hasAllIPv4 = networkOnlyHasIP4Addresses(network1) && ...networkOnlyHasIP4Addresses(network3);

const hasAllIPv6 = networkOnlyHasIP6Addresses(network1) && ...networkOnlyHasIP6Addressesd(network3);

return hasAllIPv4 || hasAllIPv6;
```


Also, if the logic about validating the IPs is not straightforward, I think an early return is better.
Meaning, if you get `true` on allV4, you can return true and avoid evaluating all the ipV6 calls.
```
const hasAllIPv4 = networkOnlyHasIP4Addresses(network1) && ...networkOnlyHasIP4Addresses(network3);
if (hasAllIPv4) {
  return true;
}
const hasAllIPv6 = networkOnlyHasIP6Addresses(network1) && ...networkOnlyHasIP6Addressesd(network3);

return hasAllIPv6;
```

src/ocm/components/clusterConfiguration/networkConfiguration/VirtualIPControlGroup.tsx
```diff
@@ -0,0 +1,193 @@
...
+            isRequired
+          >
+            <VipStaticValue
+              vipName="apiVip"
+              cluster={cluster}
+              validationErrorMessage={apiVipFailedValidationMessage}
+            />
+          </FormikStaticField>
+          <FormikStaticField
+            label="Ingress Virtual IP"
```
celdrake, line 159: I have two PRs in which I removed references to "Virtual IP" and instead just refer to "IP".
Since this is a new file, it probably won't show any conflicts, but if my issues get merged first, then this would be a regression. We can keep an eye on what gets merged first to solve this

https://github.com/openshift-assisted/assisted-ui-lib/pull/1274
https://github.com/openshift-assisted/assisted-ui-lib/pull/1263
https://github.com/openshift-assisted/assisted-ui-lib/pull/1259

src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  useEffect(() => {
+    if (values.stackType === 'dualStack' && !isUserManagedNetworking) {
+      toggleAdvConfiguration(true);
+    }
+  }, [values.stackType, toggleAdvConfiguration, isUserManagedNetworking]);
+
+  const enableSDN = useMemo(
+    () =>
+      values.machineNetworks?.every((network) => network.cidr && Address4.isValid(network.cidr)) &&
```
celdrake, line 110: This is the place where I say you could reuse the extracted function I named `networkOnlyHasIPXAddresses`

celdrake (commented):
Hi! Sorry it's a bit too late, but I looked at the PR for the issue with OVN / SDN and made some comments. Feel free to address how you see better :)

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const clusterNetworkHostPrefixHelperText = (index: number) =>
+    isSubnetIPv6(index)
+      ? 'The subnet prefix length to assign to each individual node. For example, if Cluster Network Host Prefix is set to 116, then each node is assigned a /116 subnet out of the given cidr (clusterNetworkCIDR), which allows for 4,094 (2^(128 - 116) - 2) pod IPs addresses. If you are required to provide access to nodes from an external network, configure load balancers and routers to manage the traffic.'
+      : 'The subnet prefix length to assign to each individual node. For example, if Cluster Network Host Prefix is set to 23, then each node is assigned a /23 subnet out of the given cidr (clusterNetworkCIDR), which allows for 510 (2^(32 - 23) - 2) pod IPs addresses. If you are required to provide access to nodes from an external network, configure load balancers and routers to manage the traffic.';
+
+  return (
+    <Grid hasGutter>
+      <FieldArray name="clusterNetworks">
+        {() => (
```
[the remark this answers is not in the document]
batzionb: @celdrake this is the format formik expects for FieldArray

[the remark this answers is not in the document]
celdrake: Oh, I didn't know that, thanks!

src/common/components/ui/formik/validationSchemas.ts
batzionb: yeah, I agree, it would be nice to have this logic in a function

src/common/components/clusterConfiguration/utils.ts
jgyselov: I agree that this will have to be changed if SDN is no longer default for all clusters. Maybe it will also need to take into consideration whether the cluster is single-node or multi-node. I don't want to do any changed before we get a final verdict on the defaults.

src/common/components/ui/formik/validationSchemas.ts
jgyselov: I have added two functions: `allSubnetsIPv4`and `allSubnetsIPv6` functions and I'm using those.

src/ocm/components/clusterConfiguration/networkConfiguration/networkConfigurationValidation.ts
```diff
@@ -1,48 +1,42 @@
...
-    clusterNetworks?.every((network) => network.cidr && Address6.isValid(network.cidr))) &&
-  (serviceNetworks?.every((network) => network.cidr && Address4.isValid(network.cidr)) ||
-    serviceNetworks?.every((network) => network.cidr && Address6.isValid(network.cidr)));
-
 export const getNetworkInitialValues = (
   cluster: Cluster,
   defaultNetworkValues: Partial<NetworkConfigurationValues>,
 ): NetworkConfigurationValues => {
+  const isSNOCluster = isSNO(cluster);
+  const usesIPv6 = !(
```
batzionb, line 24: looks like a bug... usesIpv6 checks if networks are in Ipv4... I'm pretty sure you'll see the bug if you open an existing cluster with IPV6 that wasn't assigned a network type yet (it's pretty hard to reproduce this)

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
```diff
@@ -0,0 +1,131 @@
...
+    e: React.ChangeEvent<HTMLInputElement>,
+    index: number,
+  ) => {
+    if (isNaN(parseInt(e.target.value))) {
+      setFieldValue(`clusterNetworks.${index}.hostPrefix`, clusterNetworkCidrPrefix(index));
+    }
+  };
+
+  const isSubnetIPv6 = (index: number) =>
+    Address6.isValid((values.clusterNetworks && values.clusterNetworks[index].cidr) || '');
```
batzionb, line 44: This won't detect IPv6 properly, it should look at formik field hostSubnet

src/ocm/components/clusterConfiguration/networkConfiguration/networkConfigurationValidation.ts
jgyselov: It's negated, so it should evaluate `true` if there is at least one IPv6 subnet. Since these data that come from the BE, I think we can assume they'll be valid. We cannot use `allSubnetsIPv6` because that wouldn't work with dual-stack where the IP versions are mixed.

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
jgyselov: I'm confused. There is no such field `hostSubnet` in the form. This is used for the cluster networks subnet+host-prefix pairs to decide the helper texts and min/max values for for each pair. Why would this not detect IPv6 properly?

src/ocm/components/clusterConfiguration/networkConfiguration/networkConfigurationValidation.ts
batzionb: arghh... I keep missing the "not" operator

src/ocm/components/clusterConfiguration/networkConfiguration/VirtualIPControlGroup.tsx
batzionb: @jkilzi please merge Celia's PRs regarding virtual ip -> ip after dual stack is merged

jkilzi: ack

celdrake: @batzionb  @jkilzi Since these are new files, I will need to also update the PR to changes these new labels. I'll update Jira to keep it in mind.

batzionb (dismissed):

src/common/components/clusterConfiguration/networkConfigurationValidation.ts
```diff
@@ -9,33 +9,12 @@ import {
   vipValidationSchema,
 } from '../ui';
 
-import { getSubnetFromMachineNetworkCidr, getHostSubnets } from './utils';
```
batzionb, line 12: code change in common - since code is copied might as well leave as is

jgyselov: Reverted. Thanks for spotting it.

batzionb (approved):

batzionb (dismissed):

src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+        if (values.vipDhcpAllocation) {
+          delete params.apiVip;
+          delete params.ingressVip;
+        } else if (values.stackType === IPV4_STACK) {
+          delete params.machineNetworks;
+        }
+      }
+
+      const { data } = await ClustersAPI.update(cluster.id, params);
+      actions.resetForm({ values: getNetworkInitialValues(data, defaultNetworkValues) });
```
rawagner, line 195: this again introduces the race and hard to follow behavior.
If the latest build passes QE, this should be changed to not reset the form at all & use autosave hook
@jkilzi WDYT ?

jkilzi: agree

jgyselov: `resetForm` removed.

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -66,15 +62,34 @@ export const getSubnetFromMachineNetworkCidr = (machineNetworkCidr?: string) =>
...
+    array2.find(
+      (network) =>
+        network.cidr === clusterNetwork.cidr && network.hostPrefix === clusterNetwork.hostPrefix,
+    ),
+  );
+
 export const isAdvNetworkConf = (
   cluster: Cluster,
-  defaultNetworkSettings: ClusterDefaultConfig,
+  defaultNetworkValues: Partial<NetworkConfigurationValues>,
```
rawagner, line 80: I would prefer `Pick` instead of `Partial` as it gives better idea on what is actually expected

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -66,15 +62,34 @@ export const getSubnetFromMachineNetworkCidr = (machineNetworkCidr?: string) =>
...
   defaultNetworkType: string,
-) =>
-  selectClusterNetworkCIDR(cluster) !== defaultNetworkSettings.clusterNetworkCidr ||
-  selectClusterNetworkHostPrefix(cluster) !== defaultNetworkSettings.clusterNetworkHostPrefix ||
-  selectServiceNetworkCIDR(cluster) !== defaultNetworkSettings.serviceNetworkCidr ||
-  (Boolean(cluster.networkType) && cluster.networkType !== defaultNetworkType);
+) => {
+  return !(
+    Boolean(cluster.networkType) &&
```
rawagner, line 84: any reason to use `Boolean` instead of `!!` (which is much faster) ?

src/common/components/ui/formik/AddButton.tsx
```diff
@@ -0,0 +1,33 @@
+import { Button, ButtonProps, ButtonVariant } from '@patternfly/react-core';
+import { PlusCircleIcon } from '@patternfly/react-icons';
+import { FieldArrayRenderProps } from 'formik';
+import React from 'react';
+
+interface AddButtonProps {
+  add: FieldArrayRenderProps['push'];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  addValue: any;
```
rawagner, line 9: better API for this component would be
```suggestion
  onAdd: VoidFunction
```

usage
```
<AddButton onAdd={() => push({ url: '', thumbprint: '' })} isInline>
  Add another Tang server
</AddButton>
```

src/common/components/ui/formik/RemovableField.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface RemovableFieldProps {
+  index: number;
+  remove: FieldArrayRenderProps['remove'];
+  showRemoveButton?: boolean;
+  buttonVariant?: ButtonVariant;
+}
+
+const REMOVE_BUTTON_EXIT_DELAY = 1500;
+
+const RemovableField: React.FC<RemovableFieldProps & Partial<TooltipProps>> = ({
```
rawagner, line 16: so what is `Partial<TooltipProps>` ? can you use `Omit` or `Pick` instead ?

src/common/components/ui/formik/RemovableField.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { Button, ButtonVariant, Tooltip, TooltipProps } from '@patternfly/react-core';
+import { MinusCircleIcon } from '@patternfly/react-icons';
+import { FieldArrayRenderProps } from 'formik';
+import './RemovableField.css';
+
+interface RemovableFieldProps {
+  index: number;
+  remove: FieldArrayRenderProps['remove'];
```
rawagner, line 9: similar to `AddButton`, lets just have `onRemove: VoidFunction`
```suggestion
  onRemove: VoidFunction
```

src/common/components/ui/formik/RemovableField.tsx
```diff
@@ -0,0 +1,44 @@
...
+  showRemoveButton?: boolean;
+  buttonVariant?: ButtonVariant;
```
rawagner, line 11: this is not actually used anywhere in the component IIUC

src/common/components/ui/formik/RemovableField.tsx
```diff
@@ -0,0 +1,44 @@
...
+      exitDelay={REMOVE_BUTTON_EXIT_DELAY}
+      flipBehavior={['right', 'bottom']}
+      position="right-start"
+      distance={1}
+      content={
+        <Button variant={ButtonVariant.plain} onClick={() => remove(index)}>
+          <MinusCircleIcon size="sm" />
+        </Button>
+      }
+      className={'remove-button--tooltip ' + className}
```
rawagner, line 36: use `classnames` package that we already depend on
```suggestion
      className={classnames(className, 'remove-button--tooltip)}
```

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -127,10 +129,18 @@ export const vipRangeValidationSchema = (
     } catch (err) {
       return this.createError({ message: err.message });
     }
-    const foundHostSubnet = hostSubnets.find((hn) => hn.subnet === hostSubnet);
-    if (foundHostSubnet?.subnet) {
-      // Workaround for bug in CIM backend. hostIDs are empty
-      return foundHostSubnet.hostIDs.length ? isInSubnet(value, foundHostSubnet.subnet) : true;
+    const foundHostSubnets = hostSubnets.filter((hn) =>
+      machineNetworks?.map((network) => network.cidr).includes(hn.subnet),
```
rawagner, line 133: we can move the map higher to do it only once

src/common/selectors/clusterSelectors.ts
```diff
@@ -45,3 +45,18 @@ export const isSubnetInIPv6 = ({
     Address6.isValid(serviceNetworkCidr || '')
   );
 };
+
+export const allSubnetsIPv4 = (
+  networks: (MachineNetwork | ClusterNetwork | ServiceNetwork)[] | undefined,
+) => {
+  return Boolean(networks?.every((network) => network.cidr && Address4.isValid(network.cidr)));
```
rawagner, line 52:
```suggestion
  return !!networks?.every((network) => network.cidr && Address4.isValid(network.cidr));
```

src/ocm/components/clusterConfiguration/ReviewCluster.tsx
```diff
@@ -47,14 +53,24 @@ const ReviewCluster: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
...
         testId="cpu-architecture"
       />
       <DetailItem
-        title="Management network CIDR"
-        value={selectClusterNetworkCIDR(cluster)}
+        title="Cluster network CIDR"
+        value={
+          <>
+            {cluster.machineNetworks?.map((network, index) => (
+              <span key={index}>
```
rawagner, line 67: cant  `network.cidr`  be used as a `key` here ?

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+  };
+
+  const isSubnetIPv6 = (index: number) => {
+    if (isDualStack) {
+      return Boolean(index);
+    } else {
+      return false;
+    }
+  };
```
rawagner, line 62:
```suggestion
  const isSubnetIPv6 = (index: number) => isDualStack ? !!index : false;
```

src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { HostSubnets, NetworkConfigurationValues } from '../../../../common/types';
+import { DUAL_STACK, NO_SUBNET_SET } from '../../../../common/config/constants';
+import { SelectField } from '../../../../common/components/ui';
+
+export interface AvailableSubnetsControlProps {
+  clusterId: Cluster['id'];
+  hostSubnets: HostSubnets;
+  hosts: Host[];
+  isRequired: boolean;
+  isMultiNodeCluster: boolean;
```
rawagner, line 14: `hosts and `isMultiNodeCluster` props are not used IIUC

rawagner (changes requested):

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -66,15 +62,34 @@ export const getSubnetFromMachineNetworkCidr = (machineNetworkCidr?: string) =>
...
+  array1.every((network) => array2.map((elem) => elem.cidr).includes(network.cidr));
+
+const clusterNetworksEqual = (array1: ClusterNetwork[], array2: ClusterNetwork[]) =>
+  array1.length === array2.length &&
+  array1.every((clusterNetwork) =>
+    array2.find(
+      (network) =>
+        network.cidr === clusterNetwork.cidr && network.hostPrefix === clusterNetwork.hostPrefix,
+    ),
+  );
```
rawagner, line 76: could be also done with lodash `xorWith` but I also fine with the current implementation

src/ocm/components/clusterConfiguration/networkConfiguration/VirtualIPControlGroup.tsx
```diff
@@ -0,0 +1,193 @@
...
+  )}`;
+
+  const {
+    'api-vip-defined': apiVipFailedValidationMessage,
+    'ingress-vip-defined': ingressVipFailedValidationMessage,
+  } = React.useMemo(() => getVipValidationsById(cluster.validationsInfo), [
+    cluster.validationsInfo,
+  ]);
+
+  const enableAllocation = useMemo(() => values.networkType === NETWORK_TYPE_SDN, [values]);
```
rawagner, line 110: is this really worth memoizing ? :)

src/common/components/clusterConfiguration/DiskEncryptionFields/TangServers.tsx
```diff
@@ -24,24 +15,9 @@ export const TangServers: React.FC<{ isDisabled?: boolean }> = ({ isDisabled = f
...
-                      variant={ButtonVariant.link}
-                      onClick={() => {
-                        remove(index);
-                      }}
-                    >
-                      <MinusCircleIcon size="sm" />
-                    </Button>
-                  }
+                <RemovableField
+                  showRemoveButton={index == 0 || isDisabled}
```
rawagner, line 19:
```suggestion
                  showRemoveButton={index === 0 || isDisabled}
```

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <FormGroup fieldId="clusterNetworks" labelInfo={isDualStack && 'Primary'}>
+            {values.clusterNetworks?.map((_, index) => {
+              return (
+                <StackItem key={index} className={'network-field-group'}>
+                  <InputField
+                    name={`clusterNetworks.${index}.cidr`}
+                    label="Cluster network CIDR"
+                    helperText={clusterCidrHelperText}
+                    isRequired
+                    labelInfo={index == 0 && values.stackType == DUAL_STACK ? 'Primary' : ''}
```
rawagner, line 74:
```suggestion
                    labelInfo={index === 0 && values.stackType === DUAL_STACK ? 'Primary' : ''}
```

rawagner, line 74:
```suggestion
                    labelInfo={index === 0 && values.stackType === DUAL_STACK ? 'Primary' : ''}
```
```suggestion
                    labelInfo={index == 0 && values.stackType == DUAL_STACK ? 'Primary' : ''}
```

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
```diff
@@ -0,0 +1,135 @@
...
+        {() => (
+          <FormGroup fieldId="serviceNetworks" labelInfo={isDualStack && 'Primary'}>
+            {values.serviceNetworks?.map((_, index) => (
+              <StackItem key={index} className={'network-field-group'}>
+                <InputField
+                  name={`serviceNetworks.${index}.cidr`}
+                  label="Service network CIDR"
+                  helperText={serviceCidrHelperText}
+                  isRequired
+                  labelInfo={index == 0 && isDualStack ? 'Primary' : ''}
```
rawagner, line 116:
```suggestion
                  labelInfo={index === 0 && isDualStack ? 'Primary' : ''}
```

rawagner, line 116:
```suggestion
                  labelInfo={index === 0 && isDualStack ? 'Primary' : ''}
```

src/common/components/ui/formik/RemovableField.css
```diff
@@ -0,0 +1,9 @@
+.remove-button--tooltip {
```
rawagner, line 1: I know this is not yet used widely, but we should try to follow BEM and ideally prefix all our rules with `ai`
```suggestion
.ai-remove-button__tooltip {
```

rawagner, line 1: we should try to follow BEM
```suggestion
.remove-button__tooltip {
```
```suggestion
.remove-button--tooltip {
```

src/common/components/ui/formik/RemovableField.css
```diff
@@ -0,0 +1,9 @@
...
+  --pf-c-tooltip__content--BackgroundColor: rgba(0, 0, 0, 0);
+  --pf-c-tooltip--BoxShadow: rgba(0, 0, 0, 0);
+  --pf-c-tooltip__content--PaddingTop: 2.25em;
+}
+
+.remove-button--tooltip--Padding-none {
```
rawagner, line 7:
```suggestion
.ai-remove-button__tooltip--padding-none {
```

rawagner, line 7:
```suggestion
.ai-remove-button__tooltip--padding-none {
```

src/common/components/ui/formik/RemovableField.tsx
```diff
@@ -0,0 +1,41 @@
...
+      flipBehavior={['right', 'bottom']}
+      position="right-start"
+      distance={1}
+      content={
+        <Button variant={ButtonVariant.plain} onClick={onRemove}>
+          <MinusCircleIcon size="sm" />
+        </Button>
+      }
+      className={classnames('remove-button--tooltip ', className)}
+      {...props}
```
rawagner, line 34: in case the TS definition is correct 
```js
Omit<TooltipProps, 'content'>
```

then this should be 
```suggestion
       {...props}
      content={
        <Button variant={ButtonVariant.plain} onClick={onRemove}>
          <MinusCircleIcon size="sm" />
        </Button>
      }
      className={classnames('remove-button--tooltip ', className)}
```

otherwise `props` can override `content`

rawagner (changes requested):
just a few small nits.

we need to add eslint rule for `==` vs `===` which is really easy to miss. I will create a PR for that

src/common/components/clusterConfiguration/DiskEncryptionFields/TangServers.tsx
jgyselov: Good point, thank you.

src/ocm/components/clusterConfiguration/networkConfiguration/AdvancedNetworkFields.tsx
celdrake: We could enhance our  `eslint` config to highlight this type of issues, and have them fixed automatically.
https://eslint.org/docs/rules/eqeqeq

Maybe we could create a separate ticket, as there could be other interesting rules to add, etc.

rawagner: totally agree https://github.com/openshift-assisted/assisted-ui-lib/pull/1175#pullrequestreview-935990044 :) Im already working on it

rawagner (approved):
great work!

jkilzi (approved):
Despite I'm not 100% sure about the UX, specifically the duplication of the cluster/service network fields when using dual-stack, this PR, in its current state, is functional and good enough.
@nirfarkas we should work on improving the UX a bit.

PR #1176, opened 2022-03-01
Remove AI lib logging through console.log.

rawagner:
The version is already included in[ html tree](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/ocm/components/Router.tsx#L21)

### DIFF
--- a/src/common/hacks.ts
+++ b/src/common/hacks.ts
@@ -1,11 +1,6 @@
 import React from 'react';
-import { getAssistedUiLibVersion } from './config';
 
 // TODO(jtomasek): Stop using this once https://github.com/developit/microbundle/issues/564#issuecomment-593146626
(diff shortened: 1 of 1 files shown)


rawagner:
@jkilzi I talked to QE and they are aware & will update tests if needed

src/common/hacks.ts
```diff
@@ -1,11 +1,6 @@
 import React from 'react';
-import { getAssistedUiLibVersion } from './config';
 
 // TODO(jtomasek): Stop using this once https://github.com/developit/microbundle/issues/564#issuecomment-593146626
 // is fixed, use `microbundle --jsxFragment React.Fragment` instead.
 // Workaround for TS17016
 window['Fragment'] = React.Fragment;
```
jkilzi, line 6: BTW, I don't think we need this too ^ anymore

jkilzi (approved):

PR #1183, opened 2022-03-04
Improve create cluster wizard flow

rawagner:


src/cim/components/ClusterDeployment/ClusterDeploymentWizardFooter.tsx
```diff
@@ -96,30 +79,18 @@ const ClusterDeploymentWizardFooter = ({
...
-      <ValidationSection
-        errorFields={errorFields}
-        requireProxy={requireProxy}
-        currentStepId={currentStepId}
-        validationsInfo={validationsInfo}
-        clusterStatus={clusterStatus}
-        hosts={hosts}
-      />
-    ) : undefined;
+  const errorsSection = (
```
jtomasek, line 84: This reintroduces the problem with the footer height since the AlertGroup is always rendered and takes space.

jtomasek: Let's fix this in a follow up

jtomasek (approved):

PR #1206, opened 2022-03-14
MGMT-8763 Persist DHCP / IP config when switching Network Management

celdrake:
https://issues.redhat.com/browse/MGMT-8763

When switching Network Management mode to "User managed", the DHCP / IP configuration should be persisted so that if the switch is done back to "Cluster managed", they are restored.

This implementation uses `localStorage` as required.
A prop called `persistDhcpSetting` has been added to `NetworkConfiguration` component. Only if this is true, the functionality is triggered. This is so the behaviour for CIM is not altered for now.


src/ocm/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -39,7 +39,7 @@ import { onFetchEvents } from '../fetching/fetchEvents';
 import { TIME_ZERO, VSPHERE_CONFIG_LINK } from '../../../common/config/constants';
 import { isSNO } from '../../../common/selectors/clusterSelectors';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
-import { diffInDaysBetweenDates } from '../../../common/sevices/DateAndTime';
+import { diffInDaysBetweenDates } from '../../../common/services/DateAndTime';
```
jkilzi, line 42: OMG!

src/common/services/NetworkSettingsService.ts
```diff
@@ -0,0 +1,28 @@
...
+type PersistDhcpConfig = Pick<
+  NetworkConfigurationValues,
+  'ingressVip' | 'apiVip' | 'vipDhcpAllocation'
+>;
+
+const STORAGE_KEY = 'network-dhcp-settings';
+
+const NetworkSettingsService = {
+  getPersistedDhcpConfig(): PersistDhcpConfig | null {
+    const config = localStorage.getItem(STORAGE_KEY);
```
jkilzi, line 12: I was thinking about using `sessionStorage` for this since once the users close the tab/window that leftover information is not relevant anymore (and actually may provide a false state)

jkilzi (commented):
One nit about the use of Local Storage...

rawagner (changes requested):
To logic when to store & when to load does not seem to be correct. 

Try:
 - fill in VIPs (autosave will store the values to backend)
 - switch to user networking (VIPs are stored to local storage)
 - switch back and change VIPs again (autosave will store the values to backend)
 - cancel wizard (new VIPs are not stored in local storage
 - go back to edit
 - VIPs stored in backend are overridden with values from your local storage which is incorrect

Another broken use-case:
 - fill in VIPs (autosave will store the values to backend)
 - switch to user networking (VIPs are stored to local storage)
 - switch back and change VIPs again (autosave will store the values to backend)
 - click Back, click Next -> you will see old values from local storage again.
 
 
 Another thing that Im not happy about is that we use these values from `network-dhcp-settings` local storage for every cluster

rawagner:
Just noticed @jkilzi feedback. Changing localstorage to sessionstorage solves my last complain


> It makes sense to me to remember the values when switching between `Network Management` types but Im not really sure if persisting these information between sessions is a good idea - for how many clusters are we planning to do it ? cc @jkilzi

jkilzi:
Replying to @rawagner 
> ... I'm not really sure if persisting this information between sessions is a good idea
Me neither, I suggested to use Session Storage instead...

> ... for how many clusters are we planning to do it
For any available, so this implementation won't work when there is more than one cluster... @celdrake FYI

celdrake:
> To logic when to store & when to load does not seem to be correct.
...
> 
> It makes sense to me to remember the values when switching between `Network Management` types but Im not really sure if persisting these information between sessions is a good idea - for how many clusters are we planning to do it ? cc @jkilzi

Yes, you are both right, I hadn't thought of those edge cases. I'll check more scenarios as well as using `sessionStorage` as well.

celdrake:
@jkilzi @rawagner Please check the last commit, the issues you reported should be fixed.

First, by making sure that the persistence / restoration of DHCP config only happens when "Network Management" changes. That prevents the restoration to overwrite newer values (IPs) set by the user.

Second, instead of using `sessionStorage`, I thought of clearing the cache when the Wizard leaves the Network step. In that way, the data will always be associated to the Cluster being edited at the moment. The downside is, if the user cancel the Wizard, the info would not be deleted, but because of the first point, it wouldn't be applied.

What do you think?

rawagner:
Looks like the current implementation should work fine but I think we either want to simplify it or change when we forget (delete) the VIPs.

Current situation is that we forget the VIPs any time the `NetworkConfigurationForm` is unmounted. 
 - Im not sure if that is ok ? @jkilzi 
 
If yes, then we do not need to use localstorage/sessionstorage. We can just store the old values to React state.
If not, then the forget/delete logic has to be changed

celdrake:
After merging Dual Stack, it is not enough to rebase the branch but we need to revisit the feature and make sure it works in all existing and new scenarios.
Closing for now.

PR #1212, opened 2022-03-16
Bug 2039789: Fix cluster name validation when not using RedHat DNS

rawagner:


src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -36,6 +40,13 @@ export const nameValidationSchema = (usedClusterNames: string[], baseDnsDomain =
         const clusterFullName = `${value}.${baseDnsDomain}`;
         return !value || !usedClusterNames.includes(clusterFullName);
       }),
+      otherwise: Yup.string().test(
+        'is-name-unique',
+        'The name is already taken.',
+        (value: string) => {
+          return validateUniqueName ? !value || !usedClusterNames.includes(value) : true;
```
mareklibra, line 47: @rawagner , please confirm that the `usedClusterNames` do not contain full domain names as for the other case here.
If you can confirm that, a small comment describing why it is different in that flow will help future maintenance.

rawagner: `usedClusterNames` are clusterDeployment CR names see https://github.com/stolostron/console/blob/main/frontend/src/routes/Infrastructure/Clusters/ManagedClusters/CreateCluster/components/assisted-installer/DetailsForm.tsx#L119

I also added a comment to the code

jtomasek (approved):

PR #1228, opened 2022-03-18
MGMT-9411: Fix host re-render

celdrake:
https://issues.redhat.com/browse/MGMT-9411

After investigating the issue, I found out that this might be a performance problem on slower computers.

This PR avoid unnecessary re-rendering of complex components, such as the Host table.

. On NetworkConfigurationForm, on form submit, only update the "base" cluster data, and not the `hosts` property. This prevents the hosts table being re-rendered.

- Avoid rendering all modals, even when they are not open.

- Memoized `ExpandComponent` on `NetworkConfigurationTable` based on profiling results.

These changes should improve the performance of the Network config page.

### DIFF
--- a/src/ocm/reducers/clusters/currentClusterSlice.ts
+++ b/src/ocm/reducers/clusters/currentClusterSlice.ts
@@ -45,9 +45,17 @@ export const currentClusterSlice = createSlice({
   initialState,
   name: 'currentCluster',
   reducers: {
-    updateCluster: (state, action: PayloadAction<Cluster>) => ({ ...state, data: action.payload }),
+    updateClusterBase: (state, action: PayloadAction<Cluster>) => {
+      // Should not overwrite the hosts once they are created
+      const originalHosts = state.data?.hosts || [];
+      return ({ ...state, data: { ...action.payload, hosts: originalHosts } });
+    },
+    updateCluster: (state, action: PayloadAction<Cluster>) => {
+      return ({ ...state, data: action.payload  });
+    },
     updateHost: (state, action: PayloadAction<Host>) => {
       const hostIndex = _.findIndex(state.data?.hosts, (host) => host.id === action.payload.id);
+
       if (hostIndex >= 0) {
         _.set(state, `data.hosts[${hostIndex}]`, action.payload);
       }
@@ -87,6 +95,7 @@ export const currentClusterSlice = createSlice({
 
 export const {
   updateCluster,
+  updateClusterBase,
   updateHost,
   cleanCluster,
   forceReload,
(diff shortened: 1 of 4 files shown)


src/ocm/reducers/clusters/currentClusterSlice.ts
```diff
@@ -45,9 +45,17 @@ export const currentClusterSlice = createSlice({
   initialState,
   name: 'currentCluster',
   reducers: {
-    updateCluster: (state, action: PayloadAction<Cluster>) => ({ ...state, data: action.payload }),
+    updateClusterBase: (state, action: PayloadAction<Cluster>) => {
+      // Should not overwrite the hosts once they are created
```
rawagner, line 49: Im not sure about this as we may lose a legitimate updates to hosts.
It also has a potential for future weird bugs which are hard to debug.

How about doing deep equals on hosts array ? Would that be too heavy to do ?
Or we could tweak the autosave debounce ?

celdrake: The data that OCM uses for `hosts` comes from a specific endpoint which is fetched in each polling cycle, and then overwrites the data that was fetched from the `clusterDetail` endpoint. I think the original hosts data is being deprecated.
https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/ocm/reducers/clusters/currentClusterSlice.ts#L28

It may be a good idea to split the two properties to simplify management of the data. Maybe it can be treated as a tech debt issue?

celdrake: Also, I tried to modify this `fetchClusterAsync` function so that it would keep the array object references and only update the  objects data, but it didn't seem to work. (That was so that polling wouldn't make the entire host table either)

jkilzi: > How about doing deep equals on hosts array? Would that be too heavy to do?
It is as expensive as copying the data.

> Or we could tweak the autosave debounce?
The debounce won't help either. Instead of getting the feel the UI gets stuck every 1sec it will get stuck now every X seconds.

Actually, the hosts in OCM are updated from the polling mechanism, there is no need to get the latest hosts data from the response to the cluster update API in the networking section because, right now, we are patching the server with cluster-related params. Info about the hosts is read-only in this screen.

rawagner: Ok, thanks for the answers. I wasnt sure how exactly are we getting the updated cluster data. The current solution works for me.

rawagner (approved):

PR #1229, opened 2022-03-18
MGMT-9034: AI Image Naming Discrepancy

jgyselov:
https://issues.redhat.com/browse/MGMT-9034

Download ISO modal:
image
```
wget -O b5514395-7b5a-41b0-8f92-89754206a620-discovery.iso 'http://10.19.130.2:6016/images/b5514395-7b5a-41b0-8f92-89754206a620?arch=x86_64&type=full-iso&version=4.9'
```
Using the "Download Discovery ISO" button:
image

### DIFF
--- a/src/ocm/components/clusterConfiguration/DiscoveryImageSummary.tsx
+++ b/src/ocm/components/clusterConfiguration/DiscoveryImageSummary.tsx
@@ -10,7 +10,8 @@ type DiscoveryImageSummaryProps = {
 };
 
 const DiscoveryImageSummary: React.FC<DiscoveryImageSummaryProps> = ({ cluster, ...restProps }) => {
-  const { imageUrl, error, isLoading } = useInfraEnvImageUrl(cluster.id);
+  const { imageUrl, infraEnvId, error, isLoading } = useInfraEnvImageUrl(cluster.id);
+
   if (error) {
     return <ErrorState />;
   }
@@ -20,7 +21,7 @@ const DiscoveryImageSummary: React.FC<DiscoveryImageSummaryProps> = ({ cluster, 
 
   return (
     <DownloadIso
-      fileName={`discovery_image_${cluster.name}.iso`}
+      fileName={infraEnvId && `${infraEnvId}-discovery.iso`}
       downloadUrl={imageUrl}
       isSNO={isSNO(cluster)}
       {...restProps}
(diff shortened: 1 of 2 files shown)


src/ocm/components/clusterConfiguration/DiscoveryImageSummary.tsx
```diff
@@ -11,6 +12,16 @@ type DiscoveryImageSummaryProps = {
...
+  const [fileName, setFileName] = useState('discovery.iso');
+
+  useEffect(() => {
+    async function setInfraEnvId() {
+      const infraEnvId = await InfraEnvsService.getInfraEnvId(cluster.id);
+      setFileName(`${infraEnvId}-discovery.iso`);
+    }
+    setInfraEnvId();
+  }, [cluster.id]);
+
```
jkilzi, line 24: We should/can get the file name from the content-disposition header
image

jkilzi: If in the future they change the file name format we won't need to do anything...

jkilzi: Right now the HEAD request is blocked by CORS (probably the same will happen in production).
I would recommend asking the BE team to add the CORS headers for this endpoint first...

jgyselov: @jkilzi But this is not for the direct download but for the "wget ..." field. I don't want to download the file in the background just to get its name for the "wget ..." field and then again if the user chooses to press the "Download Discovery ISO" button.

jkilzi: @osherdp FYI

osherdp: @jgyselov what do you mean about downloading the file in the background?
HEAD requests don't download the content itself

as for the CORS, I remember that we couldn't find any evidence for a CORS problems in production
it might still exist for staging though, so if you see anything please open a bug

jkilzi (changes requested):
So we are waiting for the BE to add the CORS headers so we can issue a HEAD request and parse the filename

jgyselov:
> So we are waiting for the BE to add the CORS headers so we can issue a HEAD request and parse the filename

@jkilzi Acknowledged.

src/ocm/components/clusterConfiguration/DiscoveryImageSummary.tsx
jkilzi: Nick's PR is merged we should be able now to call make the API call with the HEAD method

jgyselov:
Stale, closing.

PR #1251, opened 2022-03-24
Use import to lodash functions not the entire module - Common

celdrake:
Changed the imports of specific `lodash` functions to avoid loading the entire `lodash` module and reduce build time / bundle size.

This PR changes only the usages within `common` code.

More details in https://github.com/openshift-assisted/assisted-ui-lib/pull/1250


jkilzi (dismissed):
I would like one more approval :)

jkilzi (approved):

batzionb (approved):

PR #1259, opened 2022-03-28
MGMT-9151 Cluster details - fix inapropriate term 'Virtual' for non-virtual IPs

celdrake:
https://issues.redhat.com/browse/MGMT-9151

In the Cluster Details page, the API / Ingress IPs were always referred to as "virtual".
But whether or not the IP is virtual depends on the network management of the cluster. Only if the management is done by the cluster, can the IP be considered Virtual. This PR fixes that, making the term "Virtual" only appear for cluster-managed networks.

Before:

cluster-details-before

After - Cluster managed networking:
cluster-details-cluster-ok

After - User managed networking:
cluster-details-user-ok


jkilzi:
@celdrake Virtual IP is only relevant when there is more than 1 node in the cluster. Users choosing the user-managed networking can still set up a multi-node cluster along with a load-balancer they bring in.
So the term Virtual IP is only relevant when there is a need for a load-balancer. SNO is the only scenario that doesn't need a load balancer.

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -37,9 +37,14 @@ const getNetworkType = (
   return networkType;
 };
 
-const getManagementType = (clusterManagementType: boolean | undefined): string => {
+const getAPITitle = (apiName: string, isUserManagementType: boolean | undefined) => {
+  const virtualLabel = isUserManagementType ? '' : ' Virtual';
```
jkilzi, line 41:
```suggestion
  const virtualLabel = isUserManagementType ? '' : ' Virtual';
```
Should depend on whether the cluster is not an SNO instead of the `isUserManagementType`

jkilzi (changes requested):

jtomasek (approved):

PR #1261, opened 2022-03-28
Fix table actions

rawagner:
- improve on which tables, the actions are shown
 - add disabled messages
 - improve types

### DIFF
--- a/src/cim/components/Agent/tableUtils.tsx
+++ b/src/cim/components/Agent/tableUtils.tsx
@@ -2,16 +2,21 @@ import * as React from 'react';
 import { sortable, expandable, breakWord } from '@patternfly/react-table';
 import { Link } from 'react-router-dom';
 
-import { Host, HostsTableActions, getInventory, getHostname, Hostname } from '../../../common';
-import { AgentK8sResource, InfraEnvK8sResource } from '../../types';
+import {
+  Host,
+  HostsTableActions,
+  getInventory,
+  getHostname,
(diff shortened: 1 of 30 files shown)


src/cim/components/Agent/tableUtils.tsx
```diff
@@ -222,38 +227,118 @@ export const infraEnvColumn = (agents: AgentK8sResource[]): TableRow<Host> => {
...
+      if (isInstallationInProgress(agentClusterInstall)) {
+        return [false, 'It is not possible to remove a node from a cluster during installation.'];
+      }
+
+      if (
+        ['installed', 'error', 'cancelled'].includes(status.key) &&
+        agent.status?.role === 'worker' &&
+        agentClusterInstall.spec?.provisionRequirements?.workerAgents === 2
+      ) {
+        return [false, 'This host cannot be removed as there are only 2 workers in the cluster.'];
```
jtomasek, line 311: IIUC this will prevent the user from removing workers if there are 2 of them. What if user has 3 workers and removes them? Will we only allow to remove 1 of them? If not, should we enforce this when there are only 2 workers?

I am inclined towards allowing user to remove workers from the cluster. Enforcing various cases on the UI side can lead to bugs. Backend does not enforce it AFAIK.

rawagner: removed

jtomasek (approved):

PR #1263, opened 2022-03-29
MGMT-9151 Do not use 'Virtual' for IPs - CIM

celdrake:
https://issues.redhat.com/browse/MGMT-9151

This is the second part of https://github.com/openshift-assisted/assisted-ui-lib/pull/1259, only including the changes in CIM


celdrake:
It has been confirmed that we want to remove the term "virtual" from the IP titles, etc.

jtomasek (approved):

PR #1264, opened 2022-03-29
Move AssistedUILibVersion component to common and use it in CIM too

rawagner:


src/ocm/components/ui/index.ts
```diff
@@ -1 +1 @@
-export * from './AssistedUILibVersion';
+export { default as RedirectToCluster } from './RedirectToCluster';
```
jkilzi, line 1: What about this part in that file
```
export const redirectToCluster = (history: History, id: string) => {
  history.push(`${routeBasePath}/clusters/${id}`);
};
```

src/common/components/ui/AssistedUILibVersion.tsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
-import { getAssistedUiLibVersion } from '../../../common';
+import { getAssistedUiLibVersion } from '../../config';
 
-export const AssistedUILibVersion: React.FC = ({ children }) => (
+const AssistedUILibVersion: React.FC = ({ children }) => (
```
jkilzi, line 4:
```suggestion
const AssistedUILibVersion = ({ children }: React.PropsWithChildren<React.ReactNode>) => (
```

jkilzi (changes requested):

PR #1271, opened 2022-03-29
MGMT-9341: Interactive text checker in the cloud

jgyselov:
https://issues.redhat.com/browse/MGMT-9341

image
image
image
image

### DIFF
--- a/src/common/components/ui/formik/validationSchemas.ts
+++ b/src/common/components/ui/formik/validationSchemas.ts
@@ -9,6 +9,13 @@ import { ProxyFieldsType } from '../../../types';
 import { trimCommaSeparatedList, trimSshPublicKey } from './utils';
 import { ClusterNetwork, MachineNetwork, ServiceNetwork } from '../../../api/types';
 import { allSubnetsIPv4 } from '../../../selectors';
+import {
+  CLUSTER_NAME_VALIDATION_MESSAGES,
+  HOSTNAME_VALIDATION_MESSAGES,
+  LOCATION_VALIDATION_MESSAGES,
+  NAME_VALIDATION_MESSAGES,
+  UNIQUE_CLUSTER_NAME_VALIDATION_MESSAGES,
+} from './constants';
 
 const ALPHANUMBERIC_REGEX = /^[a-zA-Z0-9]+$/;
 const NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
@@ -27,17 +34,6 @@ const MAC_REGEX = /^([0-9A-Fa-f]{2}[:]){5}([0-9A-Fa-f]{2})$/;
 const BMC_REGEX =
   /^((ipmi|ibmc(\+https?)?|idrac(\+https?)?|idrac-redfish(\+https?)?|idrac-virtualmedia(\+https?)?|irmc|redfish(\+https?)?|redfish-virtualmedia(\+https?)?|ilo4(\+https)?|ilo4-virtuallmedia(\+https)?|ilo5(\+https)?|ilo5-redfish(\+https)|ilo5-virtualmedia(\+https)|https?|ftp):(\/\/([a-z0-9\-._~%!$&'()*+,;=]+@)?([a-z0-9\-._~%]+|\[[a-f0-9:.]+\]|\[v[a-f0-9][a-z0-9\-._~%!$&'()*+,;=:]+\])(:[0-9]+)?(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?|(\/?[a-z0-9\-._~%!$&'()*+,;=:@]+(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?)?)|([a-z0-9\-._~%!$&'()*+,;=@]+(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?|(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)+\/?))(\?[a-z0-9\-._~%!$&'()*+,;=:@/?]*)?(#[a-z0-9\-._~%!$&'()*+,;=:@/?]*)?$/i;
 const LOCATION_CHARS_REGEX = /^[a-zA-Z0-9-._]*$/;
-
-export const CLUSTER_NAME_VALIDATION_MESSAGES = {
-  INVALID_LENGTH: '1-54 characters',
-  INVALID_VALUE: 'Use alphanumberic characters, or hyphen (-)',
-  INVALID_START_END: 'Start and end with a letter or number.',
-};
-
-export const UNIQUE_CLUSTER_NAME_VALIDATION_MESSAGES = {
-  ...CLUSTER_NAME_VALIDATION_MESSAGES,
-  NOT_UNIQUE: 'Must be unique',
-};
 
 export const nameValidationSchema = (
   usedClusterNames: string[],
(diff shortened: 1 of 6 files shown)


src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -9,7 +9,7 @@ import { ProxyFieldsType } from '../../../types';
 import { trimCommaSeparatedList, trimSshPublicKey } from './utils';
 
 const ALPHANUMBERIC_REGEX = /^[a-zA-Z0-9]+$/;
-const CLUSTER_NAME_REGEX = /^[a-z]([-a-z0-9]*[a-z0-9])?$/;
+const CLUSTER_NAME_REGEX = /^[a-z](.*[a-z0-9])?$/;
```
jkilzi, line 12: `.*` means that the name can contain special characters in the middle (e.g. $%^&*#...)
Why is this change necessary?

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -22,28 +22,47 @@ const MAC_REGEX = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})|([0-9a-fA-F]{4}\\.[0
 const BMC_REGEX = /^((ipmi|ibmc(\+https?)?|idrac(\+https?)?|idrac-redfish(\+https?)?|idrac-virtualmedia(\+https?)?|irmc|redfish(\+https?)?|redfish-virtualmedia(\+https?)?|ilo4(\+https)?|ilo4-virtuallmedia(\+https)?|ilo5(\+https)?|ilo5-redfish(\+https)|ilo5-virtualmedia(\+https)|https?|ftp):(\/\/([a-z0-9\-._~%!$&'()*+,;=]+@)?([a-z0-9\-._~%]+|\[[a-f0-9:.]+\]|\[v[a-f0-9][a-z0-9\-._~%!$&'()*+,;=:]+\])(:[0-9]+)?(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?|(\/?[a-z0-9\-._~%!$&'()*+,;=:@]+(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?)?)|([a-z0-9\-._~%!$&'()*+,;=@]+(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?|(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)+\/?))(\?[a-z0-9\-._~%!$&'()*+,;=:@/?]*)?(#[a-z0-9\-._~%!$&'()*+,;=:@/?]*)?$/i;
 const LOCATION_CHARS_REGEX = /^[a-zA-Z0-9-._]*$/;
 
+export const CLUSTER_NAME_VALIDATION_MESSAGES = {
+  INVALID_LENGTH: '1-54 characters',
+  INVALID_VALUE: 'Use alphanumberic characters, or hyphen (-)',
+  INVALID_START_END: 'Must start with a letter and end with a letter or number.',
+};
+
+export const UNIQUE_CLUSTER_NAME_VALIDATION_MESSAGES = {
```
jkilzi, line 31:
```suggestion
export const uniqueClusterNameValidationMessages = {
```

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -22,28 +22,47 @@ const MAC_REGEX = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})|([0-9a-fA-F]{4}\\.[0
 const BMC_REGEX = /^((ipmi|ibmc(\+https?)?|idrac(\+https?)?|idrac-redfish(\+https?)?|idrac-virtualmedia(\+https?)?|irmc|redfish(\+https?)?|redfish-virtualmedia(\+https?)?|ilo4(\+https)?|ilo4-virtuallmedia(\+https)?|ilo5(\+https)?|ilo5-redfish(\+https)|ilo5-virtualmedia(\+https)|https?|ftp):(\/\/([a-z0-9\-._~%!$&'()*+,;=]+@)?([a-z0-9\-._~%]+|\[[a-f0-9:.]+\]|\[v[a-f0-9][a-z0-9\-._~%!$&'()*+,;=:]+\])(:[0-9]+)?(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?|(\/?[a-z0-9\-._~%!$&'()*+,;=:@]+(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?)?)|([a-z0-9\-._~%!$&'()*+,;=@]+(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)*\/?|(\/[a-z0-9\-._~%!$&'()*+,;=:@]+)+\/?))(\?[a-z0-9\-._~%!$&'()*+,;=:@/?]*)?(#[a-z0-9\-._~%!$&'()*+,;=:@/?]*)?$/i;
 const LOCATION_CHARS_REGEX = /^[a-zA-Z0-9-._]*$/;
 
+export const CLUSTER_NAME_VALIDATION_MESSAGES = {
```
jkilzi, line 25: Since this is just an object I think it would be more appropriate to use camelCase

```suggestion
export const clusterNameValidationMessages = {
```

src/common/components/ui/formik/validationSchemas.ts
jkilzi: I would even suggest moving these messages to a different file...

jkilzi (changes requested):

src/common/components/ui/formik/validationSchemas.ts
jgyselov: I wanted the conditions in the list to be simple, straight forward and always deal with one thing at a time.
This regex was good but it tested two things at a time and that didn't allow for separation of the conditions ("you can use these characters" and "name must start and end with ..."). I think that for the users a slightly longer list with shorter instructions is easier to comprehend than a shorter list with long and complicated instructions.

src/common/components/ui/formik/validationSchemas.ts
jgyselov: I was following a "convetion" we had with the other validation messages...
But I do agree, moving all of them to a different file might be a good idea.

nirfarkas:
Looks great!

jkilzi (commented):
LGTM
@rawagner PTAL

jkilzi:
@jgyselov we need to cover this with integration tests

jkilzi:
@jgyselov please check if the cluster name can contain capital letters, your regex won't allow it (also the previous one) but the rules don't explicitly say the characters must be all lowercase letters

jgyselov:
> @jgyselov please check if the cluster name can contain capital letters, your regex won't allow it (also the previous one) but the rules don't explicitly say the characters must be all lowercase letters

BE response: 
```
Cluster name format is not valid: 'CLUSTER'. Name must consist of lower-case letters, numbers and hyphens.
It must start with a letter and end with a letter or number.
```
I've modified the regexes and validation messages to reflect that only lowercase letters are allowed for both cluster names and host names.
@jkilzi

jkilzi:
@jgyselov can you ptal, there are merge conflicts

jgyselov:
> can you ptal, there are merge conflicts

Resolved. @jkilzi

jkilzi (approved):

jkilzi:
@jgyselov we need to port this to releases/v2.4

PR #1272, opened 2022-03-29
MGMT-9176 Network type selection rules

celdrake:
https://issues.redhat.com/browse/MGMT-9176

This PR makes sure that Assisted Installer sets valid configurations for "networkType" .

- SNO clusters and Multi-node clusters using ipV6, they must use `OVNKubernetes` as its Network Type.
- Multi-node clusters using ipV4 can use either `OVNKubernetes` or `OpenShiftSDN`. However, when a cluster is created, its default Network Type must be `OpenShiftSDN`. Users can modify this to OVN by toggling the Advanced network options and selecting it.

The fix includes:
- Setting the default Network Type on the API call to create a new cluster. This will depend on the cluster type as explained above.
- Disabling the SDN option when the cluster type does not allow it.
- Updating the Network Type of a cluster when the subnets are selected and it changes from ipV4 to ipV6 or viceversa.

SNO cluster / Multi-node cluster using ipV6:  OVN is the default, SDN is not selectable
network-type-sno

Multi-node cluster using ipV4:  SDN is the default, OVN is selectable
network-type-mc

### DIFF
--- a/src/common/components/clusterConfiguration/utils.ts
+++ b/src/common/components/clusterConfiguration/utils.ts
@@ -1,6 +1,6 @@
 import { Address4, Address6 } from 'ip-address';
 import { Cluster, ClusterDefaultConfig, Inventory, stringToJSON } from '../../api';
-import { NETWORK_TYPE_OVN, NO_SUBNET_SET } from '../../config';
+import { NO_SUBNET_SET } from '../../config';
 import {
   HostDiscoveryValues,
   HostSubnets,
@@ -66,11 +66,15 @@ export const getSubnetFromMachineNetworkCidr = (machineNetworkCidr?: string) => 
   return subnet?.address;
 };
 
-export const isAdvNetworkConf = (cluster: Cluster, defaultNetworkSettings: ClusterDefaultConfig) =>
+export const isAdvNetworkConf = (
+  cluster: Cluster,
+  defaultNetworkSettings: ClusterDefaultConfig,
+  defaultNetworkType: string,
+) =>
   selectClusterNetworkCIDR(cluster) !== defaultNetworkSettings.clusterNetworkCidr ||
   selectClusterNetworkHostPrefix(cluster) !== defaultNetworkSettings.clusterNetworkHostPrefix ||
   selectServiceNetworkCIDR(cluster) !== defaultNetworkSettings.serviceNetworkCidr ||
-  (Boolean(cluster.networkType) && cluster.networkType !== NETWORK_TYPE_OVN);
+  (Boolean(cluster.networkType) && cluster.networkType !== defaultNetworkType);
 
 export const getHostDiscoveryInitialValues = (cluster: Cluster): HostDiscoveryValues => {
   const monitoredOperators = cluster.monitoredOperators || [];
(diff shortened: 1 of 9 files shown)


batzionb:
check about what happens when you uncheck advanced:
according to the code we set to SDN when unchecking advanced
better also check if there's any other places we set SDN by default
src/common/components/clusterConfiguration/NetworkConfiguration.tsx
  const toggleAdvConfiguration = (checked: boolean) => {
    setAdvanced(checked);

    if (!checked) {
      setFieldValue('clusterNetworkCidr', defaultNetworkSettings.clusterNetworkCidr);
      setFieldValue('serviceNetworkCidr', defaultNetworkSettings.serviceNetworkCidr);
      setFieldValue('clusterNetworkHostPrefix', defaultNetworkSettings.clusterNetworkHostPrefix);
      setFieldValue('networkType', 'OpenShiftSDN');
    }
  };

src/common/config/constants.ts
```diff
@@ -216,6 +216,10 @@ export const MS_PER_DAY = 1000 * 60 * 60 * 24;
 
 export const NO_SUBNET_SET = 'NO_SUBNET_SET';
 
+export const NETWORK_TYPE_OVN = 'OVNKubernetes';
+export const NETWORK_TYPE_SDN = 'OpenShiftSDN';
```
jgyselov, line 220: What is the difference between this and using strings in the code?

celdrake: It's generally a better practice to reference values via variables.
This avoids:
- Typing errors
- Having to replace the code in many different place if the values change in the future.  
- It also lets you give "better" names. If the name is confusing, etc, your constant can avoid that confusion.

With the constants, if the values in the Backend change, we would only need to replace it here.

jkilzi: Also, the bundler cannot minify plain strings... Think about how many times the string "OVNKubernetes" (or the other one) would appear in the final bundle.

batzionb:
Are we changing the order of SDN and OVN even though SDN is default for iPV4? 
IPv4 is considered the default use case I think.

batzionb:
Are we changing the order of SDN and OVN even though SDN is default for iPV4? 
IPv4 is considered the default use case I think.

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -70,7 +70,7 @@ export const isAdvNetworkConf = (cluster: Cluster, defaultNetworkSettings: Clust
   selectClusterNetworkCIDR(cluster) !== defaultNetworkSettings.clusterNetworkCidr ||
   selectClusterNetworkHostPrefix(cluster) !== defaultNetworkSettings.clusterNetworkHostPrefix ||
   selectServiceNetworkCIDR(cluster) !== defaultNetworkSettings.serviceNetworkCidr ||
-  (Boolean(cluster.networkType) && cluster.networkType !== 'OpenShiftSDN');
+  (Boolean(cluster.networkType) && cluster.networkType !== NETWORK_TYPE_OVN);
```
batzionb, line 73: this should change to cluster.networkType !== getDefaultNetworkType(args)

batzionb (approved):

PR #1289, opened 2022-04-04
Do not update networking form after autosave

rawagner:
followup for https://issues.redhat.com/browse/MGMT-9873
needs proper testing

### DIFF
--- a/src/common/components/clusterConfiguration/NetworkConfiguration.tsx
+++ b/src/common/components/clusterConfiguration/NetworkConfiguration.tsx
@@ -60,7 +60,7 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
   const isClusterCIDRIPv6 = Address6.isValid(values.clusterNetworkCidr || '');
   const { isIPv6, defaultNetworkType, isSDNSelectable } = React.useMemo(() => {
     const isIPv6 = isSubnetInIPv6({
-      machineNetworkCidr: values.machineNetworkCidr,
+      machineNetworkCidr: cluster.machineNetworkCidr,
       clusterNetworkCidr: values.clusterNetworkCidr,
       serviceNetworkCidr: values.serviceNetworkCidr,
     });
@@ -72,7 +72,7 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
   }, [
     isMultiNodeCluster,
     values.clusterNetworkCidr,
-    values.machineNetworkCidr,
+    cluster.machineNetworkCidr,
     values.serviceNetworkCidr,
   ]);
 
--- a/src/common/types/clusters.ts
+++ b/src/common/types/clusters.ts
@@ -24,12 +24,19 @@ export type HostSubnet = {
   humanized: string;
 };
 export type HostSubnets = HostSubnet[];
-export type NetworkConfigurationValues = V2ClusterUpdateParams & {
-  hostSubnet?: string;
-  useRedHatDnsService?: boolean;
+export type NetworkConfigurationValues = Pick<
+  V2ClusterUpdateParams,
+  | 'clusterNetworkCidr'
+  | 'clusterNetworkHostPrefix'
+  | 'serviceNetworkCidr'
+  | 'apiVip'
+  | 'ingressVip'
+  | 'sshPublicKey'
+  | 'vipDhcpAllocation'
+  | 'networkType'
+> & {
+  hostSubnet: string;
   managedNetworkingType: 'userManaged' | 'clusterManaged';
-  enableProxy: boolean;
-  editProxy: boolean;
 };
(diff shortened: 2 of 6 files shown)


src/common/components/clusterConfiguration/NetworkConfiguration.tsx
```diff
@@ -60,7 +60,7 @@ const NetworkConfiguration: React.FC<NetworkConfigurationProps> = ({
   const isClusterCIDRIPv6 = Address6.isValid(values.clusterNetworkCidr || '');
   const { isIPv6, defaultNetworkType, isSDNSelectable } = React.useMemo(() => {
     const isIPv6 = isSubnetInIPv6({
-      machineNetworkCidr: values.machineNetworkCidr,
```
rawagner, line 63: we never store `machineNetworkCidr` in values. Only place to get it from is `cluster` object

src/common/types/clusters.ts
```diff
@@ -24,9 +24,18 @@ export type HostSubnet = {
   humanized: string;
 };
 export type HostSubnets = HostSubnet[];
-export type NetworkConfigurationValues = V2ClusterUpdateParams & {
```
rawagner, line 27: I opted for more defensive approach. Instead of including all `V2ClusterUpdateParams`, i just pick the ones we really use.

src/common/types/clusters.ts
```diff
@@ -24,9 +24,18 @@ export type HostSubnet = {
...
-  hostSubnet?: string;
-  useRedHatDnsService?: boolean;
```
rawagner, line 29: this seems to be never used on networking as the prop is tied to cluster details. Maybe it is some leftover from the past ?

src/ocm/components/clusterConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -76,25 +76,22 @@ const NetworkConfigurationForm: React.FC<{
...
-      ]);
+
+      const params: V2ClusterUpdateParams = {
+        apiVip: values.apiVip,
+        ingressVip: values.ingressVip,
+        sshPublicKey: values.sshPublicKey,
+        vipDhcpAllocation: values.vipDhcpAllocation,
+        networkType: values.networkType,
+      };
+
```
rawagner, line 94: more defensive approach - just pick the values we know we want.

src/ocm/components/clusterConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -131,7 +128,6 @@ const NetworkConfigurationForm: React.FC<{
 
       const { data } = await ClustersAPI.update(cluster.id, params);
       dispatch(updateClusterBase(data));
-      actions.resetForm({ values: getNetworkInitialValues(data, defaultNetworkSettings) });
```
rawagner, line 134: do not reset formik state. I did not find any use case for this.

src/common/components/clusterConfiguration/NetworkConfiguration.tsx
celdrake: I may be wrong, but couldn't we set  `machineNetworkCidr` in  `getNetworkInitialValues` and then all the fields could be read from `values`? To avoid having this different behaviour on some fields.

celdrake: True, I'll keep that in mind, thanks!

src/ocm/components/clusterConfiguration/NetworkConfigurationForm.tsx
batzionb: We don't need edit_proxy and enable_proxy? we currently send them in the patch request

rawagner: no we dont. We were sending them because of that `omit` usage. But they are not a values we sent to the backend (and its espacially true for OCM which does not use these at all, only ACM)

src/common/types/clusters.ts
```diff
@@ -24,12 +24,19 @@ export type HostSubnet = {
...
+  | 'apiVip'
+  | 'ingressVip'
+  | 'sshPublicKey'
+  | 'vipDhcpAllocation'
+  | 'networkType'
+> & {
+  hostSubnet: string;
   managedNetworkingType: 'userManaged' | 'clusterManaged';
-  enableProxy: boolean;
-  editProxy: boolean;
```
rawagner, line 32: @batzionb thanks for asking the question https://github.com/openshift-assisted/assisted-ui-lib/pull/1289#discussion_r841600296 . I actually removed these ACM-specific props and updated dependent code to make sure they do not leak to common stuff.

batzionb (approved):

PR #1302, opened 2022-04-05
Update NTP sources validation msg

rawagner:


src/cim/components/ClusterDeployment/AdditionalNTPSourcesDialogToggle.tsx
```diff
@@ -1,29 +1,8 @@
...
-          isOpen={isOpen}
-          onClose={() => setOpen(false)}
-          onAdditionalNtpSource={onAdditionalNtpSource}
-        />
-      )}
-    </>
-  );
-};
+export const AdditionalNTPSourcesDialogToggle: React.FC = () => (
+  <>
```
jtomasek, line 4: Please add a TODO that we should not use alert actions to pass additional text to the validation failures from the backend as we do here. The HOST_VALIDATIONS_FAILURE_HINTS object should be used to add additional messages from the UI.

I'll sleep better :)

rawagner: added

jtomasek (approved):

PR #1310, opened 2022-04-08
Add eslint rule to disallow ==

rawagner:
@celdrake PTAL


celdrake (approved):
Nice! Let's keep this approach if we find other `eslint` issues in Code reviews :+1:

PR #1321, opened 2022-04-20
Fix: Do not set network type automatically

jgyselov:
The UI was forcing advanced networking on SNO clusters. There is no scenario in which the `Network type` needs to be automatically set to OVN like this, so the code for that was removed.

### DIFF
--- a/src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
+++ b/src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
@@ -6,7 +6,6 @@ import { Cluster } from '../../../../common/api/types';
 import { useFeatureSupportLevel } from '../../../../common/components/featureSupportLevels';
 import { NetworkConfigurationValues } from '../../../../common/types';
 import { getLimitedFeatureSupportLevels } from '../../../../common/components/featureSupportLevels/utils';
-import { useFeature } from '../../../../common/features';
 import {
   canSelectNetworkTypeSDN,
   getDefaultNetworkType,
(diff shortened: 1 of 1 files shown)


celdrake (dismissed):

celdrake:
@jkilzi I don't think it should be closed. The line that sets the Advanced Network configuration to `true` should be removed, as it's causing the new e2e tests to fail.
Then @jgyselov checked that setting OVN to default in SNO should be handled already, so that's why the rest of the code has been removed as well.

jkilzi:
ok, then reopening the PR :)

jkilzi:
@celdrake [following this comment](https://issues.redhat.com/browse/MGMT-9176?focusedCommentId=19717265&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19717265) the requirement is to have OVN as the default option for every SNO installation. If we are still taking care of this somewhere else I'm fine merging this PR. Unfortunately, I can check it myself right now, but I would appreciate it if someone could take a look.

celdrake:
@jkilzi I checked the branch and tested an SNO and a multi-node cluster.
Both work OK, having their default as they should (OVN for SNO, SDN for multi-cluster), and the "advanced network config" is unchecked as it should. With `master` branch, the SNO would have the "advanced network config" checked initially, even with all settings being the default ones.

The only case I couldn't test is for a cluster using ipV6, as I'm not that familiar with Dual Stack to be able to test it.

jkilzi:
> @jkilzi I checked the branch and tested an SNO and a multi-node cluster. Both work OK, having their default as they should (OVN for SNO, SDN for multi-cluster), and the "advanced network config" is unchecked as it should. With `master` branch, the SNO would have the "advanced network config" checked initially, even with all settings being the default ones.
> 
> The only case I couldn't test is for a cluster using ipV6, as I'm not that familiar with Dual Stack to be able to test it.

10x I'll cover the IPv6 scenario.

jkilzi (approved):

PR #1324, opened 2022-04-25
MGMT-10058: Dual Stack should not be offered when hosts have only IPv4 addresses

jgyselov:
https://issues.redhat.com/browse/MGMT-10058

- Dual-stack will not be offered if the hosts aren't using IPv6.

Screenshot from 2022-04-25 14-36-55

- I've also added filtering to the machine network options (IPv4 for single stack and first subnet in dual-stack, IPv6 for second subnet in dual-stack):

image
image
image

- Added more specific field titles for Cluster and Service networks in dual stack:

image


jkilzi (approved):

jkilzi:
@nirfarkas should we include this in the docs for feature compatibility matrix

PR #1329, opened 2022-04-27
MGMT-8776: Add "Operators" step to the wizard

DanielleChason:
https://issues.redhat.com/browse/MGMT-8776

image

### DIFF
--- a/src/ocm/components/clusterConfiguration/CnvCheckbox.tsx
+++ b/src/ocm/components/clusterConfiguration/CnvCheckbox.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
-import { CheckboxField, Cluster, getFieldId, PopoverIcon } from '../../../common';
+import { CheckboxField, Cluster, getFieldId, CNV_LINK, PopoverIcon } from '../../../common';
 import { useFeatureSupportLevel } from '../../../common/components/featureSupportLevels';
 import { FormGroup, Tooltip } from '@patternfly/react-core';
+import { ExternalLinkAltIcon, HelpIcon } from '@patternfly/react-icons';
 import CNVHostRequirementsContent from '../hosts/CNVHostRequirementsContent';
 
 const CNVLabel: React.FC<{ clusterId: Cluster['id']; isSingleNode?: boolean }> = ({
@@ -28,6 +29,17 @@ export type CnvCheckboxProps = {
   clusterId: string;
 };
 
+const CNVHelperText: React.FC = () => {
+  return (
+    <>
+      Run virtual machines alongside containers on one platform.{' '}
+      <a href={CNV_LINK} target="_blank" rel="noopener noreferrer">
+        {'Learn more'} <ExternalLinkAltIcon />
+      </a>
+    </>
+  );
+};
+
 export const CnvCheckbox: React.FC<CnvCheckboxProps> = ({ openshiftVersion, isSNO, clusterId }) => {
   const featureSupportLevelContext = useFeatureSupportLevel();
   const name = 'useContainerNativeVirtualization';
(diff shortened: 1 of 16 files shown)


nirfarkas:
LGTM
/approve

src/ocm/components/clusterConfiguration/CnvCheckbox.tsx
```diff
@@ -31,6 +31,17 @@ export type CnvCheckboxProps = {
   clusterId: string;
 };
 
+const CNVHelperText: React.FC = () => {
```
celdrake, line 34: We don't need to define the components as `React.FC` anymore, just as plain functions.
```suggestion
const CNVHelperText = () => {
```

src/ocm/components/clusterConfiguration/ODFCheckbox.tsx
```diff
@@ -25,6 +31,17 @@ export type ODFCheckboxProps = {
   openshiftVersion?: string;
 };
 
+const ODFHelperText: React.FC = () => {
```
celdrake, line 34:
```suggestion
const ODFHelperText = () => {
```

src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -91,6 +91,19 @@ const hostDiscoveryStepValidationsMap: WizardStepValidationMap = {
   softValidationIds: [],
 };
 
+const operatorsStepValidationsMap: WizardStepValidationMap = {
```
celdrake, line 94: There is already an empty WizardStepValidationMap. I would define this as a constant to avoid repetition etc.
And just to avoid issues in case this data is mutated somehwere, I would create a function that creates this object.

Eg.
```
const buildEmptyValidationMap = () => return object;
const staticIpValidationsMap: WizardStepValidationMap = buildEmptyValidationMap();
const operatorsStepValidationsMap: WizardStepValidationMap = buildEmptyValidationMap();
```

src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -33,7 +34,7 @@ export const getClusterWizardFirstStep = (
     case 'pending-for-input':
     case 'adding-hosts':
     case 'insufficient':
-      return 'host-discovery';
+      return 'operators';
```
celdrake, line 37: I'm not very familiar with these operators, but just to confirm this is what we want.

If the cluster is insufficient, we will move to Operators and not to host discovery?
And also, I can see later that in Host Discovery we don't wait for `sufficient-masters-count` anymore. So users can access networking even though not all hosts were discovered?

celdrake (changes requested):
IIRC, these operators were in the "Host discovery" stage before.
Should they not be removed from `HostInventory` as well as related places, eg `HostDiscoveryService`, type definitions etc?

src/ocm/services/OperatorsService.tsx
```diff
@@ -0,0 +1,40 @@
...
+  OPERATOR_NAME_OCS,
+  OPERATOR_NAME_LSO,
+  HostDiscoveryValues,
+  Cluster,
+} from '../../common';
+import { getOlmOperatorCreateParamsByName } from '../components/clusters/utils';
+const OperatorsService = {
+  setOLMOperators(
+    params: V2ClusterUpdateParams,
+    values: HostDiscoveryValues,
```
celdrake, line 14: Thanks for the clean-up. However, I see a few places that are missing:

- OperatorService should use `OperatorsValues` and not `HostDiscoveryValues`
- `HostDiscoveryValues` should not include  `useExtraDisksForLocalStorage` or `useContainerNativeVirtualization`.
- The call to `setOLMOperators` in `Operators` needs to be updated with the last parameter, and it needs to be removed from `HostDiscovery` 
- `getHostDiscoveryInitialValues` should not include the moved fields.

DanielleChason: I believe I did some of those things in the past, but I lost it on one of the rebasing, I guess
I'll do it now

src/ocm/components/clusterWizard/wizardTransition.ts
DanielleChason: If I remember correctly, that's what we agreed. 
 @jkilzi  @nirfarkas  What do you think? To what step should we go if insufficient? 
And the sufficient-masters-count it's still a validation in host discovery.

celdrake:
I'll fix the merge conflicts and create a new PR

PR #1331, opened 2022-04-27
Updates swagger.yaml

DanielleChason:
I updated the image with the latest version from the BE. 
It changed our _types.ts_ file and therefore I fixed some of the files.  

@jkilzi 
In the file _hardwareInfo.ts_ they use the field:  **inventory.cpu?.clockMegahertz** as an one of the possibles parameters of a function, but from what I understood this filed doesn't exist anymore so I removed it, but I'm not sure this is the right thing to do. 
 Screenshot of the change:
 
image

### DIFF
--- a/src/common/api/types.ts
+++ b/src/common/api/types.ts
@@ -321,6 +321,10 @@ export interface ClusterCreateParams {
    */
   serviceNetworkCidr?: string; // ^(?:(?:(?:[0-9]{1,3}\.){3}[0-9]{1,3}\/(?:(?:[0-9])|(?:[1-2][0-9])|(?:3[0-2])))|(?:(?:[0-9a-fA-F]*:[0-9a-fA-F]*){2,})/(?:(?:[0-9])|(?:[1-9][0-9])|(?:1[0-1][0-9])|(?:12[0-8])))$
   /**
+   * The virtual IP used to reach the OpenShift cluster's API.
+   */
+  apiVip?: string; // ^(?:(?:(?:[0-9]{1,3}\.){3}[0-9]{1,3})|(?:(?:[0-9a-fA-F]*:[0-9a-fA-F]*){2,}))?$
+  /**
    * The virtual IP used for cluster ingress traffic.
    */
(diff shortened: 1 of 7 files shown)


src/common/api/types.ts
```diff
@@ -570,7 +575,6 @@ export interface Cpu {
   flags?: string[];
   modelName?: string;
   architecture?: string;
-  clockMegahertz?: number;
```
celdrake, line 573: Quite recently there was this commit that fixed a bug for ACM, so I think the field `clockMegahertz` is necessary (as well as keeping the usage in `getHumanizedCpuClockSpeed`)

https://github.com/openshift-assisted/assisted-ui-lib/commit/c50a119e67b030c6a4188e3fedbf3ed6a111cafb

jkilzi: @celdrake this removal comes from a change in the swagger.yaml. The BE removed this field without being aware if we were consuming it. 
cc @osherdp

osherdp: Seems like it has been changed with some kind of forward-compatibility handling: https://github.com/openshift/assisted-service/blob/master/api/v1beta1/agent_types.go#L101

Meaning you checked first for ``frequency`` and if it doesn't then using ``clockMegahertz``
So I think there was some kind of awareness about this field name migration

jkilzi: 10x @osherdp so I guess we should use just `frequency` now. Is this correct?

osherdp: @nmagnezi can you confirm? ^
I know it's an issue from more than a year ago, but should they use ``frequency`` instead of ``clockMegahertz`` as written [here](https://github.com/openshift/assisted-service/blob/master/api/v1beta1/agent_types.go#L101)?

nmagnezi: If you are asking about REST-API, yes.
The KubeAPI agent controller also does read it from `frequency`.

https://github.com/openshift/assisted-service/blob/ddb6400f03c15526c39fa6a2b1bc1f1b7e6def92/internal/controller/controllers/agent_controller.go#L900-L908

jkilzi (approved):

PR #1337, opened 2022-04-28
MGMT-10171: Form fields get invalidated immediately before entering information

batzionb:
https://issues.redhat.com/browse/MGMT-10171

Right now we show errors on an input field if it was touched.
This causes required array fields new items to become invalid right away.
Instead I propose the following:
If the value is empty - show the error only if it was previously none empty
If the value is not empty - show the error 
This change effects all input fields across the lib

### DIFF
--- a/src/common/hooks/useFieldErrorMsg.tsx
+++ b/src/common/hooks/useFieldErrorMsg.tsx
@@ -13,7 +13,7 @@ const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undef
     for none empty values - always show the error if there is one
   */
   const [field, { error }] = useField({
-    name: name,
+    name,
     validate,
   });
   const [hadValue, setHadValue] = React.useState<boolean>(false);
--- a/src/common/hooks/useFieldErrorMsg.tsx
+++ b/src/common/hooks/useFieldErrorMsg.tsx
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import { FieldValidator, useField } from 'formik';
+
+type Props = {
+  name: string;
+  inputError?: string;
+  validate?: FieldValidator;
+};
+
+const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undefined => {
+  /*
+    for empty values - show an error only if at a certain point it was none empty, to avoid new items in a field array shown as invalid right away
+    for none empty values - always show the error if there is one
+  */
+  const [field, { error }] = useField({
+    name: name,
+    validate,
+  });
+  const [hadValue, setHadValue] = React.useState<boolean>(false);
+  React.useEffect(() => {
+    if (field.value) {
+      setHadValue(field.value);
+    }
+  }, [field.value, setHadValue]);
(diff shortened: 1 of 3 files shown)


src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -0,0 +1,30 @@
...
+  validate?: FieldValidator;
+};
+
+const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undefined => {
+  /*
+    for empty values - show an error only if at a certain point it was none empty, to avoid new items in a field array shown as invalid right away
+    for none empty values - always show the error if there is one
+  */
+  const [field, { error }] = useField({
+    name: name,
```
jkilzi, line 16:
```suggestion
    name,
```

src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { FieldValidator, useField } from 'formik';
+
+type Props = {
+  name: string;
+  inputError?: string;
+  validate?: FieldValidator;
+};
+
+const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undefined => {
```
jkilzi, line 10: Am I missing something? Seems like `validate` is not in use here
```suggestion
const useFieldErrorMsg = ({ name, inputError }: Props): string | undefined => {
```

src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -0,0 +1,30 @@
...
+    validate,
+  });
+  const [hadValue, setHadValue] = React.useState<boolean>(false);
+  React.useEffect(() => {
+    if (field.value) {
+      setHadValue(field.value);
+    }
+  }, [field.value, setHadValue]);
+  const showError = field.value || (!field.value && hadValue);
+  const errorMessage = (showError ? error : '') || inputError;
```
jkilzi, line 26: Can `error` be undefined? The same for `inputError`

jkilzi (changes requested):

src/common/hooks/useFieldErrorMsg.tsx
batzionb: validate is passed to useField

src/common/hooks/useFieldErrorMsg.tsx
batzionb: yes they can both be undefined

src/common/hooks/useFieldErrorMsg.tsx
jkilzi: oops... i overlooked that :) 10x

src/common/hooks/useFieldErrorMsg.tsx
jkilzi: ok, so I guess is up to the caller to protect against this case

jkilzi (approved):

PR #1343, opened 2022-05-02
MGMT-7527: Deprecate old cluster networking fields in the UI

jgyselov:
https://issues.redhat.com/browse/MGMT-7527

### DIFF
--- a/src/ocm/components/clusterConfiguration/networkConfiguration/VirtualIPControlGroup.tsx
+++ b/src/ocm/components/clusterConfiguration/networkConfiguration/VirtualIPControlGroup.tsx
@@ -7,6 +7,7 @@ import { HostSubnets, NetworkConfigurationValues, ValidationsInfo } from '../../
 import { CheckboxField, FormikStaticField, InputField } from '../../../../common/components/ui';
 import { FeatureSupportLevelBadge } from '../../../../common/components';
 import { NETWORK_TYPE_SDN, NO_SUBNET_SET } from '../../../../common/config/constants';
+import { selectMachineNetworkCIDR } from '../../../../common';
 
 interface VipStaticValueProps {
   vipName: string;
@@ -15,7 +16,8 @@ interface VipStaticValueProps {
 }
 
 const VipStaticValue = ({ vipName, cluster, validationErrorMessage }: VipStaticValueProps) => {
-  const { vipDhcpAllocation, machineNetworkCidr } = cluster;
+  const { vipDhcpAllocation } = cluster;
+  const machineNetworkCidr = selectMachineNetworkCIDR(cluster);
 
   if (vipDhcpAllocation && cluster[vipName]) {
     return cluster[vipName];


jgyselov:
```
export const selectMachineNetworkCIDR = ({
  machineNetworks,
  machineNetworkCidr,
}: Partial<Cluster>) => head(machineNetworks)?.cidr ?? machineNetworkCidr;
export const selectClusterNetworkCIDR = ({
  clusterNetworks,
  clusterNetworkCidr,
}: Partial<Cluster>) => head(clusterNetworks)?.cidr ?? clusterNetworkCidr;
export const selectClusterNetworkHostPrefix = ({
  clusterNetworks,
  clusterNetworkHostPrefix,
}: Partial<Cluster>) => head(clusterNetworks)?.hostPrefix ?? clusterNetworkHostPrefix;
export const selectServiceNetworkCIDR = ({
  serviceNetworks,
  serviceNetworkCidr,
}: Partial<Cluster>) => head(serviceNetworks)?.cidr ?? serviceNetworkCidr;
```
Throughout the code we use these functions from [clusterSelectors.ts](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/selectors/clusterSelectors.ts#L8-L23) in /common. Can we keep using these functions and rely on the BE always sending the `machineNetworks`, `serviceNetworks`, and `clusterNetworks` arrays correctly or would creating new functions in /ocm be a better approach?

jkilzi:
> ```
> export const selectMachineNetworkCIDR = ({
>   machineNetworks,
>   machineNetworkCidr,
> }: Partial<Cluster>) => head(machineNetworks)?.cidr ?? machineNetworkCidr;
> export const selectClusterNetworkCIDR = ({
>   clusterNetworks,
>   clusterNetworkCidr,
> }: Partial<Cluster>) => head(clusterNetworks)?.cidr ?? clusterNetworkCidr;
> export const selectClusterNetworkHostPrefix = ({
>   clusterNetworks,
>   clusterNetworkHostPrefix,
> }: Partial<Cluster>) => head(clusterNetworks)?.hostPrefix ?? clusterNetworkHostPrefix;
> export const selectServiceNetworkCIDR = ({
>   serviceNetworks,
>   serviceNetworkCidr,
> }: Partial<Cluster>) => head(serviceNetworks)?.cidr ?? serviceNetworkCidr;
> ```
> 
> Throughout the code we use these functions from [clusterSelectors.ts](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/selectors/clusterSelectors.ts#L8-L23) in /common. Can we keep using these functions and rely on the BE always sending the `machineNetworks`, `serviceNetworks`, and `clusterNetworks` arrays correctly or would creating new functions in /ocm be a better approach?

@osherdp?

jkilzi:
@jgyselov Let's merge this PR, since the default place we use for looking up for the *NetworkCidrs is in the new fields.
We can open an independent RFE later to completely remove the old fields.

jkilzi (approved):

mkowalski:
Folks, be aware, `*NetworkCidr` fields are now **empty**. They may be returned in the payload, but you must never ever rely on their values. `*Networks` are the only ones to be used in both GET and POST.

POST action against `*NetworkCidr` will not cause any change to the Cluster object.
GET action against `*NetworkCidr` will return an empty value even if the respective `*Networks` contains a proper value.

jkilzi:
Yes Mat you're right, I meant the *Network fields. I've edited my previous comment. Thanks for pointing that out.

PR #1345, opened 2022-05-03
MGMT-10212: Avoid false positives checking version features

celdrake:
https://issues.redhat.com/browse/MGMT-10212

Avoid partial version matches resulting in wrong support data identification.

Eg. If we received versions `4.1` and `4.10`, the previous code would match to version `4.1`.
The regular expression would allow only `4.10`, `4.10.`, `4.10.123` etc.


jkilzi (approved):
@celdrake please MGMT-XXXX

celdrake:
> @celdrake please MGMT-XXXX

https://issues.redhat.com/browse/MGMT-10212 created

PR #1347, opened 2022-05-03
MGMT-9454: Allow transforming day1 to day2 in on-prem environment

jgyselov:
https://issues.redhat.com/browse/MGMT-9454

**Installation page**:
- Added "Add hosts" button that transforms the cluster to AddHosts kind. (Only visible if cluster is multinode, installed and environment is in on-prem.)

Screenshot from 2022-05-03 13-29-08
Screenshot from 2022-05-03 13-40-13

**Add hosts page**:
- Shows the same console information like the Installation page.
- Shows all hosts in the cluster (including the day1 hosts)
- Allows to download ISO and install ready hosts

Screenshot from 2022-05-03 13-58-44
Screenshot from 2022-05-03 13-55-59

---

Note: I couldn't successfully install a day2 host because of failing validations.

### DIFF
--- a/src/ocm/components/clusterDetail/ClusterDetail.tsx
+++ b/src/ocm/components/clusterDetail/ClusterDetail.tsx
@@ -17,6 +17,7 @@ import {
   useAlerts,
   getEnabledHosts,
   getOlmOperators,
+  isSNO,
 } from '../../../common';
 import { Cluster } from '../../../common/api/types';
 import ClusterHostsTable from '../hosts/ClusterHostsTable';
@@ -35,6 +36,9 @@ import ClusterProgressItems from '../../../common/components/clusterDetail/Clust
 import { EventsModalButton } from '../../../common/components/ui/eventsModal';
 import { onFetchEvents } from '../fetching/fetchEvents';
 import { getClusterProgressAlerts } from './getProgressBarAlerts';
+import { ClustersAPI } from '../../services/apis';
+import { updateCluster } from '../../reducers/clusters';
+import { handleApiError, ocmClient } from '../../api';
 
 type ClusterDetailProps = {
   cluster: Cluster;
@@ -46,6 +50,7 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
   const clusterVarieties = useClusterStatusVarieties(cluster);
   const { credentials, credentialsError } = clusterVarieties;
   const { monitoredOperators = [] } = cluster;
+  const canAddHosts = !ocmClient && !isSNO(cluster) && cluster.status === 'installed';
 
   return (
     <Stack hasGutter>
@@ -114,6 +119,25 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
               consoleUrl={credentials?.consoleUrl}
               id={getClusterDetailId('button-launch-console')}
             />
+          )}
+          {canAddHosts && (
+            <ToolbarButton
+              variant={ButtonVariant.primary}
+              component={(props) => (
+                <Link to={`${routeBasePath}/clusters/${cluster.id}`} {...props} />
+              )}
+              id={getClusterDetailId('button-add-hosts')}
+              onClick={async () => {
+                try {
+                  const { data } = await ClustersAPI.allowAddHosts(cluster.id);
+                  updateCluster(data);
+                } catch (e) {
+                  handleApiError(e);
+                }
+              }}
+            >
+              Add hosts
+            </ToolbarButton>
           )}
           {!isSingleClusterMode() && (
             <ToolbarButton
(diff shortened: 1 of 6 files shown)


jgyselov:
@nirfarkas PTAL

src/ocm/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -124,6 +129,26 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
...
+                component={(props) => (
+                  <Link to={`${routeBasePath}/clusters/${cluster.id}`} {...props} />
+                )}
+                id={getClusterDetailId('button-add-hosts')}
+                onClick={async () => {
+                  try {
+                    const { data } = await ClustersAPI.addHosts(cluster.id);
+                    updateCluster(data);
+                  } catch (e) {
+                    console.log(e);
```
celdrake, line 144: We could remove the `console.log`

src/ocm/components/clusterDetail/ClusterDetailStatusVarieties.tsx
```diff
@@ -21,20 +21,23 @@ type ClusterStatusVarieties = {
   fetchCredentials: () => void;
 };
 
-export const useClusterStatusVarieties = (cluster: Cluster): ClusterStatusVarieties => {
+export const useClusterStatusVarieties = (cluster?: Cluster): ClusterStatusVarieties => {
```
celdrake, line 24: Why is it that we can get here without a cluster? I think in that case , the hook can't do anything so IMO it should still be mandatory.

Then if you are in a place where there's no cluster, conditionally avoid rendering the component using the hook, etc.

src/ocm/services/apis/ClustersAPI.ts
```diff
@@ -156,6 +156,12 @@ const ClustersAPI = {
     );
   },
 
+  addHosts(clusterId: Cluster['id']) {
```
celdrake, line 159: I think a better name would be `allowAddHosts` since it's not actually adding any host.

celdrake (changes requested):

src/ocm/components/clusterDetail/ClusterDetailStatusVarieties.tsx
jgyselov: Because in `AddHosts` (where we call the `useClusterStatusVarieties`) we get the cluster from `AddHostsContext` which starts empty. I cannot prevent calling the hook if the cluster is `undefined`, which shouldn't happen but Typescript will think it can.

jgyselov: If I made `cluster` mandatory in AddHostsContext, I'd have to define it [here](https://github.com/openshift-assisted/assisted-ui-lib/blob/7520c65f81c3d5ac4540eae437ef883551a09a07/src/common/components/AddHosts/AddHostsContext.tsx#L10). I'm not sure how I could do that.

I've tried to change it so that it only triggers if `cluster.id` or `cluster.status` change (since these are the only fieldfs we read in there.

celdrake (dismissed):
Looks good!

src/ocm/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -46,6 +50,7 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
   const clusterVarieties = useClusterStatusVarieties(cluster);
   const { credentials, credentialsError } = clusterVarieties;
   const { monitoredOperators = [] } = cluster;
+  const canAddHosts = isSingleClusterMode() && !isSNO(cluster) && cluster.status === 'installed';
```
jkilzi, line 53: @jgyselov why this should apply only if the package was built using single cluster mode? (that mode is quite deprecated)

jgyselov: Did I completely misunderstand what this feature was for? I.e. the on-prem env with only one cluster to which we want to be able to add hosts after installation?

jkilzi:
@nirfarkas the attached picture contains 2 CTAs, is that OK?

src/ocm/reducers/clusters/clustersSlice.ts
```diff
@@ -3,13 +3,15 @@ import { Cluster } from '../../../common';
...
+import { ocmClient } from '../../api';
 
 export const fetchClustersAsync = createAsyncThunk<Cluster[] | void>(
   'clusters/fetchClustersAsync',
   async () => {
     try {
       const { data } = await ClustersAPI.list();
-      return data.filter((cluster) => cluster.kind === 'Cluster');
+      const isOcm = !!ocmClient;
+      return data.filter((cluster) => (isOcm ? cluster.kind === 'Cluster' : true));
```
jkilzi, line 14: @jgyselov why do we need this?

jgyselov: Because otherwise a cluster transformed to day2 (`cluster.kind = "AddHostsCluster"`) would disappear from the clusters list. I'd say disappearing clusters is undesired behaviour. @jkilzi

jkilzi (approved):

celdrake (approved):

PR #1348, opened 2022-05-03
MGMT-9477: Link to day2 cluster is irrelevant for non cloud AI

jgyselov:
https://issues.redhat.com/browse/MGMT-9477

Removed from Cluster detail card in OCM:
image

Still showing when viewed through assisted installer (unless it's on-prem single cluster mode):
image

### DIFF
--- a/src/ocm/components/clusterDetail/ClusterDetailStatusMessages.tsx
+++ b/src/ocm/components/clusterDetail/ClusterDetailStatusMessages.tsx
@@ -14,15 +14,22 @@ import { VSPHERE_CONFIG_LINK } from '../../../common/config/constants';
 import { isSNO } from '../../../common/selectors/clusterSelectors';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { calculateClusterDateDiff } from '../../../common/sevices/DateAndTime';
+import { isSingleClusterMode } from '../../config';
 
 type ClusterDetailStatusMessagesProps = {
   cluster: Cluster;
+  showAddHostsInfo?: boolean;
 };
 
-const ClusterDetailStatusMessages: React.FC<ClusterDetailStatusMessagesProps> = ({ cluster }) => {
+const ClusterDetailStatusMessages: React.FC<ClusterDetailStatusMessagesProps> = ({
+  cluster,
+  showAddHostsInfo = true,
+}) => {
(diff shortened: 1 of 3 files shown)


celdrake (approved):

src/ocm/components/clusterDetail/ClusterDetailStatusMessages.tsx
```diff
@@ -14,15 +14,22 @@ import { VSPHERE_CONFIG_LINK } from '../../../common/config/constants';
...
-const ClusterDetailStatusMessages: React.FC<ClusterDetailStatusMessagesProps> = ({ cluster }) => {
+const ClusterDetailStatusMessages: React.FC<ClusterDetailStatusMessagesProps> = ({
+  cluster,
+  showAddHostsInfo = true,
+}) => {
   const { inactiveDeletionHours } = useDefaultConfiguration(['inactiveDeletionHours']);
   const inactiveDeletionDays = Math.round((inactiveDeletionHours || 0) / 24);
   const dateDifference = calculateClusterDateDiff(inactiveDeletionDays, cluster.installCompletedAt);
+  const showAddHostsAlert =
+    showAddHostsInfo && !isSingleClusterMode() && cluster.status === 'installed' && !isSNO(cluster);
```
jkilzi, line 32: Single cluster refers to a UI mode (probably not relevant anymore at this point, but worth double-checking that) where you are allowed to install only one cluster, there is no clusters list but instead, you are supposed to be taken directly to the discovery step. Even if you have a single cluster, possibly a multi-node one, adding hosts to such a cluster is something that users should be able to do.

```suggestion
    showAddHostsInfo && cluster.status === 'installed' && !isSNO(cluster);
```

jkilzi (changes requested):

src/ocm/components/clusterDetail/ClusterDetailStatusMessages.tsx
jgyselov: But this alert is not relevant for single cluster mode. It points specifically to console.

For adding hosts in single cluster mode, we have a button in the *bottom action bar*. If we want an alert to tell the users they can add hosts in single cluster mode, it will need different content.
@jkilzi

jkilzi: Sorry, you are right :)

jkilzi (approved):

PR #1366, opened 2022-05-10
NO-ISSUE: Removes react-scripts

jkilzi:
## Motivation
We have been using `react-scripts` because it has been so long a convenient way for not having to set up eslint, prettier, and jest independently. The drawback with this lib is that drags with it a lot of 3rd party dependencies that have CVEs opened against them constantly. In order to mitigate these issues ideally, we would update react-scripts but it is not always possible, so we need to consider if every indirect CVE that is being opened against us is worth handling or not. The situation right now is that we can manage without it, and it will allow us to drop the number of issues reported by `yarn audit` **from 97 to 3 issues!**

## What has changed?
- `eslint` and its plugin ecosystem had been bumped to the latest version
- `prettier` (same)
- `typescript` bumped to v3.9
- We can now lint the files using type-checking information, no need to execute `tsc --noEmit` in a separate step.
- Code formatting is executed through `eslint`, do not run `prettier` since it runs automatically as an eslint rule  

## Tradeoffs
- `@typescrtipt-eslint` introduced new rules, most of them had been disabled in order to lower the number the files that had been affected by this change  
  - I recommend enabling them gradually in order to improve the code quality
- Jest and the test files are gone, they were never used in a practical manner and we use other means for testing our code now. 

## IDE integrations
I've tested the repo in order to see if I still get visual feedback with vscode and webstorm, and couldn't find any issues.

## Addresses:
* https://bugzilla.redhat.com/show_bug.cgi?id=2055585
* and probably other bugs that we are still not aware of them


package.json
```diff
@@ -38,14 +38,12 @@
...
-    "test:type-check": "tsc --noEmit",
-    "test:lint": "eslint --ext .js,.jsx,.ts,.tsx --color --max-warnings=0 ./src",
-    "test:unit": "cross-env CI=1 react-scripts test --env=jsdom",
+    "test": "run-s test:circular lint",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 src/index.ts",
-    "test:watch": "react-scripts test --env=jsdom",
-    "lint": "eslint --ext .js,.jsx,.ts,.tsx --color --fix ./src",
+    "lint": "eslint ./src --ext .js,.jsx,.ts,.tsx --color",
+    "format": "yarn run lint --fix",
     "prettier": "prettier --write '**/*.{js,jsx,ts,tsx,json,md,scss,yaml,yml}'",
```
jtomasek, line 45: Should we remove this in favor of `format` script above?

.eslintrc
```diff
@@ -1,46 +0,0 @@
...
-  },
-  "plugins": ["prettier", "react", "react-hooks", "@typescript-eslint"],
-  "extends": [
-    "eslint:recommended",
-    "plugin:cypress/recommended",
-    "plugin:@typescript-eslint/recommended",
-    "plugin:react/recommended",
-    "plugin:prettier/recommended",
-    "prettier/@typescript-eslint",
-    "prettier/react"
```
jtomasek, line 27: IIUC these are from eslint-config-prettier and ensure that there are no collisions between eslint and prettier rules. Shouldn't we keep these?

https://prettier.io/docs/en/integrating-with-linters.html

.eslintrc.js
```diff
@@ -0,0 +1,61 @@
...
+    ecmaVersion: 2017,
+    sourceType: 'module',
+  },
+  settings: {
+    react: {
+      version: 'detect',
+    },
+  },
+  plugins: ['@typescript-eslint', 'react', 'react-hooks'],
+  extends: ['eslint:recommended', 'plugin:react/recommended', 'plugin:prettier/recommended'],
```
jkilzi, line 23: @jtomasek the last plugins covers the integration with prettier

```suggestion
  extends: ['eslint:recommended', 'plugin:react/recommended', 'plugin:prettier/recommended'],
```

rawagner (approved):
Looks great, glad to see react-scripts removed. Great job @jkilzi !

package.json
jkilzi: The reason I separated them is that in CI you shouldn't execute `eslint ... --fix`
Also, please notice I dropped the `--max-warnings=0` flag.
There are 2 reasons for this: 
1. In practice, we ended up disabling rules in order to silence the linter instead of dealing with the problematic code.
2. Now we'll see the warnings and we can work iteratively in order to resolve them.

src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -140,7 +142,8 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({
           alerts
             .filter((alert) => alert.title === 'Failed to set role')
             .forEach((a) => removeAlert(a.key));
-        } catch (error) {
+        } catch (e) {
+          const error = e as Parameters<typeof handleApiError>[0];
```
batzionb, line 146: In static IP branch I fixed the handleApiError to get unknown instead of AxiosError.
It's correct for typesscript to complain about this, when the error is a string for example, not directly an API error, this error handling fails. So after the static ip merge we won't need this casting.

jkilzi: @batzionb please post a snippet of how you're fixing it in your PR and I'll commit the same changes to mine

batzionb: Look at the file  
src/ocm/api/utils.ts
and src/ocm/api/types.ts
in the [static ip PR](https://github.com/openshift-assisted/assisted-ui-lib/pull/1328/files)

image

With this change, when the functions handleApiError and getErrorMessage are used there are no TS errors

batzionb (approved):

batzionb (dismissed):

celdrake (approved):

batzionb (approved):

PR #1372, opened 2022-05-16
MGMT-9638 Replaced api-vip-connected host validation for ignition-dow…

celdrake:
…nloadble


jkilzi:
@rawagner this change is in a shared component, the `api-vip-connected` host validation was renamed to `ignition-downloadable`. Can you PTAL, I'm wondering if it does conflict with CIM...

celdrake:
> @rawagner this change is in a shared component, the `api-vip-connected` host validation was renamed to `ignition-downloadable`. Can you PTAL, I'm wondering if it does conflict with CIM...

You can see a few lines before this change, in https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/components/hosts/HostValidationGroups.tsx#L133 that the map uses the key `ignition-downloadable` whereas the alert uses the `api-vip-connected` value, so it would never be displayed.

So in theory this change should not conflict, it would only fix a bug of the value mismatch.

celdrake:
The Jira ticket had the wrong validation key `ignition-downladable` , I fixed it to ` ignition-downloadable`.

jkilzi (approved):

PR #1374, opened 2022-05-16
MGMT-10377: Progress bar's notifications are now in the same size of the bar

DanielleChason:
https://issues.redhat.com/browse/MGMT-10377

**now:**
image

**before:**
image


jkilzi:
I believe the table should also be shrinked? @nirfarkas wdyt?

jkilzi (approved):

jkilzi:
> I believe the table should also be shrinked? @nirfarkas wdyt?

Following Nir's reply, the answer is: no.

celdrake (approved):

PR #1377, opened 2022-05-17
MGMT-10402 Show media-connected warning in Discovery step

celdrake:
https://issues.redhat.com/browse/MGMT-10402

The hosts validationInfo of type `media-connected` was not being displayed in the Host Discovery phase. For the next steps, the message was being shown because they show all failed validations.

After the fix:
media-connected-ok
.


celdrake:
@jtomasek The same code can be found in the CIM component, so please take a look if it affects CIM as well.

jkilzi (approved):

PR #1387, opened 2022-05-19
MGMT-10185: UI not blocking "localhost" name changing

jgyselov:
https://issues.redhat.com/browse/MGMT-10185

I also specified that characters need to be lowercase, since uppercase letters are not allowed.

Before:
image

After:
image

### DIFF
--- a/src/common/components/ui/formik/validationSchemas.ts
+++ b/src/common/components/ui/formik/validationSchemas.ts
@@ -11,13 +11,13 @@ import { ClusterNetwork, MachineNetwork, ServiceNetwork } from '../../../api/typ
 import { allSubnetsIPv4 } from '../../../selectors';
 
 const ALPHANUMBERIC_REGEX = /^[a-zA-Z0-9]+$/;
-const CLUSTER_NAME_REGEX = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/;
+const NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
+const NAME_CHARS_REGEX = /^[a-z0-9-.]*$/;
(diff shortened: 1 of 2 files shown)


rawagner:
@jgyselov you sure about the lowercase requirement ? I think when we tried it, it worked fine IIRC

rawagner:
I think the DNS resolution is case insensitive

jgyselov:
Maybe it's a new BE requirement? It definitely doesn't let me put uppercase characters in the hostname. @rawagner 
image

rawagner:
you're right. I just checked on k8s API and it is the same.

celdrake:
@jgyselov Does it only affect the hosts being renamed? Or I three hosts are discovered with `localhost` name, would it be a problem?

jgyselov:
> Does it only affect the hosts being renamed? Or I three hosts are discovered with `localhost` name, would it be a problem?

No host can be named `localhost`. If it is, it's automatically `insufficient` and the cluster cannot be installed until those hosts are renamed. @celdrake

celdrake:
> 

What I mean is that when you are renaming 1 host, the other 2 will still have the invalid name "localhost". 
Just to confirm, that the validation does not block renaming the hosts individually etc.

jgyselov:
> What I mean is that when you are renaming 1 host, the other 2 will still have the invalid name "localhost". Just to confirm, that the validation does not block renaming the hosts individually etc.

I don't think I specifically tried this scenario where all host are name `localhost` at the beginning. But the validations are always just checking the potential new name against these conditions so it shouldn't block anything.

jkilzi (approved):

jkilzi (approved):

PR #1391, opened 2022-05-19
MGMT-10226 Show ignition-downloadable message

celdrake:
https://issues.redhat.com/browse/MGMT-10226

There were 2 reasons why it wasn't showing:
- The new signal "ignition-downloable" was not listed in the Host discovery step.
- The host status in Host Discovery did not have the needed component.

I simulated it with the e2e, as unfortunately I don't have a Day 2 cluster with the issue.
ignition-downloadable-message


jtomasek:
Reverting. `ignition-downloadable` validation is used only on day2 so it should not be required by `hosts-discovery` wizard step. Adding it there does not fix the day2 problem of showing all validations.

PR #1401, opened 2022-05-26
MGMT-10485: Static IP shows wrong view after backend failing when filling in wrong ipv6 address

batzionb:
https://issues.redhat.com/browse/MGMT-10485
Managed to reproduce the issue with the following steps:
1. create static IP cluster
2. in form view fill in network wide configuration steps, dual stack 
3. fill in invalid address in ipv6 fieldv - notice that data isn't save because it isn't valid
4. **switch to yaml** view and don't fill in data, thus wizard steps change without persistence of yaml view
5. click on cluster details
6. click on next
issue is the steps in the wizard don't correlate with the current status of the infra env, and we get an invalid UI state


jkilzi:
@batzionb please port this PR to releases/v2.4

PR #1406, opened 2022-05-30
MGMT-10392 Fix broken comparison due to extra array

celdrake:
https://issues.redhat.com/browse/MGMT-10392
There was a bug in the previous fix https://github.com/openshift-assisted/assisted-ui-lib/pull/1389

The comparison was not correct because of an extra array in the object.
Now the comparison is fixed, the PATCH should not happen if no changes exist.


jkilzi:
@celdrake please port it to releases/v2.4

jkilzi (approved):

PR #1407, opened 2022-05-31
MGMT-10478: Static IP UI IPv4 and IPv6 addresses allow invalid address

DanielleChason:
https://issues.redhat.com/browse/MGMT-10478

image

image


DanielleChason:
Now the validation works for different ipv6 formatting: 
For example:
image

and:
image

src/ocm/components/clusterConfiguration/staticIp/commonValidationSchemas.tsx
```diff
@@ -50,6 +60,65 @@ export const getIpAddressValidationSchema = (protocolVersion: 'ipv4' | 'ipv6') =
...
+          if (compareIPV6Addresses(new Address6(CATCH_ALL_IP.ipv6), new Address6(value))) {
+            return false;
+          }
+        } else {
+          if (value === CATCH_ALL_IP.ipv4) {
+            return false;
+          }
+        }
+      } catch (e) {
+        return false;
```
batzionb, line 115: you should return true if you catch an error because it means it didn't manage to parse the ip address and doesn't have to do with this validation, a different one should handle this.
I know in other examples it isn't handled this way but I believe this is the correct approach and it's the way getIpAddressInSubnetValidationSchema handles this. It has a comment there about why we return true in this case

src/ocm/components/clusterConfiguration/staticIp/commonValidationSchemas.tsx
```diff
@@ -50,6 +60,65 @@ export const getIpAddressValidationSchema = (protocolVersion: 'ipv4' | 'ipv6') =
...
 };
 
+export const compareIPV6Addresses = (address1: Address6, address2: Address6) => {
+  if (JSON.stringify(address1.toByteArray()) === JSON.stringify(address2.toByteArray())) {
+    return true;
+  }
+  return false;
+};
+
+export const isNotLocalHostIPAddress = (protocolVersion: 'ipv4' | 'ipv6') => {
```
batzionb, line 70: you can avoid the code duplication of isNotLocalHostIPAddress and isNotCatchAllIpAddress by creating a function getIsNotValidIpAddress that gets the invalid addresses, which is the dictionary LOCAL_HOST_IP or CATCH_ALL_IP. But this is just a code improvement you can leave this as is, it's good enough

DanielleChason: I wanted to them separately because I think they can be useful for other validations

batzionb (dismissed):

src/ocm/components/clusterConfiguration/staticIp/commonValidationSchemas.tsx
batzionb: But it's good enough as is so I approved

src/ocm/components/clusterConfiguration/staticIp/commonValidationSchemas.tsx
```diff
@@ -50,6 +60,65 @@ export const getIpAddressValidationSchema = (protocolVersion: 'ipv4' | 'ipv6') =
   );
 };
 
+export const compareIPV6Addresses = (address1: Address6, address2: Address6) => {
+  if (JSON.stringify(address1.toByteArray()) === JSON.stringify(address2.toByteArray())) {
+    return true;
+  }
+  return false;
```
jkilzi, line 67:
```suggestion
  return JSON.stringify(address1.toByteArray()) === JSON.stringify(address2.toByteArray());
```

batzionb (approved):

jkilzi (approved):

PR #1409, opened 2022-05-31
MGMT-10640: Mismatch between UI and API regarding clustername length

jgyselov:
https://issues.redhat.com/browse/MGMT-10640

**OCM without unique name validation:**

image

**OCM with unique name validation:**

image

ACM has unique name validations and a minimum of 2 characters.


jgyselov:
@rawagner PTAL if it works with your side.

rawagner:
@jgyselov validation seems to work fine on ACM side. I see an issue though - you have `Use lowercase alphanumberic characters, or hyphen (-)` twice for `unique` validation (you can see it in your screenshot too)

PR #1411, opened 2022-05-31
MGMT-10540: Input validation for interface names in Static IP yaml view

DanielleChason:
https://issues.redhat.com/browse/MGMT-10540

image


src/ocm/components/clusterConfiguration/staticIp/components/YamlView/yamlViewValidationSchema.tsx
```diff
@@ -57,7 +57,8 @@ const macInterfaceMapValidationSchema = Yup.array<MacInterfaceMap>().of(
       .concat(getUniqueValidationSchema(getAllMacAddresses)),
     logicalNicName: Yup.string()
       .required(requiredMsg)
-      .concat(getUniqueValidationSchema(getInterfaceNamesInCurrentHost)),
+      .concat(getUniqueValidationSchema(getInterfaceNamesInCurrentHost))
+      .max(15, 'Interface name must be 15 characters at most.'),
```
jkilzi, line 61: Can an interface name contain any of these characters: `!@#$%^&*()` ????

jkilzi: @osherdp any chance we can use your help with regards to the validations that need to be performed on the interface name?

jkilzi (changes requested):

src/ocm/components/clusterConfiguration/staticIp/components/YamlView/yamlViewValidationSchema.tsx
osherdp: Seems like it can really be anything: https://unix.stackexchange.com/questions/451368/allowed-chars-in-linux-network-interface-names
I mean, there are conventions and also auto-generated names by linux that are the more likely to get, but we might have different cases in general

jkilzi (approved):

PR #1419, opened 2022-06-01
MGMT-10642: Hostname in UI chars limit is 253 (BE limit is 63) [BACKPORT]

jgyselov:
https://issues.redhat.com/browse/MGMT-10642

Added hostname validation regex. Hostnames must be valid (maximum of 62 characters between dots).

Before:
image

After:
image
image
image

### DIFF
--- a/src/common/components/ui/formik/validationSchemas.ts
+++ b/src/common/components/ui/formik/validationSchemas.ts
@@ -28,6 +28,8 @@ const PROXY_DNS_REGEX = /^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]
 const IP_V4_ZERO = '0.0.0.0';
 const IP_V6_ZERO = '0000:0000:0000:0000:0000:0000:0000:0000';
 const MAC_REGEX = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})|([0-9a-fA-F]{4}\\.[0-9a-fA-F]{4}\\.[0-9a-fA-F]{4})$”/;
+const HOST_NAME_REGEX = /^[^.]{1,63}(?:[.][^.]{1,63})*$/;
+
 //Source of information: https://github.com/metal3-io/baremetal-operator/blob/main/docs/api.md#baremetalhost-spec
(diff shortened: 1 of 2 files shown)


src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -27,7 +27,9 @@ const DNS_NAME_REGEX = /^([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}$/;
 const PROXY_DNS_REGEX = /^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[._]?$/;
 const IP_V4_ZERO = '0.0.0.0';
 const IP_V6_ZERO = '0000:0000:0000:0000:0000:0000:0000:0000';
-const MAC_REGEX = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})|([0-9a-fA-F]{4}\\.[0-9a-fA-F]{4}\\.[0-9a-fA-F]{4})$”/;
+const MAC_REGEX = /^([0-9A-Fa-f]{2}[:]){5}([0-9A-Fa-f]{2})$/;
+const HOST_NAME_REGEX = /^[^.]{1,63}(?:[.][^.]{1,63})*$/;
```
jkilzi, line 31: @jgyselov I think it was a good thing that you used the exact same regex the BE is using in order to validate this field, and I was about to ask you to just add a comment above this line with a permalink to the file where this regex is located in assisted-service code base

jgyselov: Well, the regex comes from [here](https://github.com/jgyselov/assisted-service/blob/master/internal/host/hostutil/host_utils.go#L70).

And the reason why I decided against using the exact same regex as they have in the BE is that I believe that if we have the conditions broken down into what is now six parts, each part should only check what the label says. So if this new label says "no. of characters between dots must be 1-63" I don't want this condition checking what kind of characters there are or if the overall length is <253 or anything else. I want the users to see what exactly made this condition fail and if they fix that exact problem that the label describes, I want the condition to pass.

In this case, the BE regex is broken down into three parts (_Use lowercase alphanumberic characters, dot (.) or hyphen (-)_, _Must start and end with an lowercase alphanumeric character_, and _Number of characters between dots (.) must be 1-63_)

(But I see that I accidentally changed the `MAC_REGEX` so I will revert that.)

jkilzi: ack

jkilzi: OK, I see in this example, it will fail the other steps of the validation: [a-z0-9-.]

jkilzi: For the record the BE's regex is located here: https://github.com/openshift/assisted-service/blob/284ef21780b9de4222bd4ca2a0e281f81f3ac321/internal/host/hostutil/host_utils.go#L70

PR #1428, opened 2022-06-07
MGMT-10703: Minimum requirements for SNO disk show 120GB

jkilzi:
https://issues.redhat.com/browse/MGMT-10703

- Removes the SNO requirements section from the modal in multi-node clusters
- Does some cleanup in the OCM folder
- includes one small change in CIM (@rawagner fyi)


celdrake:
@jkilzi Nice refactoring!
In terms of the content, I noticed some small details:

In SNO:

- The sentence ends with 2 "dots".
- This is not new, but should there be a "comma" before "and"? (it applies also to multi-node)
- For SNO, should we not also show the disk write speed requirements as we do for multi-node?

sno-requirements

In multi-node:
- See comment above regarding the "comma" + and

celdrake (approved):

PR #1429, opened 2022-06-07
MGMT-10110 Read-only cluster for non owner users

celdrake:
https://issues.redhat.com/browse/MGMT-10110

### Related PRs:
https://github.com/openshift-assisted/assisted-ui/pull/569  (mostly for documentation purposes)
https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/3398  (for the real behaviour on OCM)

### Description
When users other that the cluster owner are accessing assisted installer, they may not have edit permissions. Therefore the UI should not allow them to edit the cluster in any way, only to see the wizard and navigate through it.

There are different ways to do it depending on the application:

1. For standalone UI, we simulate this behaviour by defining a new Environment variable in `assisted-ui`.
The variable name is `REACT_APP_CLUSTER_PERMISSIONS`. See the README in `assisted-ui` for more details.
https://github.com/celdrake/assisted-ui/blob/MGMG-10110-read-only-cluster/README.md

When `canEdit` is set to `false`,  fields on the wizard become disabled for existing clusters, and the user can move to the next steps even if there are errors in previous ones. No PATCH should happen on the cluster.

2. For OCM, we receive the actual permissions from the subscription details.

**Note** This PR only covers Day 1 wizard. Day 2 would be separate to keep the PR smaller.

### DIFF
--- a/src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
+++ b/src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
@@ -2,51 +2,53 @@ import React, { useEffect } from 'react';
 import { Alert, AlertVariant, FormGroup, Stack, StackItem } from '@patternfly/react-core';
 import { FieldArray, useFormikContext, FormikHelpers } from 'formik';
 import { Cluster, MachineNetwork } from '../../../../common/api/types';
-import { HostSubnet, HostSubnets, NetworkConfigurationValues } from '../../../../common/types';
+import { HostSubnet, NetworkConfigurationValues } from '../../../../common/types';
 import { DUAL_STACK, NO_SUBNET_SET } from '../../../../common/config/constants';
 import { SelectField } from '../../../../common/components/ui';
 import { Address4, Address6 } from 'ip-address';
+import useClusterPermissions from '../../../hooks/useClusterPermissions';
+
+const subnetSort = (subA: HostSubnet, subB: HostSubnet) =>
+  subA.humanized.localeCompare(subB.humanized);
 
 const toFormSelectOptions = (subnets: HostSubnet[]) => {
-  return subnets
-    .sort((subA, subB) => subA.humanized.localeCompare(subB.humanized))
-    .map((hn, index) => ({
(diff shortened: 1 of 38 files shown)


celdrake:
@jkilzi We cannot use a version from `assisted-ui-lib` with these changes in an OCM version that does not pass the new Environment Variable. We could do it by adding a try/catch on the function `getBasePermissions` so that when used from OCM, it defaults to the current behaviour while in development it can be read-only if wanted.

src/ocm/components/clusterConfiguration/networkConfiguration/AvailableSubnetsControl.tsx
```diff
@@ -77,18 +97,13 @@ export const AvailableSubnetsControl = ({
...
-                      options={
-                        (index === 1 ? IPv6Subnets.length > 0 : IPv4Subnets.length > 0)
-                          ? [makeNoSubnetSelectedOption(hostSubnets)].concat(
-                              toFormSelectOptions(index === 1 ? IPv6Subnets : IPv4Subnets),
-                            )
-                          : [makeNoSubnetAvailableOption()]
-                      }
-                      isDisabled={!hostSubnets.length}
+                      options={buildOptions(machineSubnets)}
+                      isDisabled={isViewerMode || !hostSubnets.length}
```
jgyselov, line 106: There is no scenario where the select would need to be disabled based on the `hostSubnets`, so the `isDisabled` props 
 got removed from the selects in https://github.com/openshift-assisted/assisted-ui-lib/pull/1438. You can remove it from the conditions here, too.

jgyselov: > Probably the only condition should be `isDisabled={isViewerMode}`

That's exactly what I meant :slightly_smiling_face:

celdrake: I agree, once the code is updated in `master` I'll adjust this PR.

celdrake:
Creating a new PR with a simpler implementation.

PR #1430, opened 2022-06-07
MGMT-9380: Assisted Installer Single Node + Day 2 workers

jgyselov:
https://issues.redhat.com/browse/MGMT-9380
https://issues.redhat.com/browse/MGMT-10731
https://issues.redhat.com/browse/MGMT-10086
https://issues.redhat.com/browse/MGMT-10701

image

Preview of what the alert will look like once OpenShift 4.11 is fully supported:
image
image

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -1,0 +1,12 @@
+const MIN_OCP_VERSION = '4.11.0';
+
+export const canAddHostSNO = (ocpVersion: string, minVersion: string = MIN_OCP_VERSION) => {
+  const ocpVersionSplit = ocpVersion.match(/\d+(\.\d+)+/);
+  if (ocpVersionSplit?.length) {
+    const ocpVersionParsed = ocpVersionSplit[0].split('.').map((num) => Number(num)) || [-1];
+    const minVersionParsed = minVersion.split('.').map((num) => Number(num));
+
+    return ocpVersionParsed.every((v, i) => v >= minVersionParsed[i]);
+  }
+  return false;
+};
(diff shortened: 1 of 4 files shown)


src/common/utils.ts
```diff
@@ -0,0 +1,12 @@
+const MIN_OCP_VERSION = '4.11.0';
```
romfreiman, line 1: hm, didnt we add an api that defines the support matrix?

jkilzi: @osherdp can we get this info^ from the feature-support-levels API?
@jgyselov PR has merge conflicts

osherdp: I see no reason why not
Epic developers may give a better answer than me @omertuc @paul-maidment

omertuc: There is

https://github.com/openshift/assisted-service/blob/208bbd5876a6f3b6c1bb290e8bc17c3e90972fdb/swagger.yaml#L3626

jgyselov: I am aware of it and I'm working on integrating it with this feature.

src/ocm/components/AddHosts/utils.ts
```diff
@@ -7,14 +8,18 @@ export const canAddHost = ({ cluster }: { cluster: OcmClusterType }) => {
     if (cluster.aiCluster) {
       if (!isSNO(cluster.aiCluster)) {
         return cluster.aiCluster.status === 'installed';
+      } else {
```
celdrake, line 11: It would be more readable to invert the conditions to avoid the double negation on the "else"
```suggestion
      if (isSNO(cluster.aiCluster)) {
       return (
          cluster.aiCluster.status === 'installed' &&
          canAddHostSNO(cluster.aiCluster.openshiftVersion || cluster.openshift_version)
        );
      } else {
          return cluster.aiCluster.status === 'installed';
      }
```

celdrake (changes requested):

celdrake (approved):

PR #1431, opened 2022-06-07
MGMT-10708: Base Domain name - required message - fix

DanielleChason:
https://issues.redhat.com/browse/MGMT-10708

**before:**
image

**now**: 
image

image


jkilzi (approved):

jkilzi:
@nirfarkas shouldn't those helper texts be colored red?
@celdrake is this what we discussed this morning right? PTAL

celdrake:
> @nirfarkas shouldn't those helper texts be colored red? @celdrake is this what we discussed this morning right? PTAL

Yes, this is what I mentioned. The styling is lost as a consequence of some data changes. I spoke to @DanielleChason to let her know where the problem was and hopefully fix it too in this task.

DanielleChason:
I changed the error message color back to red, with the help of @celdrake 
image

src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -23,7 +23,10 @@ const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undef
     }
   }, [field.value, setHadValue]);
   const showError = field.value || (!field.value && hadValue);
-  const errorMessage = (showError ? error : '') || inputError;
+  let errorMessage = (showError ? error : '') || inputError;
+  if (errorMessage) {
```
celdrake, line 27: Since this is a workaround to some unexpected data we are receiving, I think it's best to indicate why we are doing this change. Either define a function whose name implies it, or add a comment for it.

Otherwise someone may be tempted to remove this code since theoretically at this point, `errorMessage` must be a string already. 
Also, it may be better to split the lines using "dot" as the separator instead of  "comma" so they read as independent phrases and are easier to understand.

celdrake (changes requested):

src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -23,7 +23,10 @@ const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undef
     }
   }, [field.value, setHadValue]);
   const showError = field.value || (!field.value && hadValue);
-  const errorMessage = (showError ? error : '') || inputError;
+  let errorMessage = (showError ? error : '') || inputError;
```
jkilzi, line 26: Why not just do this?
```suggestion
  let errorMessage = showError ? error : inputError;
```

jkilzi (changes requested):

src/common/hooks/useFieldErrorMsg.tsx
DanielleChason: This is code from before, but sure, I'll change it

celdrake: I don't think the code is equivalent so it should not be changed.
If  `inputError` is truthy, it's used as a fallback if there is no error or we don't want to display it.
This only seems to be used in `BMCForm` but I think it would affect it.

src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -7,6 +7,14 @@ type Props = {
   validate?: FieldValidator;
 };
 
+// With getRichTextValidation we can receive an array of strings so it needs to be converted to string
+const adaptErrorInvalidFormat = (error: undefined | string []): string | undefined => {
```
jkilzi, line 11:
```suggestion
const adaptErrorInvalidFormat = (error?: string []): string | undefined => {
```

src/common/hooks/useFieldErrorMsg.tsx
```diff
@@ -23,7 +31,7 @@ const useFieldErrorMsg = ({ name, inputError, validate }: Props): string | undef
     }
   }, [field.value, setHadValue]);
   const showError = field.value || (!field.value && hadValue);
-  const errorMessage = (showError ? error : '') || inputError;
+  const errorMessage = adaptErrorInvalidFormat((showError ? error : '') || inputError);
```
jkilzi, line 34: if `error` is a falsy value then we'll display `inputError` despite `showError === true`. A bug?

jkilzi (changes requested):

src/common/hooks/useFieldErrorMsg.tsx
celdrake: So, if there is no error from the validation side, it will display the error coming from `inputError` (from Formik directly).

I'm assuming those would be two different errors, and that the "inputError" needs to be shown if it's truthy.
https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/cim/components/Agent/BMCForm.tsx#L75

jkilzi (approved):

celdrake (approved):

PR #1450, opened 2022-06-16
MGMT-10663 auto assign role

celdrake:
https://issues.redhat.com/browse/MGMT-10663

When a host has the "auto-assign" role, we show whether it would be a  `control plane node` or a `worker` based on the `suggested_role`. To avoid confusion for the user, the label and description of the Dropdown items has changed:

- If the cluster has schedulable_masters on, then the label for "Control plane nodes" becomes  "Control plane, worker".
- If the cluster has schedulable_masters off, then the label for "Control plane nodes" stays the same
The description also takes into consideration the value of `schedulable_masters`

Dropdown items: label and description for `master` with `schedulable_masters`
host-role-desc-master

Dropdown items: Label 
host-role-master-worker

Dropdown toggle: 
role-2-schedulable

### DIFF
--- a/src/common/components/hosts/RoleCell.tsx
+++ b/src/common/components/hosts/RoleCell.tsx
@@ -11,17 +11,25 @@ export type RoleCellProps = {
   onEditRole?: (role: HostUpdateParams['hostRole']) => Promise<any>;
   displayTooltip?: boolean;
   position?: DropdownProps['position'];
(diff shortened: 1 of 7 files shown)


celdrake:
Closing since it's not clear if the UI is going to need any change and with translations most of the work will need heavy adaptations.

PR #1455, opened 2022-06-21
NO-ISSUE: Fix Eslint warnings for no-floating-promises

celdrake:
Removes linting errors due to the "floating-promises" rule.
This can be easily avoided by explicitly indicating we are not using the promise result (using `void`)

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md

The rule is now activated to avoid adding new violations of this rule.


jkilzi (approved):

jkilzi:
ESlint warnings count dropped to 396 with this change!
https://github.com/openshift-assisted/assisted-ui-lib/runs/6981835692?check_suite_focus=true#step:5:653

PR #1456, opened 2022-06-21
NO_ISSUE: reduce eslint warnings

celdrake:
Fixes a few ESlint warnings from the following rules:

Completely removed:
`@typescript-eslint/restrict-plus-operands`
`@typescript-eslint/require-await`

Partially fixed from `@typescript-eslint/restrict-template-expressions` and similar rules where TS sees the variable as potentially undefined. We will need to fix them individually so they have not been dropped entirely.

Drops about 50 errors.

### DIFF
--- a/src/ocm/components/clusterDetail/utils.tsx
+++ b/src/ocm/components/clusterDetail/utils.tsx
@@ -30,7 +30,7 @@ export const downloadClusterInstallationLogs = async (
       saveAs(data, fileName);
     }
   } catch (e) {
-    handleApiError(e, async (e) => {
+    handleApiError(e, (e) => {
       addAlert({
         title: 'Could not download cluster installation logs.',
         message: getApiErrorMessage(e),
--- a/src/ocm/components/clusterDetail/utils.tsx
+++ b/src/ocm/components/clusterDetail/utils.tsx
@@ -72,6 +72,7 @@ const getClusterResources = (cluster: Cluster, resourcePath: string): number => 
     )
     .map((host) => stringToJSON<Inventory>(host.inventory) || {})
     .map((inventory) => get(inventory, resourcePath, 0))
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
     .reduce((total, value) => total + value, 0);
 
   return result;
--- a/src/ocm/components/clusterDetail/utils.tsx
+++ b/src/ocm/components/clusterDetail/utils.tsx
@@ -71,8 +71,7 @@ const getClusterResources = (cluster: Cluster, resourcePath: string): number => 
         (host.role === 'master' && singleNodeMode),
     )
     .map((host) => stringToJSON<Inventory>(host.inventory) || {})
-    .map((inventory) => get(inventory, resourcePath, 0))
-    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+    .map((inventory) => get(inventory, resourcePath, 0) as number)
     .reduce((total, value) => total + value, 0);
 
   return result;
--- a/src/ocm/components/hosts/ResetHostModal.tsx
+++ b/src/ocm/components/hosts/ResetHostModal.tsx
@@ -28,7 +28,7 @@ const ResetHostModal: React.FC<ResetHostModalProps> = ({ isOpen, hostname, onClo
       </Button>,
     ]}
   >
-    Are you sure you want to reset host{` ${hostname}` || ''} ?
+    Are you sure you want to reset host{` ${hostname || ''}`} ?
   </Modal>
 );
 
--- a/src/ocm/hooks/useInfraEnv.ts
+++ b/src/ocm/hooks/useInfraEnv.ts
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useInfraEnvId } from '.';
 import { Cluster, InfraEnv, InfraEnvUpdateParams } from '../../common';
-import { getErrorMessage } from '../api';
+import { getErrorMessage } from '../../common/utils';
 import { InfraEnvsAPI } from '../services/apis';
 
 export default function useInfraEnv(clusterId: Cluster['id']) {
--- a/src/ocm/hooks/useInfraEnv.ts
+++ b/src/ocm/hooks/useInfraEnv.ts
@@ -1,14 +1,12 @@
 import React from 'react';
 import { useInfraEnvId } from '.';
 import { Cluster, InfraEnv, InfraEnvUpdateParams } from '../../common';
-import { APIErrorMixin } from '../api/types';
+import { getErrorMessage } from '../api';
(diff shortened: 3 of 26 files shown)


src/ocm/components/hosts/ResetHostModal.tsx
```diff
@@ -28,7 +28,7 @@ const ResetHostModal: React.FC<ResetHostModalProps> = ({ isOpen, hostname, onClo
       </Button>,
     ]}
   >
-    Are you sure you want to reset host{` ${hostname}` || ''} ?
+    Are you sure you want to reset host{` ${hostname || ''}`} ?
```
jkilzi, line 31: same as in https://github.com/openshift-assisted/assisted-ui-lib/pull/1456/files#r902576434

src/ocm/services/apis/ClustersAPI.ts
```diff
@@ -178,7 +178,7 @@ const ClustersAPI = {
 
   listByOpenshiftId(openshiftId: Cluster['openshiftClusterId']) {
     return client.get<Cluster[]>(
-      `${ClustersAPI.makeBaseURI()}?openshift_cluster_id=${openshiftId}`,
+      `${ClustersAPI.makeBaseURI()}?openshift_cluster_id=${openshiftId || ''}`,
```
jkilzi, line 181: In this case, if `openshiftId` is undefined then there is no point in calling this API

jkilzi (approved):

src/ocm/components/clusterDetail/utils.tsx
```diff
@@ -72,6 +72,7 @@ const getClusterResources = (cluster: Cluster, resourcePath: string): number =>
     )
     .map((host) => stringToJSON<Inventory>(host.inventory) || {})
     .map((inventory) => get(inventory, resourcePath, 0))
```
jtomasek, line 74: Following change fixes all linting errors on that function

```suggestion
    .map((inventory) => get(inventory, resourcePath, 0) as number)
```

src/ocm/components/clusterDetail/utils.tsx
```diff
@@ -72,6 +72,7 @@ const getClusterResources = (cluster: Cluster, resourcePath: string): number =>
...
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
```
jtomasek, line 75: not needed with above change

jtomasek (changes requested):

src/ocm/components/clusterDetail/utils.tsx
celdrake: I was not sure if the data could contain non-numbers.  I'll apply your suggestion, thanks!.

src/ocm/services/apis/ClustersAPI.ts
celdrake: Agree, it's just that with the current type definition it needs the fallback for TS.

jtomasek:
A few additional linting fixes here: https://github.com/openshift-assisted/assisted-ui-lib/pull/1458

jtomasek (dismissed):

src/ocm/hooks/useClusterPreflightRequirements.ts
```diff
@@ -3,21 +3,24 @@ import { alertsSlice } from '../../common/reducers';
...
+
+type ErrorType = APIErrorMixin & Error;
 
 const { addAlert } = alertsSlice.actions;
 
 const useClusterPreflightRequirements = (clusterId: Cluster['id']) => {
   const [preflightRequirements, setPreflightRequirements] =
     React.useState<PreflightHardwareRequirements>();
-  const [error, setError] = React.useState();
+  const [error, setError] = React.useState<ErrorType>();
```
jtomasek, line 15: Not entirely related to your changes but I suspect that this should be a string - an error message rather than an ErrorType?

src/ocm/hooks/useClusterPreflightRequirements.ts
```diff
@@ -3,21 +3,24 @@ import { alertsSlice } from '../../common/reducers';
...
 
   React.useEffect(() => {
     const fetchFunc = async () => {
       try {
         const { data } = await ClustersAPI.getPreflightRequirements(clusterId);
         setPreflightRequirements(data);
       } catch (e) {
-        setError(e);
+        setError(e as ErrorType);
```
jtomasek, line 23: This is how I fix it in my PR https://github.com/openshift-assisted/assisted-ui-lib/pull/1458/files#diff-fb2591b635c55d45a38368c2a889bbb2558786a5aa870ad0cfbcd47b2b7451cfR21

src/ocm/hooks/useInfraEnv.ts
```diff
@@ -16,7 +18,7 @@ export default function useInfraEnv(clusterId: Cluster['id']) {
         setInfraEnv(infraEnv);
       }
     } catch (e) {
-      setError(e);
+      setError(e as ErrorType);
```
jtomasek, line 21: Same here

src/ocm/hooks/useInfraEnvId.ts
```diff
@@ -3,16 +3,18 @@ import { Cluster, InfraEnv } from '../../common';
...
-  const [error, setError] = React.useState<APIErrorMixin & Error>();
+  const [error, setError] = React.useState<ErrorType>();
 
   const findInfraEnvId = React.useCallback(async () => {
     try {
       const infraEnvId = await InfraEnvsService.getInfraEnvId(clusterId);
       setInfraEnv(infraEnvId);
     } catch (e) {
-      setError(e);
+      setError(e as ErrorType);
```
jtomasek, line 17: And here as well. Looking at the usage the error is basically only used as a boolean (if (error) ...) so I think it would be better to just store the error message?

src/ocm/hooks/useClusterPreflightRequirements.ts
celdrake: I'll revert this change in the PR given that you have addressed it in a better way in yours.

src/ocm/hooks/useClusterPreflightRequirements.ts
celdrake: I agree. Please look at the new commit https://github.com/openshift-assisted/assisted-ui-lib/pull/1456/commits/0e57cafb04cce22dd9b6866c5ebf23c87dde52bf

Changes done:
`useClusterPreflightRequirements`: reverted because it's addressed in your PR
`useInfraEnv`, `useInfraEnvId` and `useInfraEnvImageUrl`: these are not handled in your PR. They are now defined as strings given they are being used in reality more as booleans.

jkilzi:
@celdrake merge conflicts...

jkilzi (approved):

PR #1458, opened 2022-06-22
NO_ISSUE Lint fixes

jtomasek:
- Fixes error message handling in catch clauses
- Additional lint fixes


src/ocm/services/HostsService.ts
```diff
@@ -15,7 +15,7 @@ const HostsService = {
       const hosts = await HostsService.listHostsBoundToCluster(clusterId);
       const hostDeleteParams = hosts.map<[string, string]>(({ infraEnvId, id }) => [
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
celdrake, line 17:
```suggestion
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
Can be removed

src/ocm/components/clusterConfiguration/DiscoveryImageForm.tsx
```diff
@@ -61,7 +61,8 @@ const DiscoveryImageForm: React.FC<DiscoveryImageFormProps> = ({
           formikActions.setStatus({
             error: {
               title: 'Failed to download the discovery Image',
-              message: getErrorMessage(error),
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
celdrake, line 64: Same about the exception, best to remove them ?

src/ocm/components/AddHosts/HostsClusterDetailTab.tsx
```diff
@@ -127,21 +128,24 @@ const HostsClusterDetailTabContent: React.FC<HostsClusterDetailTabProps> = ({
...
-            <>
-              {e.request._url.includes('/import')
-                ? 'Failed to create wrapping cluster for adding new hosts.'
-                : 'Failed to fetch cluster for adding new hosts.'}
-              <br />
-              Check your connection and <TryAgain />.
-            </>,
-          );
+          if (isApiError(e) && e.request) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
celdrake, line 132: Could we remove the exceptions? They are already warnings from .estlintrc, and it makes them easier to find then when we remove the exception.

celdrake (changes requested):
Good to see the numbers drop!

celdrake (approved):

src/ocm/services/HostsService.ts
```diff
@@ -14,8 +14,7 @@ const HostsService = {
       const promises: AxiosPromise<void>[] = [];
       const hosts = await HostsService.listHostsBoundToCluster(clusterId);
       const hostDeleteParams = hosts.map<[string, string]>(({ infraEnvId, id }) => [
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        infraEnvId!,
+        infraEnvId || '',
```
jkilzi, line 17: How can we delete a host if we don't have the infraEnvId !?!?! :)

jkilzi (approved):

src/ocm/services/HostsService.ts
jgyselov: We can't. :upside_down_face:

jkilzi: I think the assumption was that [here](https://github.com/jtomasek/assisted-ui-lib/blob/26937b7e99ddd35b9cd13b1b4a5bf180a4b4a836/src/ocm/services/HostsService.ts#L15) we'll get for sure  a list of hosts for which all of them have the `infraEnvId` field set

celdrake: The problem is the same as with other places, TS thinks `infraEnvId` can be unset because of the API type definition.
https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/api/types.ts#L887

If we want to avoid having fallbacks where it doesn't make sense, maybe we can use `assert` .
`declare function assert(value: unknown): asserts value;`

Another solution would be to redefine the API types making the fields mandatory where we know they should be, so the code could use the updated type and not the base API type.
eg.  something like https://bobbyhadz.com/blog/typescript-make-property-required
```
type WithRequiredProperty<Type, Key extends keyof Type> = Type & {
  [Property in Key]-?: Type[Property];
}
```

PR #1460, opened 2022-06-23
MGMT-10901: Clean up the networking file structure

jgyselov:
https://issues.redhat.com/browse/MGMT-10901


jgyselov:
@openshift-assisted/ui-dev-cim PTAL, I moved a couple files from `/common` to `/cim` as we were no longer using them on the OCM side. I would appreciate it if you could let me know that everything works properly with these changes -- I ran some static tests, but I have no way of actually checking the behaviour.

jkilzi:
@jgyselov please take into account that if you put the changes for OCM in a different PR then we might be able to merge them independently

jkilzi (approved):

celdrake:
@jgyselov please fix conflicts

celdrake (approved):

PR #1461, opened 2022-06-23
MGMT-10894: Adding IPv6 info in the cluster details section

jgyselov:
https://issues.redhat.com/browse/MGMT-10894

**Dual stack cluster:**
image

**Ipv4 cluster:**
image


jgyselov:
@nirfarkas PTAL.

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -74,82 +75,112 @@ const getDiskEncryptionEnabledOnStatus = (diskEncryption: DiskEncryption['enable
...
+          <DetailItem
+            title="Cluster network host prefix (IPv4)"
+            value={selectClusterNetworkHostPrefix(cluster)}
+            testId="host-prefix-ipv4"
+          />
+          {isDualstack ? (
+            <>
+              <DetailItem
+                title="Cluster network CIDR (IPv6)"
+                value={cluster.clusterNetworks && cluster.clusterNetworks[1].cidr}
```
celdrake, line 151:
```suggestion
                value={cluster.clusterNetworks[1].cidr}
```
It is ensured by the previous conditions that there must be more than 1 `clusterNetwork`

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -74,82 +75,112 @@ const getDiskEncryptionEnabledOnStatus = (diskEncryption: DiskEncryption['enable
...
+                testId="cluster-network-cidr-ipv6"
+              />
+              <DetailItem
+                title="Cluster network host prefix (IPv6)"
+                value={cluster.clusterNetworks && cluster.clusterNetworks[1].hostPrefix}
```
celdrake, line 156:
```suggestion
                value={cluster.clusterNetworks[1].hostPrefix}
```
Same here

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -74,82 +75,112 @@ const getDiskEncryptionEnabledOnStatus = (diskEncryption: DiskEncryption['enable
...
+          ) : undefined}
+          <DetailItem
+            title="Service network CIDR (IPv4)"
+            value={selectServiceNetworkCIDR(cluster)}
+            testId="service-network-cidr-ipv4"
+          />
+          {isDualstack ? (
+            <DetailItem
+              title="Service network CIDR (IPv6)"
+              value={cluster.serviceNetworks && cluster.serviceNetworks[1].cidr}
```
celdrake, line 169:
```suggestion
              value={cluster.serviceNetworks[1].cidr}
```
Same here

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -74,82 +75,112 @@ const getDiskEncryptionEnabledOnStatus = (diskEncryption: DiskEncryption['enable
...
-          isHidden={cluster.diskEncryption?.enableOn === 'none'}
-          testId="disk-encryption"
-        />
-        <ClusterFeatureSupportLevelsDetailItem cluster={cluster} />
-      </DetailList>
-    </GridItem>
-  </>
-);
+          <DetailItem
+            title="Cluster network host prefix (IPv4)"
```
celdrake, line 143: I think in other cases we didn't want to show "IPv4" when it's a Single Stack cluster.
Let's see what @nirfarkas thinks

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -74,82 +75,112 @@ const getDiskEncryptionEnabledOnStatus = (diskEncryption: DiskEncryption['enable
...
+            value={selectClusterNetworkHostPrefix(cluster)}
+            testId="host-prefix-ipv4"
+          />
+          {isDualstack ? (
```
celdrake, line 147: A minor thing, but if we only render in one clause, I'd use the pattern `isDualstack && render()` instead of `isDualStack ? render() : undefined`

celdrake (changes requested):

src/ocm/components/clusterDetail/ClusterProperties.tsx
jgyselov: I wanted to but I get errors when I try that: 
```
Type 'false | Element' is not assignable to type 'string | number | ReactElement<any, string | JSXElementConstructor<any>> | (ReactElement<any, string | JSXElementConstructor<any>> & string) | undefined'.
Type 'false' is not assignable to type 'string | number | ReactElement<any, string | JSXElementConstructor<any>> | (ReactElement<any, string | JSXElementConstructor<any>> & string) | undefined'.
```
The same construction is used earlier with `externalMode`.

celdrake: No problem, thanks for checking :)

src/common/selectors/clusterSelectors.ts
```diff
@@ -66,17 +61,23 @@ export const allSubnetsIPv4 = (
...
 
-export const isIPv4 = ({
-  machineNetworks,
-  clusterNetworks,
-  serviceNetworks,
-}: Cluster | NetworkConfigurationValues) =>
-  allSubnetsIPv4(machineNetworks) &&
-  allSubnetsIPv4(clusterNetworks) &&
-  allSubnetsIPv4(serviceNetworks);
+export const isDualStack = (cluster: Cluster) =>
```
celdrake, line 73: Before,  `isIPv4` returned true if all three elements had only networks using IPv4. So dual stack would be true if ANY of the three elements had at least one item with IPv6.

The new `isDualStack` returns true if ALL three elements have networks using IPv6, so the logic has slightly changed.

Could you please confirm that this is correct? I'm still learning how Dual Stack should work.

jkilzi: There should be 2 situations:
1. A cluster could be configured to use DS
2. The cluster is already configured to use DS

For the first case, this happens when the cluster.host_networks include at least one IPv4 and one IPv6 address (excluding loopback addresses probably? @mkowalski WDYT?).
The second occurs when the machine, cluster, and service networks include at least one IPv4 and one IPv6 address

src/common/selectors/clusterSelectors.ts
```diff
@@ -57,14 +61,30 @@ export const allSubnetsIPv4 = (
   return !!networks?.every((network) => network.cidr && Address4.isValid(network.cidr));
 };
 
+export const hasIpv6Subnet = (
+  networks: (MachineNetwork | ClusterNetwork | ServiceNetwork)[] | undefined,
```
jkilzi, line 65: The idea behind this module is that all the functions act on an instance of the cluster object :)

src/common/selectors/clusterSelectors.ts
mkowalski: The above is correct.

You can assume the hosts *can form* a dual-stack cluster if `cluster.host_networks` have IPv4 *and* IPv6 available there.

The cluster *is already configured* as a dual-stack if all 3 *-networks have IPv4 *and* IPv6 subnets added. The scenario where only some of *-networks have dual-stack subnets but not all of them should be stopped by the validations in the service.

src/common/selectors/clusterSelectors.ts
jgyselov: That hasn't been the case for a while...
What would be best - rename the file to something like `helpers.ts` or `utils.ts` or move some of the functions to somewhere else?

src/common/selectors/clusterSelectors.ts
jgyselov: I've renamed the functions to `isDualStack` and `canBeDualStack`. I hope that makes things a bit clearer. :slightly_smiling_face:

src/common/selectors/clusterSelectors.ts
jkilzi: [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle), each module must have only one duty. `clusterSelectors.ts` should standardize the way we read data out of the Cluster model. Having files names `utils/helpers` all over the place makes it difficult to track what their scope is.
So if it hasn't been the case for a while, now is a good opportunity to improve it (as long as it doesn't require refactoring the whole code base, of course)

src/ocm/components/clusterConfiguration/ReviewCluster.tsx
```diff
@@ -55,7 +55,7 @@ const ReviewCluster: React.FC<{ cluster: Cluster }> = ({ cluster }) => {
       />
       <DetailItem
         title={'Stack type'}
-        value={isIPv4(cluster) ? 'IPv4' : 'Dual-stack'}
+        value={isDualStack(cluster) ? 'Dual-stack' : 'IPv4'}
```
jkilzi, line 58: FYI, that the day they'll implement single-stack w/IPv6 only, we'll need to change this...

jgyselov: We will need to change a lot of things once that day comes :slightly_smiling_face:

jgyselov:
I moved the functions that didn't fit into `clusterSelectors` into different files, mostly `networkConfigurations/utils` since we already had similar functions there.

src/ocm/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -87,7 +88,7 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
             </TextContent>
             <ClusterHostsTable cluster={cluster} skipDisabled />
           </GridItem>
-          <ClusterProperties cluster={cluster} />
+          <ClusterProperties cluster={cluster} isDualStack={isDualStack(cluster)} />
```
jkilzi, line 91: If ClusterProperties receives the whole cluster obj, why does it need an additional `isDualStack` prop? :)

src/ocm/components/clusterDetail/ClusterProperties.tsx
```diff
@@ -73,7 +75,11 @@ const getDiskEncryptionEnabledOnStatus = (diskEncryption: DiskEncryption['enable
   return diskEncryptionType;
 };
 
-const ClusterProperties: React.FC<ClusterPropertiesProps> = ({ cluster, externalMode = false }) => (
+const ClusterProperties = ({
+  cluster,
+  isDualStack = false,
```
jkilzi, line 80: isDualStack can be inferred from the cluster obj

src/common/selectors/clusterSelectors.ts
```diff
@@ -1,70 +1,82 @@
...
-  clusterNetworks,
-  serviceNetworks,
-}: Cluster | NetworkConfigurationValues) =>
-  allSubnetsIPv4(machineNetworks) &&
-  allSubnetsIPv4(clusterNetworks) &&
-  allSubnetsIPv4(serviceNetworks);
+export const selectIpv6HostPrefix = ({ clusterNetworks }: Pick<Cluster, 'clusterNetworks'>) =>
+  clusterNetworks && clusterNetworks[1].hostPrefix;
+
+export const isArmArchitecture = ({ cpuArchitecture }: Partial<Cluster>) =>
```
jkilzi, line 81:
```suggestion
export const isArmArchitecture = ({ cpuArchitecture }: Pick<Cluster, 'cpuArchitecture'>) =>
```

jgyselov: Oops, missed that one.

celdrake (dismissed):

jkilzi (commented):
PTAL at my comments, just as suggestions

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -104,6 +105,47 @@ export const isAdvNetworkConf = (
...
+  clusterNetworkCidr,
+  machineNetworkCidr,
+  serviceNetworkCidr,
+}: Partial<Cluster>) => {
+  return (
+    Address6.isValid(clusterNetworkCidr || '') ||
+    Address6.isValid(machineNetworkCidr || '') ||
+    Address6.isValid(serviceNetworkCidr || '')
+  );
+};
```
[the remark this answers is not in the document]
jgyselov: They probably could, but in my mind they fit better into clusterConfiguration. :slightly_smiling_face:

jkilzi (approved):

celdrake (approved):

src/common/components/clusterConfiguration/utils.ts
```diff
@@ -105,6 +106,44 @@ export const isAdvNetworkConf = (
...
+export const isDualStack = ({
+  machineNetworks,
+  clusterNetworks,
+  serviceNetworks,
+}: Pick<Cluster, 'machineNetworks' | 'clusterNetworks' | 'serviceNetworks'>) =>
+  areNetworksDualStack(machineNetworks) &&
+  areNetworksDualStack(clusterNetworks) &&
+  areNetworksDualStack(serviceNetworks);
+
+export const isSubnetInIPv6 = ({
```
celdrake, line 138:
```suggestion
export const isSubnetInIPv6 = ({
  clusterNetworkCidr,
  machineNetworkCidr,
  serviceNetworkCidr,
}: Pick<Cluster, 'clusterNetworkCidr' | 'machineNetworkCidr' | 'serviceNetworkCidr'>) =>

I just noticed this TS error. IMO we should make sure that Eslint catches these errors as our code will be more robust and we'll catch errors automatically. @jkilzi I'll try to fix this so tht linting fails in these cases.
```

celdrake (changes requested):

src/common/components/clusterConfiguration/utils.ts
celdrake: The issue is one that we have excluded, `@typescript-eslint/no-unsafe-call` and to fix the errors won't be very quick, so for now we won't have them fixed.

celdrake (approved):

jkilzi (approved):

PR #1462, opened 2022-06-23
MGMG-10924 Avoid OlmOperators failing when its value is null and not undefined

celdrake:
https://issues.redhat.com/browse/MGMT-10924

We have detected that sometimes we receive `null` for arrays from the API.
Because of this, the defaults in functions etc won't work, it only works for `undefined` / not passed value.

I fixed the `olmOperators` but I suspect we could run into trouble with other fields so I suggest to try to fix the route cause and only handle empty arrays for missing data

Eg. request to retrieve the day2 cluster, the response contains null values.
import-null-arrays
.


jgyselov (approved):

jkilzi (approved):

src/common/components/clusterDetail/utils.ts
```diff
@@ -1,4 +1,10 @@
 import { MonitoredOperatorsList } from '../../api';
 
-export const getOlmOperators = (monitoredOperators: MonitoredOperatorsList = []) =>
-  monitoredOperators.filter((operator) => operator.operatorType === 'olm');
+export const getOlmOperators = (monitoredOperators: MonitoredOperatorsList | undefined) => {
```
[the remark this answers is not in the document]
celdrake: Thanks, good idea

src/common/selectors/clusterSelectors.ts
```diff
@@ -68,3 +74,12 @@ export const isIPv4 = ({
   allSubnetsIPv4(machineNetworks) &&
   allSubnetsIPv4(clusterNetworks) &&
   allSubnetsIPv4(serviceNetworks);
+
+export const getOlmOperators = (monitoredOperators: MonitoredOperatorsList | undefined) => {
```
jkilzi, line 78:
```suggestion
export const getOlmOperators = (cluster: Partial<Cluster>) => {
```

jgyselov: Maybe even `Pick<Cluster, ...>` instead of `Partial<>`? So that we are more explicit about what we are using from the Cluster object in the function?

src/common/components/clusterDetail/utils.ts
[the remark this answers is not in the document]
celdrake: And also updated to use directly the `cluster` object

src/common/selectors/clusterSelectors.ts
celdrake: Agree, see the last commits for the update :+1:

src/ocm/components/clusterDetail/ClusterDetailStatusVarieties.tsx
```diff
@@ -28,12 +28,11 @@ export const useClusterStatusVarieties = (cluster?: Cluster): ClusterStatusVarie
 
   const clusterId = cluster?.id;
   const clusterStatus = cluster?.status;
-  const clusterMonitoredOperators = cluster?.monitoredOperators;
-
-  const olmOperators = getOlmOperators(cluster?.monitoredOperators);
+  const clusterMonitoredOperators = cluster?.monitoredOperators || [];
```
jkilzi, line 31:
```suggestion
  const clusterMonitoredOperators = getMonitoredOperators(cluster);
  ...
  // clusterSelectors.ts
  export const getMonitoredOperators = (cluster?: Pick<Cluster, 'monitoredOperators'>) =>
    cluster?.monitoredOperators || [];
```

jkilzi (changes requested):
one more nit

src/common/selectors/clusterSelectors.ts
```diff
@@ -68,3 +68,9 @@ export const isIPv4 = ({
   allSubnetsIPv4(machineNetworks) &&
   allSubnetsIPv4(clusterNetworks) &&
   allSubnetsIPv4(serviceNetworks);
+
+export const getOlmOperators = (cluster?: Pick<Cluster, 'monitoredOperators'>) => {
+  // monitoredOperators can sometimes be either undefined or also null, we must use the fallback
+  const monitoredOperators = cluster?.monitoredOperators || [];
```
jkilzi, line 74:
```suggestion
  const monitoredOperators = getMonitoredOperators(cluster); // or maybe "selectMonitoredOperators"
```

PR #1465, opened 2022-06-24
MGMG-10924 Avoid OlmOperators failing when its value is null and not undefined

celdrake:
Port of https://github.com/openshift-assisted/assisted-ui-lib/pull/1462


src/common/components/clusterDetail/utils.ts
```diff
@@ -1,4 +1,10 @@
 import { MonitoredOperatorsList } from '../../api';
 
-export const getOlmOperators = (monitoredOperators: MonitoredOperatorsList = []) =>
-  monitoredOperators.filter((operator) => operator.operatorType === 'olm');
+export const getOlmOperators = (monitoredOperators: MonitoredOperatorsList | undefined) => {
```
jkilzi, line 3: @celdrake please move this to clusterSelectors

jkilzi (changes requested):

celdrake:
@jkilzi I applied the new commits from the original PR.

jkilzi (approved):

PR #1466, opened 2022-06-24
NO_ISSUE Unversion settings file from VS ccode

celdrake:
The file `.vscode/settings.json` is versioned despite being in `.gitignore`.

They must have updated the settings definition, because I keep getting a difference in
```
  "editor.quickSuggestions": {
    "comments": "on",
    "strings": "on",
    "other": "on"
  }
```

So either we update the commited file (but maybe not everyone has the same IDE version), or remove it.
If merged, those using VsCode will need to re-create the file, only it won't show in git anymore.


jkilzi (approved):

celdrake:
Closed in favor of updating the file https://github.com/openshift-assisted/assisted-ui-lib/pull/1470

PR #1476, opened 2022-06-28
MGMT-10862: Make use of feature-support-levels for day2 SNO

jgyselov:
https://issues.redhat.com/browse/MGMT-10862

UHC-portal PR: https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/3433

### DIFF
--- a/src/common/components/clusterConfiguration/SNODisclaimer.tsx
+++ b/src/common/components/clusterConfiguration/SNODisclaimer.tsx
@@ -2,16 +2,17 @@ import React from 'react';
 import { Alert, AlertVariant, List, ListItem, Stack, StackItem } from '@patternfly/react-core';
 import { CheckboxField } from '../ui';
 import { SupportLevel } from '../../types';
-import { useFormikContext } from 'formik';
-import { canAddHostSNO } from '../../utils';
-import { ClusterDetailsValues } from '../clusterWizard/types';
 
 type SNODisclaimerProps = {
-  isDisabled?: boolean;
   snoSupportLevel: SupportLevel;
+  isDisabled?: boolean;
+  snoExpansionSupported?: boolean;
 };
-const SNODisclaimer = ({ isDisabled = false, snoSupportLevel }: SNODisclaimerProps) => {
-  const { values } = useFormikContext<ClusterDetailsValues>();
+const SNODisclaimer = ({
+  snoSupportLevel,
+  isDisabled = false,
+  snoExpansionSupported = false,
+}: SNODisclaimerProps) => {
   if (!['dev-preview', 'supported'].includes(snoSupportLevel)) {
     //if tech preview or unsupported there's no definition which warning to show
     return null;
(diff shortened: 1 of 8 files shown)


src/ocm/components/AddHosts/utils.ts
```diff
@@ -1,15 +1,32 @@
...
-import { canAddHostSNO } from '../../../common/utils';
 import Day2ClusterService from '../../services/Day2ClusterService';
+import {
+  getFeatureSupportLevelsMap,
+  getVersionSupportLevelsMap,
+} from '../featureSupportLevels/FeatureSupportLevelProvider';
 import { OcmClusterType } from './types';
 
 export const canAddHost = ({ cluster }: { cluster: OcmClusterType }) => {
+  const featureSupportLevelsMap =
```
celdrake, line 10: We could extract all this logic to a separate function. Following our new approach, It could be a `clusterSelector` given it needs the cluster and the feature to check.

`
const isSNOExpansionSupported = hasSupportedFeature(cluster, 'SINGLE_NODE_EXPANSION');
`

src/common/components/clusterConfiguration/SNODisclaimer.tsx
```diff
@@ -2,16 +2,17 @@ import React from 'react';
 import { Alert, AlertVariant, List, ListItem, Stack, StackItem } from '@patternfly/react-core';
 import { CheckboxField } from '../ui';
 import { SupportLevel } from '../../types';
-import { useFormikContext } from 'formik';
-import { canAddHostSNO } from '../../utils';
```
celdrake, line 6: The `canAddHostSNO` function can be completely removed (from `src/comon/utils`)

src/ocm/components/AddHosts/utils.ts
```diff
@@ -1,15 +1,32 @@
...
+  const isSNOExpansionSupported =
+    versionSupportLevels && versionSupportLevels['SINGLE_NODE_EXPANSION'] === 'supported';
+
   if (Day2ClusterService.getOpenshiftClusterId(cluster)) {
     if (cluster.aiCluster) {
       if (isSNO(cluster.aiCluster)) {
         return (
           cluster.aiCluster.status === 'installed' &&
-          canAddHostSNO(cluster.aiCluster?.openshiftVersion || cluster.openshift_version)
+          cluster.aiCluster.openshiftVersion &&
```
celdrake, line 28: Adding this condition makes checking the level support unnecessary when we don't have a `aiCluster.openshiftVersion`, as the results from the validation will be discarded.

I think it can be rewritten as

``` getVersionSupportLevelsMap(
      cluster.aiCluster?.openshiftVersion,
      featureSupportLevelsMap,
    );
```
and there's no need to check the version later.

celdrake (changes requested):

src/ocm/components/AddHosts/utils.ts
jgyselov: The `cluster` here isn't really the cluster we use in clusterSelectors. I'd rather keep it separate.

src/ocm/components/AddHosts/utils.ts
jgyselov: I moved the whole logic elsewhere so hopefully the `canAddHost` function is now cleaner...

src/ocm/components/AddHosts/utils.ts
celdrake: True, that makes sense

src/ocm/components/AddHosts/utils.ts
celdrake: Thanks, I think it's much clearer now :)

celdrake (approved):

jkilzi (approved):

PR #1478, opened 2022-06-28
NO_ISSUE: Temporarily remove translations in Common code

celdrake:
The translations for OCM are currently not being applied. 

While the problem is being fixed, we can roll back to the plain text, only for Common code (affecting OCM).


jgyselov (approved):

celdrake:
@jkilzi Seems like we won't need it, and it can be closed.

ammont82:
With this PR in assisted-ui: https://github.com/openshift-assisted/assisted-ui/pull/573 we resolve the problem with translations in common code of assisted-ui-lib

PR #1481, opened 2022-06-28
fix: infraEnvError is now a string

jkilzi:
eslint warnings dropped to 328


celdrake (approved):
I missed this one, sorry.

Although I would expect this type of error would have been caught by our linter.
Let's see if we can improve it further in the future
ts-error
.

PR #1482, opened 2022-06-29
MGMT-10420: Extra spaces in installation page

jgyselov:
https://issues.redhat.com/browse/MGMT-10420

I have slightly decreased the spacing in and around the cluster progress item and also between the alerts. It's not exactly by half as the design says in the following image but this is the best that can be achieved without overwriting Patternfly variables.

image

### Before and after images:

**Before:**
image

**After:**
image

---

**Before:**
image

**After:**
image

---

**Before:**
image

**After:**
image

### DIFF
--- a/src/common/components/clusterDetail/ProgressBarTexts.tsx
+++ b/src/common/components/clusterDetail/ProgressBarTexts.tsx
@@ -12,45 +12,34 @@ type HostProgressProps = {
 };
 
 export const ProgressBarTexts = ({ hosts, hostRole }: HostProgressProps) => {
-  const filteredHosts = hosts.filter((host) => host.role && hostRole === host.role);
-  const failedHostsCount = filteredHosts.filter((host) => host.status === 'error').length;
-  const hostCountText = (hostRole: HostRole) =>
-    failedHostsCount === 0
-      ? `${filteredHosts.length} ${pluralize(
-          filteredHosts.length,
-          hostRole === 'master' ? 'control plane node' : 'worker',
-        )}`
-      : `${failedHostsCount}/${filteredHosts.length} ${pluralize(
-          filteredHosts.length,
-          hostRole === 'master' ? 'control plane node' : 'worker',
-        )}`;
+  const hostRoleText = pluralize(hosts.length, hostRole === 'master' ? 'Control Plane' : 'Worker');
+  const hostCountText = `${hosts.length} ${pluralize(
+    hosts.length,
+    hostRole === 'master' ? 'control plane node' : 'worker',
+  )}`;
 
-  const getHostName = (hostRole: HostRole): string => {
-    if (hostRole === 'master') {
-      return 'Control Plane';
-    }
-    return 'Workers';
-  };
-
-  if (filteredHosts.some((host) => ['cancelled', 'error'].includes(host.status))) {
+  if (hosts.some((host) => ['cancelled', 'error'].includes(host.status))) {
+    const failedHostsCount = hosts.filter((host) => host.status === 'error').length;
     return (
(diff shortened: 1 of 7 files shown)


nirfarkas:
Looks great! Thanks!

src/common/components/clusterDetail/ProgressBarTexts.tsx
```diff
@@ -25,6 +24,7 @@ export const ProgressBarTexts: React.FC<HostProgressProps> = ({ hosts, hostRole
           filteredHosts.length,
           hostRole === 'master' ? 'control plane node' : 'worker',
         )}`;
+
   const getHostName = (hostRole: HostRole): string => {
```
jkilzi, line 28: Should be declared outside the component

src/common/components/clusterDetail/ProgressBarTexts.tsx
```diff
@@ -1,18 +1,17 @@
...
 type HostProgressProps = {
   hosts: Host[];
   hostRole: HostRole;
 };
 
-export const ProgressBarTexts: React.FC<HostProgressProps> = ({ hosts, hostRole }) => {
+export const ProgressBarTexts = ({ hosts, hostRole }: HostProgressProps) => {
   const filteredHosts = hosts.filter((host) => host.role && hostRole === host.role);
   const failedHostsCount = filteredHosts.filter((host) => host.status === 'error').length;
   const hostCountText = (hostRole: HostRole) =>
```
jkilzi, line 17: Cab be declared outside the component, just pass in the `filteredHosts` and `failedHostsCount` as additional arguments.

jkilzi (changes requested):

src/common/components/clusterDetail/ProgressBarTexts.tsx
jgyselov: It doesn't even need to be a function :grin:

jkilzi (dismissed):

src/common/components/clusterDetail/ProgressBarTexts.tsx
```diff
@@ -12,45 +12,39 @@ type HostProgressProps = {
 };
 
 export const ProgressBarTexts = ({ hosts, hostRole }: HostProgressProps) => {
-  const filteredHosts = hosts.filter((host) => host.role && hostRole === host.role);
-  const failedHostsCount = filteredHosts.filter((host) => host.status === 'error').length;
-  const hostCountText = (hostRole: HostRole) =>
+  const hostRoleText = pluralize(hosts.length, hostRole === 'master' ? 'Control Plane' : 'Worker');
+  const failedHostsCount = hosts.filter((host) => host.status === 'error').length;
+
+  const hostCountText =
```
celdrake, line 18: I can only suggest a small change, to remove some duplicate code.

```
  const hostCountText = `${hosts.length} ${pluralize(
    hosts.length,
    hostRole === 'master' ? 'control plane node' : 'worker',
  )}`
```

then 
```
  if (hosts.some((host) => ['cancelled', 'error'].includes(host.status))) {
    const failedHostsCount = hosts.filter((host) => host.status === 'error').length;
    return (
      <ClusterProgressItem icon={<ExclamationCircleIcon color={dangerColor.value} />}>
        <>
          {hostRoleText}
          <br />
          <small>{failedHostsCount}/{hostCountText} failed</small>
        </>
      </ClusterProgressItem>
    );
  }
```

celdrake (commented):
Good job with the refactor and extracting the new component!
Check my suggestion but it's pretty minor so it can be skiped :)

src/common/components/clusterDetail/ProgressBarTexts.tsx
jgyselov: Changed, thanks for the suggestion!

PR #1488, opened 2022-07-01
MGMT-10953 Show beta versions as unsupported

celdrake:
https://issues.redhat.com/browse/MGMT-10953

If the user selected a "beta" version, we now show a message indicating that the cluster won't be supported due to it.
Added to Review and Create page as well as the OCM card.
The exclamation mark on the version won't be shown in the OCM card because it's rendered from the OCM side, and hidden from Assisted UI.

New design, showing the different warning combinations (same UX for Review and Create and Cluster Details card):

beta-version-only-feature-warning

beta-version-both-warnings

beta-version-only-beta-warning2


src/ocm/components/featureSupportLevels/ReviewClusterFeatureSupportLevels.tsx
```diff
@@ -109,43 +120,59 @@ export const getFeatureSupportLevelTitle = (fullySupported: boolean): string =>
...
+      testId="feature-support-levels"
+    />
+  );
 };
 
-const ClusterFeatureSupportLevelsDetailItem = ({ ...props }: ItemProps) => (
+const ClusterFeatureSupportLevelsDetailItem = ({ ...props }: SupportLevelProps) => (
   <WithErrorBoundary title="Failed to load feature support levels review">
-    <Item {...props}></Item>
+    <SupportLevel {...props}></SupportLevel>
```
jkilzi, line 175:
```suggestion
    <SupportLevel {...props} />
```

jkilzi (changes requested):

src/ocm/hooks/useOpenshiftVersions.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { CpuArchitecture, OpenshiftVersionOptionType } from '../../common';
```
jgyselov, line 1: What is the purpose of this file?

jgyselov (approved):

src/ocm/hooks/useOpenshiftVersions.spec.ts
celdrake: I created some unit tests to validate that the new functions worked fine, and thought it would be great to have them tested.
But right now we don't have the test engine so in reality they wouldn't be executed.

So it's not useful unless we add the test engine back. Maybe we should create a ticket for it?

celdrake: Removed, we'll create a task to better handle this.

src/ocm/hooks/useOpenshiftVersions.tsx
```diff
@@ -23,13 +26,25 @@ const sortVersions = (versions: OpenshiftVersionOptionType[]) => {
     .reverse();
 };
 
+export const findVersionItemByVersion = (
+  versions: OpenshiftVersionOptionType[],
+  version: string,
+): OpenshiftVersionOptionType | undefined => {
+  return versions.find(({ value: versionKey }) => {
+    // For version 4.10 match 4.10, 4.10.3, not 4.1, 4.1.5
+    const versionNameMatch = new RegExp(`^${versionKey}(\\..*)?$`);
```
jkilzi, line 35:
```suggestion
    const versionNameMatch = new RegExp(`^${versionKey}(\\..+)?$`);
```

jkilzi (changes requested):

src/ocm/hooks/useOpenshiftVersions.tsx
celdrake: Fixed

jkilzi (approved):

jkilzi (approved):

PR #1493, opened 2022-07-05
MGMT-10951 Improve Sentry logging for AI errors

celdrake:
https://issues.redhat.com/browse/MGMT-10951

We want to be able to filter clearly the Sentry errors originating from Assisted Installer UI lib.
For that, I added two new tags called `app_section` and `section_version` that should be displayed with the existing ones.
I chose this names based on the existing tags, and in theory they shuld be created when the first event using it happens.
ocm-tags

It looks like we won't be able to test this before Staging, given that the allowed origins blocks us from sending these requests from our local environments.


jkilzi (approved):

celdrake:
Closing, will reopen when we can sync with OCM.

PR #1495, opened 2022-07-06
Add data-testid to remove button tang servers

DanielleChason:
I need that for the tests 
image


jkilzi:
@DanielleChason let's sync about this tomorrow plz

PR #1512, opened 2022-07-15
BZ2090999: Validate BMC protocols for disabled network provision

mareklibra:


src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -470,18 +471,27 @@ export const day2ApiVipValidationSchema = Yup.string().test(
...
 
-export const bmcAddressValidationSchema = Yup.string().matches(BMC_REGEX, {
-  message: 'Value "${value}" is not valid BMC address.', // eslint-disable-line no-template-curly-in-string
-  excludeEmptyString: true,
-});
+export const bmcAddressValidationSchema = (isProvisioningNetworkDisabled: boolean) =>
+  Yup.string()
+    .test(
+      'bmc-address-provisioning-network-disabled',
+      'Since the provisioningnetwork is disabled, the suported protocol is "redfish-virtualmedia://" only.',
```
jkilzi, line 478:
```suggestion
      'Since the provisioning network is disabled, the supported protocol is "redfish-virtualmedia://" only.',
```

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -470,18 +471,27 @@ export const day2ApiVipValidationSchema = Yup.string().test(
...
+      (value: string): boolean =>
+        !isProvisioningNetworkDisabled || !!value?.match(DISABLED_PROVISIONING_NETWORK_BMC_REGEX),
+      // Fall-through to the next more generic REGEX
+    )
+    .matches(BMC_REGEX, {
+      message: 'Value "${value}" is not valid BMC address.', // eslint-disable-line no-template-curly-in-string
```
jkilzi, line 484:
```suggestion
      message: 'Value "${value}" is not a valid BMC address.', // eslint-disable-line no-template-curly-in-string
```

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -470,18 +471,27 @@ export const day2ApiVipValidationSchema = Yup.string().test(
...
-  message: 'Value "${value}" is not valid BMC address.', // eslint-disable-line no-template-curly-in-string
-  excludeEmptyString: true,
-});
+export const bmcAddressValidationSchema = (isProvisioningNetworkDisabled: boolean) =>
+  Yup.string()
+    .test(
+      'bmc-address-provisioning-network-disabled',
+      'Since the provisioningnetwork is disabled, the suported protocol is "redfish-virtualmedia://" only.',
+      (value: string): boolean =>
+        !isProvisioningNetworkDisabled || !!value?.match(DISABLED_PROVISIONING_NETWORK_BMC_REGEX),
```
jkilzi, line 480: Is the optional chaining operator needed here? (in `... !!value?. ...`)
If `value` is always a string then it shouldn't be needed, but if it could be null or undefined then it should, and in this case `value` is not annotated properly inline 479

jkilzi (changes requested):

rawagner:
all new strings need i18n

src/common/components/ui/formik/validationSchemas.ts
mareklibra: So far we have always at least empty string here, but I feel like assuming it here blindly is not a good idea in the long-term, so changing the parameter type.

rawagner (dismissed):

src/common/components/ui/formik/validationSchemas.ts
jkilzi: Marek just to be clear, my question is only about the use of `?.`. I'm completely fine with the `!!`. Although the use of `excludeEmptyString: true` makes it hard to understand whether the empty string should be accepted or not in this case.

mareklibra:
After discussion on slack, reduced to check for either redfish-virtualmedia or idrac-virtualmedia

jtomasek (approved):

rawagner (approved):

jtomasek (approved):

PR #1522, opened 2022-07-19
Do not allow warnings in updated files

celdrake:
Added a new step in the On-pull-request flow that would throw an error if any of the updated files in the PR has linting warnings.
This should help bring down the number of warnings.

### DIFF
--- a/.github/workflows/on-pull-request.yaml
+++ b/.github/workflows/on-pull-request.yaml
@@ -32,6 +32,20 @@ jobs:
         # --frozen-lockfile: doesn't generate a yarn.lock file, fails if an update is needed.
         run: yarn install --frozen-lockfile
 
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v23.2
+        with:
+          files: |
+            src/**/*.ts
+            src/**/*.tsx
+
+      - name: Lint all changed files
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            yarn lint-touched-file $file
+          done
+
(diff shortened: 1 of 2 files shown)


rawagner (approved):

package.json
```diff
@@ -38,6 +38,7 @@
     "test": "run-s test:circular lint format:check",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 ./src/index.ts",
     "lint": "eslint . --color",
+    "lint-touched-file": "eslint --color --max-warnings=0",
```
jkilzi, line 41: In general, I think it will be a good idea to decompose the yarn build script.
Such a change would allow running the circle-deps check and linter in separate steps, and also allow for lining only the files that changed in the PR instead of the entire code base!

.github/workflows/on-pull-request.yaml
```diff
@@ -32,6 +32,20 @@ jobs:
...
+        with:
+          files: |
+            src/**/*.ts
+            src/**/*.tsx
+
+      - name: Lint all changed files
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            yarn lint-touched-file $file
+          done
```
jkilzi, line 47: Please be aware that all these files you run the lint in this step will be lint again in the build step. Please consider decomposing the build step.

jkilzi (changes requested):

jkilzi:
Closing in this one in favor of implementing the previous approach in order to reduce the number of warnings.

rawagner:
The previous approach does not guarantee that the number of warnings will be reduced. Even if there are things that could be improved, this is still miles better that the other PR IMO :/

PR #1528, opened 2022-07-20
MGMT-11107 Keep the NTP tooltip open even after polling

celdrake:
https://issues.redhat.com/browse/MGMT-11107

When the tooltip for NTP was open, as soon as a new polling happened, the tooltip would close.
This was due to the re-rendering of `canEditDisks` which itself changed every time the `cluster` updated.

I changed it so that the actions only change when the required data change.

### DIFF
--- a/src/ocm/components/hosts/HostsDiscoveryTable.tsx
+++ b/src/ocm/components/hosts/HostsDiscoveryTable.tsx
@@ -24,13 +24,14 @@ import {
 } from '../../../common/components/hosts/tableUtils';
 import HostsTable from '../../../common/components/hosts/HostsTable';
 import { HostDetail } from '../../../common/components/hosts/HostRowDetail';
-import { TableRow } from '../../../common/components/hosts/AITable';
+import { ExpandComponentProps, TableRow } from '../../../common/components/hosts/AITable';
 import { ValidationsInfo } from '../../../common/types/hosts';
 import { usePagination } from '../../../common/components/hosts/usePagination';
 import { useTranslation } from '../../../common/hooks/use-translation-wrapper';
 import HostsTableEmptyState from '../hosts/HostsTableEmptyState';
 import HardwareStatus from './HardwareStatus';
 import { AdditionalNTPSourcesDialogToggle } from './AdditionaNTPSourceDialogToggle';
+import { onDiskRoleType } from '../../../common/components/hosts/DiskRole';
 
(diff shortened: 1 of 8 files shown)


src/ocm/components/hosts/HostsDiscoveryTable.tsx
```diff
@@ -109,9 +89,23 @@ const HostsDiscoveryTable: React.FC<HostsDiscoveryTableProps> = ({ cluster }) =>
...
+        <HostDetail
+          key={host.id}
+          host={host}
+          onDiskRole={onDiskRole}
+          canEditDisks={actionChecks.canEditDisks}
+          AdditionalNTPSourcesDialogToggleComponent={AdditionalNTPSourcesDialogToggle}
+        />
+      );
+    };
+  }, [onDiskRole, actionChecks.canEditDisks]);
```
rawagner, line 107: component inside component is an antipattern. It will break React's reconciliation loop.
Pass additional props to `HostDetail` could be done via context

rawagner (changes requested):

src/ocm/components/hosts/HostsDiscoveryTable.tsx
celdrake: I've been looking at this but I've not been able to improve this.
From what I see, the `ExpandComponent` is forcing us to return this builder function which creates the HostDetail component.

I looked at if I could pass the actions from the context (from `useHostsTable`), but for that I'd need to pass the entire `cluster` object and then it would refresh when the cluster does.

I'm doing a slight change from the initial code, please let me know what you think.

rawagner: Just create your own context, wrap the table in it and load the content of the context in ExpandedComponent.

I know it is a bit too complicated but still better than passing all the actions and who knows what else would be needed for other cases

celdrake: It took me a while, but I think I managed to do as you suggested. PTAL.

rawagner: great, that is exactly what I wanted to see

rawagner (approved):

celdrake:
As I was reviewing the new Storage Step, I have realised that the functionality for disk edition etc changes because of the Storage step. So better to check this issue after it is merged, as it's a smaller bug.

PR #1543, opened 2022-07-28
MGMT-11344 iso download timing 2.7

celdrake:
Port of https://github.com/openshift-assisted/assisted-ui-lib/pull/1541

### DIFF
--- a/src/ocm/hooks/useInfraEnvImageUrl.ts
+++ b/src/ocm/hooks/useInfraEnvImageUrl.ts
@@ -1,18 +1,15 @@
 import React from 'react';
 import useInfraEnvId from './useInfraEnvId';
-import { Cluster, PresignedUrl } from '../../common';
+import { Cluster } from '../../common';
 import { InfraEnvsAPI } from '../services/apis';
-import { getErrorMessage } from '../../common/utils';
 
 export default function useInfraEnvImageUrl(clusterId: Cluster['id']) {
-  const { infraEnvId, error: infraEnvIdError } = useInfraEnvId(clusterId);
-  const [imageUrl, setImageUrl] = React.useState<PresignedUrl['url']>();
-  const [error, setError] = React.useState('');
+  const { infraEnvId } = useInfraEnvId(clusterId);
 
-  const getImageUrl = React.useCallback(async () => {
+  const getImageUrl = React.useCallback(async (): Promise<string> => {
     try {
       if (!infraEnvId) {
-        return;
+        return '';
       }
       const {
         data: { url },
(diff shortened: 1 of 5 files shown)


src/ocm/hooks/useInfraEnvImageUrl.ts
```diff
@@ -4,35 +4,30 @@ import { Cluster, PresignedUrl } from '../../common';
...
-  const [imageUrl, setImageUrl] = React.useState<PresignedUrl['url']>();
-  const [error, setError] = React.useState('');
+  const { infraEnvId, error: infraEnvError } = useInfraEnvId(clusterId);
 
-  const getImageUrl = React.useCallback(async () => {
+  const getImageUrl = React.useCallback(async (): Promise<ImgUrl> => {
     try {
       if (!infraEnvId) {
-        return;
+        return { url: '', error: infraEnvError || 'Missing infraEnv' };
```
jkilzi, line 18:
```suggestion
        return { url: '', error: infraEnvError || 'Missing infraEnv' };
```
can we omit the `url` prop?

src/ocm/hooks/useInfraEnvImageUrl.ts
```diff
@@ -4,35 +4,30 @@ import { Cluster, PresignedUrl } from '../../common';
...
       }
       const {
         data: { url },
       } = await InfraEnvsAPI.getImageUrl(infraEnvId);
       if (!url) {
         throw 'Failed to retrieve the image URL, the API returned an invalid URL';
       }
-      setImageUrl(url);
+      return { url, error: '' };
```
jkilzi, line 26:
```suggestion
      return { url, error: '' };
```
can we omit the `error` prop?

src/ocm/hooks/useInfraEnvImageUrl.ts
```diff
@@ -4,35 +4,30 @@ import { Cluster, PresignedUrl } from '../../common';
...
-      setError(getErrorMessage(e));
-    }
-  }, [infraEnvId]);
-
-  React.useEffect(() => {
-    if (infraEnvIdError) {
-      setError(infraEnvIdError);
-    } else if (!imageUrl) {
-      void getImageUrl();
+      return { url: '', error: getErrorMessage(e) };
```
jkilzi, line 28:
```suggestion
      return { url: '', error: getErrorMessage(e) };
```
can we omit the `url` prop here too?

jkilzi (approved):
LGTM + left some comments

src/ocm/hooks/useInfraEnvImageUrl.ts
celdrake: Since we want to return two different things, we need to wrap the response payload in an object.
If we just return an string, we wouldn't know if it's an error or the URL.

PR #1548, opened 2022-07-28
MGMT-11313 Make Popover icon links by default

celdrake:
https://issues.redhat.com/browse/MGMT-11313

There seems to be a difference in the order of the CSS files between development and standalone builds.
Depending on which file is last, a Button using an "a" component will have the lineHeight different from the `.pf-c-button` class or the `.pf-c-form__group-label-help` classes.

Looks like it's an edge case, that can also be fixed by setting `padding:0` using a custom class.  This avoids changing the behaviour of all the PopoverIcon buttons that use links.

### DIFF
--- a/src/ocm/components/clusterConfiguration/ArmCheckbox.tsx
+++ b/src/ocm/components/clusterConfiguration/ArmCheckbox.tsx
@@ -20,6 +20,8 @@ const getLabel = (openshiftVersion: string) => {
     <>
       Use arm64 CPU architecture{' '}
       <PopoverIcon
+        component="a"
+        buttonClassName="small-button-link"
         noVerticalAlign
         bodyContent={
           <p>
--- a/src/ocm/components/clusterConfiguration/ArmCheckbox.tsx
+++ b/src/ocm/components/clusterConfiguration/ArmCheckbox.tsx
@@ -5,7 +5,7 @@ import {
   getFieldId,
   ClusterCreateParams,
   HelperText,
-  PopoverIcon,
+  PopoverHelperIcon,
   CpuArchitecture,
   OpenshiftVersionOptionType,
 } from '../../../common';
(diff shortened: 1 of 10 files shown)


src/ocm/components/clusterConfiguration/ArmCheckbox.tsx
```diff
@@ -20,6 +20,8 @@ const getLabel = (openshiftVersion: string) => {
     <>
       Use arm64 CPU architecture{' '}
       <PopoverIcon
+        component="a"
```
jtomasek, line 23: Why not using `variant="plain"` instead?

Also looking at patternfly examples, I am seeing that they're not using actual `Button` component but rather define a custom `<button>` for form label help popovers. I'd like to avoid this if possible but a solution could be to create a dedicated component for form label help popovers.
Also note that our `PopoverIcon` component does not set `HelpIcon`'s `noVerticalAlign` to true whereas the Patternfly example does that.

jtomasek (changes requested):
It would be really good to avoid having to define css overrides to fix such issues. Let's make sure we're using PF correctly before doing that.

src/ocm/components/clusterConfiguration/ArmCheckbox.tsx
celdrake: I can't find the examples using the `button` element, could you please share the link?
I'm seeing here they are using `Button` component https://www.patternfly.org/v4/components/popover

src/ocm/components/clusterConfiguration/HostInventory.tsx
```diff
@@ -53,15 +53,15 @@ const PlatformIntegrationLabel: React.FC = () => (
           </List>
         </>
       }
-      buttonOuiaId="platform-integration-vSphere-popover"
+      ouiaId="platform-integration-vSphere-popover"
```
celdrake, line 56: @jgyselov I'm assuming Pendo won't have a problem clicking on the element as long as it has the same ID. But it won't be a button anymore (it's what was causing the styling problems)

celdrake:
Sorry, I realised the color is wrong now. The initial color is taken from the button, and it has the hover effect which makes it darker. I'll see how to implement this.

celdrake:
Closing as it didn't properly work and we're looking at a new approach.

PR #1537, opened 2022-07-25
MGMT-9920: New Storage step

DanielleChason:
https://issues.redhat.com/browse/MGMT-9920

I implemented the storage step:
image

Currently, it only contains information about the hosts and their disks. In the future node labeling and disk selection should be added (BE support is needed). The disk section is also removed from the "host discovery" step. 

image

image

In the storage step hosts can be added (delete host, change host name, etc..): 

image

Because of that, there are also the proper validations: 

image

PR #1553, opened 2022-08-01
Feature/storage step

celdrake:
Continuation of https://github.com/openshift-assisted/assisted-ui-lib/pull/1537

https://issues.redhat.com/browse/MGMT-9920

https://user-images.githubusercontent.com/829045/182338161-a2dc8be7-af98-415f-8369-8ad050e449d6.mp4


src/common/api/types.ts
```diff
@@ -1253,10 +1253,8 @@ export interface HostUpdateParams {
    * A string which will be used as Authorization Bearer token to fetch the ignition from ignitionEndpointUrl.
    */
   ignitionEndpointToken?: string;
-  /**
-   * Labels to be added to the corresponding node.
-   */
-  nodeLabels?: NodeLabelParams[];
+  validationsInfo?: string;
+  nodeLabels?: string;
```
jkilzi, line 1257:
```suggestion
  validationsInfo?: string;
  nodeLabels?: string;
```

Where are these fields coming from? Are they present in swagger.yaml as well? I don't see the swagger.yaml modified here...

celdrake: You are right, I don't think it should be included for now.
I reverted it and we can add it with all its required changes it when Node labelling is implemented.

jkilzi (approved):

PR #1554, opened 2022-08-03
Enable seting host installation disk

jtomasek:
Host discovery and host selection wizard steps now allow user to set
installation disk in the disks list on expanded host detail

Fixes BZ2081597

### DIFF
--- a/src/cim/components/Agent/AgentsSelectionTable.tsx
+++ b/src/cim/components/Agent/AgentsSelectionTable.tsx
@@ -22,14 +22,30 @@ import DefaultEmptyState from '../../../common/components/ui/uiState/EmptyState'
 import { usePagination } from '../../../common/components/hosts/usePagination';
 import { useFormikHelpers } from '../../../common/hooks/useFormikHelpers';
 import { useTranslation } from '../../../common/hooks/use-translation-wrapper';
+import { HostDetail } from '../../../common/components/hosts/HostRowDetail';
+import { ExpandComponentProps } from '../../../common/components/hosts/AITable';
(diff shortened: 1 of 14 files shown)


src/cim/components/Agent/AgentsSelectionTable.tsx
```diff
@@ -22,21 +22,38 @@ import DefaultEmptyState from '../../../common/components/ui/uiState/EmptyState'
...
 type AgentsSelectionTableProps = {
   matchingAgents: AgentK8sResource[];
   width?: number;
   onEditRole?: AgentTableActions['onEditRole'];
+  onSetInstallationDiskId?: AgentTableActions['onSetInstallationDiskId'];
   onEditHost?: AgentTableActions['onEditHost'];
   onHostSelect?: VoidFunction;
 };
 
+export const getExpandComponent = (
```
rawagner, line 40: this is anti-pattern as every call of `getExpandComponent` will create a new component which breaks React's reconciliation cycle. To pass the additional props to `HostDetail` we can use context. This is the same as https://github.com/openshift-assisted/assisted-ui-lib/pull/1528#discussion_r927518445

rawagner (changes requested):

src/cim/components/Agent/AgentsSelectionTable.tsx
jtomasek: Done

celdrake (approved):
This new `HostsTableDetailContext` will be useful in other places :)

rawagner (approved):

PR #1556, opened 2022-08-04
Static IP UI form view with vlans generates wrong nmstate yaml

jgyselov:
https://issues.redhat.com/browse/MGMT-11430

If the user chooses to use vlan, we move the `ipv4` and `ipv6` configs under the vlan interface.


jkilzi (approved):

src/ocm/components/clusterConfiguration/staticIp/data/formDataToInfraEnvField.ts
```diff
@@ -84,7 +84,20 @@ const getNmstateObject = (
     interfaces.unshift(getEthernetInterface(REAL_NIC_NAME, realProtocolConfigs));
```
batzionb, line 84: @jgyselov @jkilzi 
IIUC the ips will be added also to the ethernet interface and also to the vlan interface.
From what I saw the expected yaml shouldn't contain the ips in the ethernet interface

PR #1558, opened 2022-08-04
Static IP: form to yaml vlan (#1556) (r/2.7)

jkilzi:
Co-authored-by: Celia Amador <camadorg@redhat.com>
port to 2.7


batzionb:
@jgyselov @celdrake 
IIUC after save and reload of cluster, or going forward and back from the wizard step, incase of vlan the ip addresses might not show. this is because it looks for the ethernet interface to find the ips and not the vlans

jkilzi:
> @jgyselov @celdrake IIUC after save and reload of cluster, or going forward and back from the wizard step, incase of vlan the ip addresses might not show. this is because it looks for the ethernet interface to find the ips and not the vlans

This issue has already been reported here https://issues.redhat.com/browse/MGMT-11529

PR #1567, opened 2022-08-08
MGMT-8776: Feature/operators step

celdrake:
Continuation of https://github.com/openshift-assisted/assisted-ui-lib/pull/1329
https://issues.redhat.com/browse/MGMT-8776

Current status:

- Fixed conflicts after rebasing with Storage step and some small fixes
- Moved the disks-associated warnings to the Storage step.
- Kept previous behaviour if hosts are insufficient, the first step is Host Discovery, not operators.

storage-step


src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -72,7 +76,6 @@ const hostDiscoveryStepValidationsMap: WizardStepValidationMap = {
   cluster: {
     groups: [],
     validationIds: [
-      'sufficient-masters-count',
```
celdrake, line 75: We need to understand why this condition has been removed, IMO if this validation should still be required.

celdrake: I added back, otherwise it would allow us to continue to Networking with an invalid number of hosts.

celdrake:
@nirfarkas PTAL to the final version, specially to the warnings about disks. 
We removed them from Host Discovery and moved them to Storage

ammont82 (approved):

src/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -53,7 +54,10 @@ const staticIpValidationsMap: WizardStepValidationMap = {
     validationIds: [],
   },
   softValidationIds: [],
-};
+});
+
+const staticIpValidationsMap = buildEmptyValidationsMap();
+const operatorsStepValidationsMap = buildEmptyValidationsMap();
```
jtomasek, line 60: There are certain operators related validations which should be IMHO included in operator's step validations map:

Host validations:
https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/api/types.ts#L1279-L1283

Cluster validations:
https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/api/types.ts#L516-L519

jtomasek (changes requested):

src/ocm/components/clusterWizard/wizardTransition.ts
celdrake: Sorry, I missed this.
The problem if we add the validations in this page, is that the users won't know what the actual issue is.

In Host Discovery, they are able to see which requirements are unfulfilled for each host, and they can figure out what they need to do, but not on Operators step.

jkilzi: These cluster and hosts validation will pop out only after the hosts started to show up during the discovery step. 
Therefore we cannot move the validations from the discovery step.

celdrake:
Added the new commit to move to Operators only the first time that the cluster has been created. After that, a cluster with no hosts would move to Host Discovery
https://github.com/openshift-assisted/assisted-ui-lib/pull/1567/commits/d59dadc61902eb68afa3de88c32200a9a820c63e

jkilzi (approved):

PR #1569, opened 2022-08-09
MGMT-10585 - No need DNS for IPv6 on Static network configurations in case Dual Stack

jgyselov:
https://issues.redhat.com/browse/MGMT-10585

**IPv4 form** and the yaml file it generates: [ipv4.txt](https://github.com/openshift-assisted/assisted-ui-lib/files/9288061/ipv4.txt).

image

**Dualstack form** and the yaml file it generates: [dualstack.txt](https://github.com/openshift-assisted/assisted-ui-lib/files/9288084/dualstack.txt).

image


celdrake (approved):
Thanks for the generated YAML examples! Looks good :)

ammont82 (approved):

PR #1583, opened 2022-08-11
MGMT-11577: Take OCP version from OCM when creating day-2 infraenv/cluster

omertuc:
`openshiftVersion` is an optional parameter in the Assisted API when
importing clusters or creating infra-envs.

For importing clusters, it's unused so it doesn't matter too much.

For creating-infraenvs, this parameter is used to decide on the RHCOS
version that will be used during discovery/installation.

In day-2, when the parameter is not provided, the Assisted Service uses
the latest available RHCOS version. While this is not an ideal strategy,
and it should be fixed, it would be even better if the UI could simply
tell the Assisted Service which OCP version is being reported by OCM.
This way the service doesn't have to guess.

We've seen recent reports where when the discovery image is RHCOS 4.11
and the OCP version of the target cluster is 4.8, the installation
doesn't work (see https://issues.redhat.com/browse/MGMT-11577).

This commit makes it so that when the UI creates a day-2 imported
cluster and infra-env, it will now also fill the optional
`openshiftVersion` parameter from whatever OCM reports for that cluster.

This should help the Assisted Installer choose a more fitting RHCOS
image, and increase the chances of the day-2 installation succeeding.


omertuc:
I have not tested this as I don't know how to run/build the UI, see this PR merely as a suggestion and feel free to create a proper PR

jkilzi (approved):

PR #1584, opened 2022-08-15
MGMT-11614: Allow dot to be used in the cluster name

jtomasek:


src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -21,8 +21,8 @@ import { TFunction } from 'i18next';
 const ALPHANUMBERIC_REGEX = /^[a-zA-Z0-9]+$/;
 const NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
 const NAME_CHARS_REGEX = /^[a-z0-9-.]*$/;
-const CLUSTER_NAME_CHARS_REGEX = /^[a-z0-9-]*$/;
-const CLUSTER_NAME_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
+const CLUSTER_NAME_CHARS_REGEX = /^[a-z0-9-.]*$/;
```
jtomasek, line 24: Need to tweak this regex further to match https://github.com/openshift/assisted-service/pull/4270/files#diff-72619caf255df33ccaa77f4a5ea4b9980c41cea8fbbbde96f8c62e9085282195R32
(no multiple dots, no combination of hypen and a dot etc.

rawagner (dismissed):

celdrake (approved):

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -22,8 +22,9 @@ import { TFunction } from 'i18next';
 const ALPHANUMBERIC_REGEX = /^[a-zA-Z0-9]+$/;
 const NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
 const NAME_CHARS_REGEX = /^[a-z0-9-.]*$/;
-const CLUSTER_NAME_CHARS_REGEX = /^[a-z0-9-]*$/;
-const CLUSTER_NAME_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
+const CLUSTER_NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
+// NOTE: based on https://github.com/openshift/assisted-service/blob/master/internal/cluster/validations/validations.go#L32
+const CLUSTER_NAME_REGEX = /^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)*$/;
```
jgyselov, line 27: The problem with this regex, while it sure is smarter and perhaps more efficient, the resulting validation messages can get confusing. Like in this case the users will be questioning why is the second validation failing, what is wrong with the characters they used?

image

I suggest we keep the conditions just as separate as the messages. If the message is about allowed characters, the validation test shouldn't be looking at anything else. If the message is about the first and last character, again the test should not be looking at what's in the middle.

jgyselov (changes requested):

src/common/components/ui/formik/validationSchemas.ts
jtomasek: I've merged the two conditions into one to be on the safe side and make the validation compatible with the backend. (https://github.com/openshift-assisted/assisted-ui-lib/pull/1584/commits/6336a16002aef6c54c6133028f3b552c2160f992)
Screenshot 2022-09-20 at 15 22 48

rawagner (dismissed):

src/common/components/ui/formik/validationSchemas.ts
jgyselov: I'm still afraid it might confuse some users. The conditions for the cluster name are complicated, but smashing them into one sentence will not help the users.

image

The conditions for the cluster name are complicated, but smashing them into one sentence will not help the users.

I think we should have something similar for the cluster name as we have for the host names.

image

jtomasek: I understand there are UX concerns, however there is IMHO also a valid concern of a correctness of the separate condition validations. Those are not working correctly at the moment even for the hostnames (putting multiple dots for example) so the individual regexes need to be updated and ensured that they are correct. Otherwise we will hit bugs such as the one I am fixing with this PR. Using the same regex on the UI side as on the backend makes us sure there won't be such bugs.

Anyway, I'll try to craft a regex that resembles the backend one as much as possible in functionality.

jtomasek: I've changed the `CLUSTER_NAME_REGEX` to be based off of the backend regex and tweaked it so the validation does not fail for first and last characters which is covered by the separate condition.
Screenshot 2022-09-20 at 17 29 21
Screenshot 2022-09-20 at 17 29 35
Screenshot 2022-09-20 at 17 29 58

jgyselov (approved):

rawagner (approved):

PR #1589, opened 2022-08-17
MGMT-11434: Add an inline notification when all hosts originated from the same platform

ammont82:
Related to https://issues.redhat.com/browse/MGMT-11434

This is the notification where all the hosts are originated from vsphere:
image

### DIFF
--- a/src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
+++ b/src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
@@ -53,7 +53,7 @@ const ValidationSection = ({
           isInline={true}
           data-testid="discover-vsphere-hosts"
         >
-          Since all of your hosts originated from vSphere platform,you now have the option to
+          Since all of your hosts originated from vSphere platform, you now have the option to
           integrate with vSphere. Switch the 'Integrate with vSphere' toggle to get started.
         </Alert>
       )}
--- a/src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
+++ b/src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
@@ -33,6 +33,7 @@ const ValidationSection = ({
   const { currentStepId } = useClusterWizardContext();
   const validationsInfo = cluster && selectClusterValidationsInfo(cluster);
   const { t } = useTranslation();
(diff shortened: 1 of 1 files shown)


celdrake:
@ammont82 Could you please add a screenshot with the result?

src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
```diff
@@ -33,6 +33,7 @@ const ValidationSection = ({
   const { currentStepId } = useClusterWizardContext();
   const validationsInfo = cluster && selectClusterValidationsInfo(cluster);
   const { t } = useTranslation();
+  const allHostsAreVsphere = cluster && cluster.platform?.type === 'vsphere';
```
celdrake, line 36: So, in reality it's the cluster that is determined to be "vsphere", not the hosts?
If one host was not in vsphere, would the platform type still be `vsphere`?

ammont82: Is confusing because I only find "vsphere" platform type in clusters. In hosts I can't find this information

ammont82: I've sended the code to a guy that have one vsphere local environment to test it

src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
```diff
@@ -45,6 +46,17 @@ const ValidationSection = ({
...
         </Alert>
       )}
+      {allHostsAreVsphere && (
+        <Alert
+          title="Discover the full potential of vSphere integration"
+          variant={AlertVariant.info}
+          isInline={true}
+          data-testid="discover-vsphere-hosts"
+        >
+          Since all of your hosts originated from vSphere platform,you now have the option to
```
celdrake, line 56:
```suggestion
          Since all of your hosts originated from vSphere platform, you now have the option to
```

celdrake (changes requested):
According to what we discussed, the condition is OK.
I just noticed a missing space in the screenshot, please add it

Also I think the quotes in "Integrate with vsphere" should be double not single?

celdrake (approved):

ammont82:
Replaced by https://github.com/openshift-assisted/assisted-ui-lib/pull/1596

PR #1592, opened 2022-08-18
MGMT-10110 Readonly cluster

celdrake:
https://issues.redhat.com/browse/MGMT-10110

### Description
The UI now has a mode for users that are not allowed to edit a cluster, to be able to see it and not update it.
They can go through the wizard, even when validations are failing on a previous step.

### Related PRs:

**Assisted UI**: Allows to set an environment variable to make the Assisted UI application be in read-only mode.
https://github.com/openshift-assisted/assisted-ui/pull/569

**Openshift Assisted UI**: Includes the adaptations to the existing Read-only Integration tests.
https://github.com/openshift-assisted/openshift-assisted-ui/pull/18

**UHC Portal**: OCM sends the cluster permissions so that the UI can determine if the cluster is in edition or viewer mode.
https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/3398

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "prepare": "install-peers",
     "test": "run-s test:circular lint format:check",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 ./src/index.ts",
-    "lint": "eslint . --max-warnings=303 --color",
+    "lint": "eslint . --max-warnings=293 --color",
     "lint:fix": "yarn run lint --fix",
     "format": "prettier --write '**/*.{json,md,scss,yaml,yml}'",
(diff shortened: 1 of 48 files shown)


ammont82 (approved):

ammont82 (dismissed):

ammont82 (commented):
I've found problems in  LVM operator (is a new opeator added in https://github.com/openshift-assisted/assisted-ui-lib/pull/1629).
Needs to be changed to OcmCheckboxField to be disabled in readonly clusters.
image

jkilzi (approved):

ammont82 (approved):

PR #1612, opened 2022-08-30
MGMT-11800 - Adding a cluster tag for SaaS

jgyselov:
https://issues.redhat.com/browse/MGMT-11800

Adds a `ui_ocm` tag to the cluster create/update/install parameters if it's not already there when using OCM.

### DIFF
--- a/src/ocm/components/clusterWizard/ClusterDetails.tsx
+++ b/src/ocm/components/clusterWizard/ClusterDetails.tsx
@@ -49,16 +49,18 @@ const ClusterDetails: React.FC<ClusterDetailsProps> = ({ cluster, infraEnv }) =>
         'openshiftVersion',
       ]);
       try {
-        const cluster = await ClusterDetailsService.update(clusterId, params);
-        dispatch(updateCluster(cluster));
-        canNextClusterDetails({ cluster }) && clusterWizardContext.moveNext();
+        if (cluster) {
+          const updatedCluster = await ClusterDetailsService.update(cluster, params);
+          dispatch(updateCluster(updatedCluster));
+          canNextClusterDetails({ cluster: updatedCluster }) && clusterWizardContext.moveNext();
+        }
       } catch (e) {
         handleApiError(e, () =>
           addAlert({ title: 'Failed to update the cluster', message: getApiErrorMessage(e) }),
         );
       }
     },
-    [addAlert, clearAlerts, dispatch, clusterWizardContext],
+    [clearAlerts, cluster, dispatch, clusterWizardContext, addAlert],
   );
 
   const handleClusterCreate = React.useCallback(
--- a/src/ocm/components/clusterWizard/ClusterDetails.tsx
+++ b/src/ocm/components/clusterWizard/ClusterDetails.tsx
@@ -41,7 +41,7 @@ const ClusterDetails: React.FC<ClusterDetailsProps> = ({ cluster, infraEnv }) =>
   }, [errorOCPVersions, addAlert]);
 
   const handleClusterUpdate = React.useCallback(
-    async (clusterId: Cluster['id'], values: V2ClusterUpdateParams) => {
+    async (cluster: Cluster, values: V2ClusterUpdateParams) => {
       clearAlerts();
       const params: V2ClusterUpdateParams = omit(values, [
         'highAvailabilityMode',
@@ -60,7 +60,7 @@ const ClusterDetails: React.FC<ClusterDetailsProps> = ({ cluster, infraEnv }) =>
         );
       }
     },
-    [clearAlerts, cluster, dispatch, clusterWizardContext, addAlert],
+    [clearAlerts, dispatch, clusterWizardContext, addAlert],
   );
 
   const handleClusterCreate = React.useCallback(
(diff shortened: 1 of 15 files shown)


src/ocm/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -49,16 +49,18 @@ const ClusterDetails: React.FC<ClusterDetailsProps> = ({ cluster, infraEnv }) =>
...
+          canNextClusterDetails({ cluster: updatedCluster }) && clusterWizardContext.moveNext();
+        }
       } catch (e) {
         handleApiError(e, () =>
           addAlert({ title: 'Failed to update the cluster', message: getApiErrorMessage(e) }),
         );
       }
     },
-    [addAlert, clearAlerts, dispatch, clusterWizardContext],
+    [clearAlerts, cluster, dispatch, clusterWizardContext, addAlert],
```
jkilzi, line 63: Including `cluster` here is a good idea only if it is memoized upfront, otherwise its reference will change with every polling iteration

jgyselov: Sending `cluster` as a function argument instead.

src/ocm/components/hosts/use-hosts-table.tsx
```diff
@@ -161,23 +160,23 @@ export const useHostsTable = (cluster: Cluster) => {
...
+        const { data } = await ClustersService.update(cluster, {
           apiVipDnsName: apiVip,
         });
         dispatch(updateCluster(data));
       } catch (e) {
         handleApiError(e, () => onError(getApiErrorMessage(e)));
       }
     },
-    [cluster.id, dispatch],
+    [cluster, dispatch],
```
celdrake, line 171: I don't like that now we have a dependency on the entire `cluster` object, which means that on each polling, the effects will have to be recalculated. This could affect performance.

Also, I see a problem that now the UI is in charge of not overwriting tags that we didn't create, and in theory we could run into race conditions were someone is setting a tag while someone else is on the UI, and some could be lost.

And it can happen the other way around, the users could do some changes on the UI, but later they decide to tag them via the cli with an automated script for example, and the UI tag could be lost.

So I'd prefer an API which lets add "add/remove" a tag, rather than patching the tags to "set" them.

src/ocm/services/ClustersService.ts
```diff
@@ -22,6 +23,27 @@ const ClustersService = {
...
+    }
+    return ClustersAPI.update(cluster.id, params);
+  },
+
+  updateClusterTags(
+    clusterTags: Cluster['tags'],
+    params: V2ClusterUpdateParams,
+  ): V2ClusterUpdateParams {
+    const tags = clusterTags?.split(',') || [];
+    if (tags?.includes(AI_UI_TAG)) {
```
celdrake, line 39: `tags` at this point would always be an array, no need for the `?` which makes the code harder to read.

src/ocm/services/ClustersService.ts
```diff
@@ -22,6 +23,27 @@ const ClustersService = {
...
+      params.tags = clusterTags;
+    } else {
+      tags?.push(AI_UI_TAG);
+      params.tags = tags?.join(',');
```
celdrake, line 43: Maybe we could only PATCH sending the `tags` property if the `AI_UI_TAG` is absent.
That would reduce the probability of clashes.

celdrake (changes requested):

celdrake:
@jkilzi Just a note, that a cluster created on the UI is not guaranteed to have the `ui_ocm` tag later.
It can be updated at any moment using the API and the tag can be removed without the UI being aware of it.

src/ocm/components/clusterWizard/ClusterDetails.tsx
```diff
@@ -41,24 +41,26 @@ const ClusterDetails: React.FC<ClusterDetailsProps> = ({ cluster, infraEnv }) =>
...
       const params: V2ClusterUpdateParams = omit(values, [
         'highAvailabilityMode',
         'pullSecret',
         'openshiftVersion',
       ]);
       try {
-        const cluster = await ClusterDetailsService.update(clusterId, params);
-        dispatch(updateCluster(cluster));
-        canNextClusterDetails({ cluster }) && clusterWizardContext.moveNext();
+        if (cluster) {
```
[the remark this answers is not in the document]
jgyselov: Hasty fix, `if` is not needed.

src/ocm/services/ClustersService.ts
jgyselov: Good idea :+1:

src/ocm/components/hosts/use-hosts-table.tsx
celdrake: @jgyselov I understand now that we are aware of this limitation from the API and that we are OK with it.
Also, speaking to @jkilzi he mentioned that through tools like Kibana we would have the entire history of requests for a specific server so that would identify 100% of the clusters that at some point sent the OCM_UI tag.

src/ocm/services/ClustersService.ts
celdrake: @jgyselov I can't see the change on this, is it pushed?

jgyselov: No, I didn't push the changes yet. I'll do that once I add the tags on install.

celdrake (approved):
Good work!

ammont82 (approved):

ammont82 (commented):
LGTM

PR #1616, opened 2022-08-30
MGMT-11340 - Mass host rename - UI shows localhost4 as valid name

jgyselov:
https://issues.redhat.com/browse/MGMT-11340

`localhost4` and `localhost6` are forbidden host names. The UI will show an error and prevent such name change.

**Before:**
image

**After:**
image

### DIFF
--- a/src/common/components/hosts/MassChangeHostnameModal.tsx
+++ b/src/common/components/hosts/MassChangeHostnameModal.tsx
@@ -31,11 +31,11 @@ import {
 import { Host } from '../../api';
 import { getHostname as getHostnameUtils, getInventory } from './utils';
 import { ActionCheck } from './types';
-import { getErrorMessage } from '../../utils';
 import { useTranslation } from '../../hooks/use-translation-wrapper';
 
 import './MassChangeHostnameModal.css';
 import { TFunction } from 'i18next';
+import { getApiErrorMessage } from '../../../ocm/api';
 
 const getHostname = (host: Host) => {
   const inventory = getInventory(host);
@@ -311,7 +311,7 @@ const MassChangeHostnameModal: React.FC<MassChangeHostnameModalProps> = ({
             formikActions.setStatus({
               error: {
                 title: t('ai:Failed to update host'),
-                message: getErrorMessage(e),
+                message: getApiErrorMessage(e),
               },
             });
           }
--- a/src/common/components/hosts/MassChangeHostnameModal.tsx
+++ b/src/common/components/hosts/MassChangeHostnameModal.tsx
@@ -81,6 +81,17 @@ const validationSchema = (
     hostname: richNameValidationSchema(t, usedHostnames, initialValues.hostname).required(),
   });
 
+const updateHostnameValidationResult = (
+  validationResult: { [key: string]: string[] } | undefined,
+  message: string,
+) => {
+  validationResult = {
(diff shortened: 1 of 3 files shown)


omertuc:
If the backend gives an error for one of the hostnames in the batch rename, the UI should probably display that backend error to the user and say for which hostname the issue occurred, this could prevent the undetailed error we had previously:

image

Even without the UI remembering to cross check the blacklist during batch rename (which is a good change, regardless)

src/common/components/hosts/MassChangeHostnameModal.tsx
```diff
@@ -114,6 +117,16 @@ const withTemplate =
         ),
       };
     }
+
+    if (newHostnames.some((hostname) => FORBIDDEN_HOSTNAMES.includes(hostname || ''))) {
```
celdrake, line 121: I'd like to avoid the code duplication from lines 109 to 128.

Maybe we could extract the function that updates the `validationResult` as all it needs to know is the type of error. Or some other way, but the pattern is the same for the 2 cases.

celdrake (changes requested):

src/common/components/hosts/MassChangeHostnameModal.tsx
celdrake: Also, I can enter `localhost{{n}}` and it tells me it's forbidden, but if I enter `localhost.localdomain{{n}}` it won't.
Shouldn't either both be accepted or rejected?

jgyselov: `localhost.localdomain{{n}}` isn't forbidden. We only forbid hostnames that resolve to loopback addresses (see your `/etc/hosts` file for that. If you put in `localhost{{n}}.localdomain{{n}}`, that would be rejected.

celdrake: Oh, then what happens is that `localhost{{n}}` is rejected because it generates `localhost4` which is forbidden.
Thanks for the explanation!

celdrake (approved):

jgyselov:
I've updated the PR to show the BE error reason if we have it. Ideally, of course, the UI validations should catch all invalid hostnames well before we send them to the BE but @omertuc is right, if errors happen, the message shouldn't be this generic.

Illustrative screenshot with the updated error message and localhost validations turned off:
image

src/common/components/hosts/MassChangeHostnameModal.tsx
```diff
@@ -291,7 +312,7 @@ const MassChangeHostnameModal: React.FC<MassChangeHostnameModalProps> = ({
             formikActions.setStatus({
               error: {
                 title: t('ai:Failed to update host'),
-                message: getErrorMessage(e),
+                message: isApiError(e) ? e?.response?.data?.reason : getErrorMessage(e),
```
celdrake, line 315: Could we use `getApiErrorMessage` which does pretty much the same?
(Unless it doesn't return the proper message if `error.response?.data?.message` is set)

jgyselov: Good idea, I missed that function.

celdrake (approved):

ammont82 (approved):

PR #1620, opened 2022-09-02
MGMT-11814 Display FieldArray errors with getRichTextValidation

celdrake:
When using `validate={getRichTextValidation}` in a form, messages from FielArray were not displaying correctly. 
The problem is that `getRichTextValidation` generates an object where the paths are the individual fields, whereas `validate` is looking for the format of the parent field.

For example, for `diskEncryptionTangServers`, it needs the structure:

```
{diskEncryptionTangServers: [
{ url: 'Url error for Tang Server', thumbprint: 'Thumbprint error for Tang Server' }
]}
```
but it is seeing

`{ 'diskEncryptionTangServers[0].thumbprint': 'Some error}`

Now we are detecting the subfields and generating the needed format from the `diskEncryptionTangServers[0].thumbPrint` etc errors.

The PR also includes some clean-up of the Disk Encryption and other components.

### DIFF
--- a/src/common/components/ui/formik/utils.ts
+++ b/src/common/components/ui/formik/utils.ts
@@ -1,8 +1,9 @@
 import { FormikErrors, FormikTouched } from 'formik';
 import * as Yup from 'yup';
+import { Address4 } from 'ip-address';
+import set from 'lodash/set';
 import { OpenshiftVersionOptionType } from '../../../types';
 import { ClusterNetwork, MachineNetwork, ServiceNetwork } from '../../../api';
-import { Address4 } from 'ip-address';
 
 export const getFieldId = (fieldName: string, fieldType: string, unique?: string) => {
   unique = unique ? `${unique}-` : '';
@@ -31,14 +32,6 @@ export const getFormikErrorFields = <FormikValues>(
 export const getDefaultOpenShiftVersion = (versions: OpenshiftVersionOptionType[]) =>
   versions.find((v) => v.default)?.value || versions[0]?.value || '';
 
-export const labelsToArray = (labels: { [key in string]: string } = {}): string[] => {
-  const result: string[] = [];
-  for (const key in labels) {
-    result.push(`${key}=${labels[key]}`);
-  }
-  return result;
-};
-
 // strValues: array of 'key=value' items
 export const parseStringLabels = (
   strValues: string[],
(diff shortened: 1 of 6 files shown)


celdrake:
@rawagner @jtomasek I checked on CIM, and there could be one form which could be impacted. It's the `BMCForm`.
Could you check that it behaves correctly there?

Thanks!

ammont82 (dismissed):

src/ocm/components/clusterConfiguration/OcmClusterDetailsFormFields.tsx
```diff
@@ -49,31 +49,33 @@ const BaseDnsHelperText: React.FC<{ name?: string; baseDnsDomain?: string }> = (
...
     nameInputRef.current?.focus();
   }, []);
-  const atListOneDiskEncryptionEnableOn =
-    values.enableDiskEncryptionOnMasters || values.enableDiskEncryptionOnWorkers;
 
   const { t } = useTranslation();
-  // TODO(mlibra): Disable fields based on props passed from the caller context. In CIM, the name or domain can not be edited.
   return (
     <Form id="wizard-cluster-details__form">
+      <DiskEncryptionControlGroup
```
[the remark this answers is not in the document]
celdrake: My bad, that was for quicker testing but shouldn't be on the PR. Thanks for spotting it!

ammont82 (dismissed):

jgyselov (dismissed):
Approving but let's make sure we get a LGTM from CIM before we merge.

src/common/components/ui/formik/utils.ts
```diff
@@ -53,32 +46,57 @@ export const parseStringLabels = (
   return labels;
 };
 
-// Input: ['foo=bar', 'key=value', 'foo=blee']
-// Result: ['key=value', 'foo=blee']
-export const uniqueLabels = (labelPairs: string[]): string[] =>
-  labelsToArray(parseStringLabels(labelPairs));
+type innerError = { path: string; message: string };
+type innerErrors = { inner: innerError[] };
+type fieldErrors = { [key: string]: string[] };
```
rawagner, line 51: types should start with capital letter

src/common/components/ui/formik/utils.ts
```diff
@@ -53,32 +46,57 @@ export const parseStringLabels = (
...
-  labelPairs.filter((pair) => allowedKeys.includes(pair.split('=')[0]));
+const fieldErrorReducer = (errors: innerError[]): fieldErrors => {
+  return errors.reduce(
+    (memo, { path, message }) => ({
+      ...memo,
+      [path]: (memo[path] || []).concat(message), // eslint-disable-line
+    }),
+    {},
+  );
+};
```
rawagner, line 61: we can remove the `// eslint-disable-line`
```suggestion
const fieldErrorReducer = (errors: innerError[]): fieldErrors => {
  return errors.reduce<fieldErrors>(
    (memo, { path, message }) => ({
      ...memo,
      [path]: (memo[path] || []).concat(message),
    }),
    {},
  );
};
```

rawagner (changes requested):

src/common/components/ui/formik/utils.ts
celdrake: Fixed, thanks.
Will check if we can enforce it using `https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md`

src/common/components/ui/formik/utils.ts
celdrake: Done

jgyselov (approved):

rawagner (approved):

PR #1621, opened 2022-09-03
MGMT-11313: text is not aligned

jkilzi:
Fixes:
- [MGMT-11313](https://issues.redhat.com/browse/MGMT-11313)
- [MGMT-11801](https://issues.redhat.com/browse/MGMT-11801)

### DIFF
--- a/src/common/components/ui/PopoverIcon.tsx
+++ b/src/common/components/ui/PopoverIcon.tsx
@@ -25,11 +25,10 @@ const PopoverIcon: React.FC<PopoverIconProps> = ({
 }) => (
   <Popover {...props}>
     <Button
-      style={{ padding: 0 }}
       component={component}
       variant={variant}
       onClick={(e) => e.preventDefault()}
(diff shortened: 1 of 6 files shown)


ammont82 (approved):

celdrake:
@jkilzi It doesn't appear to look the same for all of the Popover icons. Also, I don't know if in the ACM side some icons could be affected.
icon-padding-result

src/common/components/ui/PopoverIcon.tsx
```diff
@@ -25,6 +25,7 @@ const PopoverIcon: React.FC<PopoverIconProps> = ({
 }) => (
   <Popover {...props}>
     <Button
+      style={{ padding: 0 }}
```
celdrake, line 28: Maybe we should only set top/bottom padding and leave the left/right padding as is.
See the attached screenshot.

ammont82 (approved):

celdrake (approved):

PR #1629, opened 2022-09-09
MGMT-11005 LVM operator

celdrake:
https://issues.redhat.com/browse/MGMT-11005

### Description

Added a new Operator, for Openshift Data Foundation Logical Volume Manager (LVM)
This operator is used instead of ODF, for Single Node clusters.

Support for this feature as returned by latest API:
- Openshift versions < 4.11, unsupported
- Openshift versions 4.11.x, dev preview.  UI must prevent users from selecting CNV + LVM on Single Node clusters, as they are incompatible. 
- Openshift versions >=4.12, supported, UI must take into account that selecting CNV will automatically add LVM, and it cannot be unselected.
 
### Images and videos

1. LVM unsupported for SNO using v4.10
lvm-4 10-unsupported-improved

2. ODF shown instead of LVM for Multi-node clusters
lvm-multinode-odf

3. LVM for Openshift v4.11 (tech-preview, not compatible with CNV)
https://user-images.githubusercontent.com/829045/189357447-651633c6-8e55-4e9b-8c35-2d98dba96c5b.mp4

4. LVM for Openshift v4.12 (Adding CNV should automatically install LVM, simulated to display UI effect)
https://user-images.githubusercontent.com/829045/189357578-bfc4c6f3-91b0-44b0-8961-80874b2a0e73.mp4

5. LVM requirement validations
https://user-images.githubusercontent.com/829045/189359128-67dea893-d20d-440d-b423-44eab53a78cb.mp4


celdrake:
@jkilzi  I have a question regarding the feature flags. 
I made LVM work the same way as the other operators, except that I don't know if we need to activate it from `uhc-portal`.
See https://gitlab.cee.redhat.com/service/uhc-portal/-/blob/master/src/redux/actions/featureActions.js

nirfarkas:
I think the tooltip for the disabled LVMO should be rephrased to:

“OpenShift Data Foundation Logical Volume
Manager is enabled only for OpenShift 4.11 and above.”

ammont82 (dismissed):

celdrake:
> I think the tooltip for the disabled LVMO when the OCP version is not compatible should be rephrased to:
> 
> “OpenShift Data Foundation Logical Volume Manager is enabled only for OpenShift 4.11 and above.”

Done, you can check the updated image

celdrake:
> @jkilzi I have a question regarding the feature flags. I made LVM work the same way as the other operators, except that I don't know if we need to activate it from `uhc-portal`. See https://gitlab.cee.redhat.com/service/uhc-portal/-/blob/master/src/redux/actions/featureActions.js

As discussed, I have removed the feature flag in UI-lib for LVM. So no need to add it to Uhc-portal too.
https://github.com/openshift-assisted/assisted-ui-lib/pull/1629/commits/670553de3f6bd42c70bd621bcc92397ace65c937

ammont82 (approved):

PR #1632, opened 2022-09-12
Use host/bmh status to determine infra env hosts tab warning

jtomasek:
Fixes BZ2126047


src/cim/components/InfraEnv/InfraEnvHostsTabAgentsWarning.tsx
```diff
@@ -1,28 +1,54 @@
...
+  const bmhsWithoutAgents = infraBMHs.filter((bmh) =>
+    agentsBMHNames.includes(bmh.metadata?.name || ''),
   );
+
+  const agentStates = infraAgents.map((agent) => getAgentStatus(agent, true).status);
+  const bmhStates = bmhsWithoutAgents.map((bmh) => getBMHStatus(bmh).state);
+
+  const hostStates = [...agentStates, ...bmhStates];
+
+  const problemStates = [
```
rawagner, line 32: how about adding `type` to `HostStatusDef` ? So we have all these in one place

jtomasek: Quite often the list differs based on the specific use case. In other words, there will be cases when the list of states belonging to a certain status type will differ and the type property on the HostStatusDef would not be correct. So I am inclined to keeping the list just here.

rawagner (approved):

ammont82 (approved):

PR #1634, opened 2022-09-13
MGMT-11687 - Static IP form view generates vlan nmstate yaml with unnecessary ethernet interface

jgyselov:
https://issues.redhat.com/browse/MGMT-11687

Generated yaml examples:
[old_nmstate.txt](https://github.com/openshift-assisted/assisted-ui-lib/files/9556402/old_nmstate.txt)
[new_nmstate.txt](https://github.com/openshift-assisted/assisted-ui-lib/files/9556403/new_nmstate.txt)


batzionb:
@jgyselov 
A change in the serialization (frontend -> backend) in file [formDataToInfraEnvField.ts](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/ocm/components/clusterConfiguration/staticIp/data/formDataToInfraEnvField.ts)
Might require a change in the deserialization (backend -> frontend) in the file [formDataFromInfraEnvField.ts](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/ocm/components/clusterConfiguration/staticIp/data/formDataFromInfraEnvField.ts)

src/ocm/components/clusterConfiguration/staticIp/data/formDataToInfraEnvField.ts
```diff
@@ -74,36 +76,44 @@ const getNmstateObject = (
...
-    });
+    if (networkWide.useVlan && networkWide.vlanId) {
+      const vlanInterface = getVlanInterface(
+        REAL_NIC_NAME,
+        networkWide.vlanId,
+        realProtocolConfigs,
+      );
+
+      interfaces.unshift(vlanInterface);
+      routeConfigs.forEach((route) => {
```
batzionb, line 103: updating routeConfigs for vlan is not required anymore, what's done in line 90 is already correct

src/ocm/components/clusterConfiguration/staticIp/data/formDataToInfraEnvField.ts
```diff
@@ -74,36 +76,44 @@ const getNmstateObject = (
         ),
       };
       nicName = getDummyNicName(protocolVersion);
-      interfaces.push(getEthernetInterface(nicName, protocolConfigs));
+
+      if (networkWide.useVlan && networkWide.vlanId) {
```
batzionb, line 80: I suggest a function
getInterface, that gets the name, protocolConfigs and networkWide.
It will return EthernetInterface | VlanInterface (actually these types can be combined now to type Interface with optional vlan section)
It should be reused in line 95
This will make this function much more readable, it's becoming hard to maintain

batzionb (changes requested):

src/ocm/components/clusterConfiguration/staticIp/data/formDataToInfraEnvField.ts
```diff
@@ -74,36 +75,28 @@ const getNmstateObject = (
...
-      interfaces.push(getEthernetInterface(nicName, protocolConfigs));
+      interfaces.push(getInterface(nicName, protocolConfigs, networkWide));
     }
+
     routeConfigs.push(
-      getRouteConfig(protocolVersion, networkWide.ipConfigs[protocolVersion].gateway, nicName),
+      getRouteConfig(
+        protocolVersion,
+        networkWide.ipConfigs[protocolVersion].gateway,
+        networkWide.useVlan && networkWide.vlanId ? `${nicName}.${networkWide.vlanId}` : nicName,
```
batzionb, line 85: One last polish? :)
The code to generate the vlan name repeats itself also in the function getInterface
Would be nice to have it in a function, something like getVlanNicName

batzionb (dismissed):

src/ocm/components/clusterConfiguration/staticIp/data/formDataToInfraEnvField.ts
jgyselov: Done :slightly_smiling_face:

batzionb (approved):

celdrake (approved):
Tested and I can see that the generated format is correct :100:

PR #1636, opened 2022-09-13
MGMT-11710 - "We lost that page" displays when BE is unavailable

jgyselov:
https://issues.redhat.com/browse/MGMT-11710

No redirect will happen and a generic error will be shown.

https://user-images.githubusercontent.com/87187179/189924125-07c643b7-4043-48a7-9a88-3aadb9c7d8f3.mp4

### DIFF
--- a/src/ocm/components/clusters/ClusterPage.tsx
+++ b/src/ocm/components/clusters/ClusterPage.tsx
@@ -180,7 +180,11 @@ const ClusterPage: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
     );
   }
 
-  return <Redirect to="/clusters" />;
+  return (
+    <PageSection variant={PageSectionVariants.light} isFilled>
+      <ErrorState fetchData={fetchCluster} actions={errorStateActions} />
+    </PageSection>
+  );
 };
 
 export default ClusterPage;


jkilzi:
@jgyselov can please show how this would look inside the uhc-portal? (I just want to verify it won't redirect to a different page when the API is down). 
Also, @lalon4 asked for an automatic retry mechanism. Can we do that?

jgyselov:
> @jgyselov can please show how this would look inside the uhc-portal? (I just want to verify it won't redirect to a different page when the API is down).

What you see in the video _in my original comment_ is UHC portal, but I suppose you mean when viewed through the cluster detail going from the OCM clusters list? Currently, an error there makes the AI card disappear but doesn't redirect anywhere. We can either keep it that way, or we could display an error as well.

https://user-images.githubusercontent.com/87187179/189936174-a6183200-66c6-495a-bae8-2352b3bb8253.mp4

> Also, @lalon4 asked for an automatic retry mechanism. Can we do that?

There is a **try again** button to fetch the cluster+hosts in the error component and the cluster polling is still running. Is that enough?

ammont82 (approved):

celdrake:
> There is a **try again** button to fetch the cluster+hosts in the error component and the cluster polling is still running. Is that enough?

I don't think that would enough, because instead of going to a different page, now Cypress will see the error message. 
But it wouldn't know if this is a one-off, and would make the tests less consistent.

I think we could see if we can retry the request twice, and only if it fails twice, we'd show the error message.

jgyselov:
We closed the Jira issue as duplicate so I'm closing this PR as well.

PR #1661, opened 2022-09-27
Display error message if errors are instance of Error

guillaumevincent:
In try catch block, error.message is valid only of error object is instance of Error. This patch fixes some error we may have with Typescript 4.


rawagner:
Thanks @guillaumevincent ! We have this utility func https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/utils.ts#L3 which would be great to use everywhere..we dont do that yet obviously :) Can you update your changes to use the `getErrorMessage` ?

guillaumevincent:
Sure, let me do this

PR #1668, opened 2022-09-30
WIP MGMT-11268 nutanix integration

celdrake:
WIP - needs to send the correct data when the platform integration is selected.


src/ocm/hooks/useClusterSupportedPlatforms.ts
```diff
@@ -41,7 +47,8 @@ export default function useClusterSupportedPlatforms(clusterId: string) {
   }, [error, addAlert, clusterId, alerts]);
 
   return {
-    isPlatformIntegrationSupported,
+    isPlatformIntegrationSupported: supportedPlatformIntegration !== undefined,
+    supportedPlatformIntegration: supportedPlatformIntegration as SupportedPlatformIntegrationType,
```
jkilzi, line 51: This should work the other way around. If `isPlatformIntegrationSupported` is true then `supportedPlatformIntegration` should hold a useful value.

Also the logic assigning the value to `supportedPlatformIntegration` is not correct now. The API could return something like: `['nutanix', 'vsphere']`. In this case the platform integration is not supported! Although, through the API might be possible to discover different hosts from different VM platforms, in the UI we allow integration only when all the hosts are originated from the same platform vendor (either Nutanix or VMware/vSphere).

jkilzi (changes requested):

src/ocm/hooks/useClusterSupportedPlatforms.ts
celdrake: Based on what we discussed, I assumed the platform integration must be unique. Meaning there cannot be mixed hosts of different platforms. So how would the UI know which of the platform is needs to use for the integration, if there is more than 1? The `cluster.platform.type` can hold a single value

src/common/selectors/clusterSelectors.ts
```diff
@@ -125,5 +125,7 @@ export const selectSchedulableMasters = (cluster: Cluster): boolean => {
   return cluster.schedulableMasters || false;
 };
 
+export const SupportedPlatformIntegrations = ['vpshere', 'nutanix'];
```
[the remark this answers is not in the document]
celdrake: I prefer that the UI is explicit about which `PlatformTypes` it considers as platform integration.
By doing this, if a new value is added on the BE, the UI would work as it did before that change. Whereas if we negate it to "anything but baremetal or none", we are not in control of when a new "platform" is enabled and we cannot say for sure if this new platform would work fine (or maybe it wasn't an actual platform integration)

celdrake:
Closing in favor of https://github.com/openshift-assisted/assisted-ui-lib/pull/1667

PR #1669, opened 2022-10-02
MGMT-10626: Add reclaiming and reclaiming-rebooting host states

mareklibra:


jkilzi (dismissed):

src/common/api/swagger.yaml
```diff
@@ -5915,6 +5921,7 @@ definitions:
       - 'belongs-to-majority-group'
       - 'valid-platform-network-settings'
       - 'ntp-synced'
+      - 'time-synced-between-host-and-service'
```
celdrake, line 5924: Adding this constant requires an update of `src/common/config/constants`.
It already has a missing key, `service-has-sufficient-spoke-kube-api-access`, so ideally we can fix the two issues in this PR.

celdrake (changes requested):

jtomasek (changes requested):
I think the new statuses need to be also added to `AGENT_FOR_SELECTION_STATUSES` list in the `src/cim/components/helpers/agents.ts`. Also they need to be added to the condition in `src/common/components/hosts/HostStatus.ts` (line 111 next to 'unbinding') to ensure a correct popover is displayed.

src/common/api/swagger.yaml
mareklibra: Time passed, the API was updated within another PR so removing the commit here.

celdrake (approved):

jtomasek (approved):

PR #1670, opened 2022-10-03
Bug 2131292: Enhance version calculation from releaseImage

jtomasek:
* Enhance the version calculation from release image url
* fall back to clusterImageSet name when the version is not retrieved from release image url

Fixes BZ2131292

Signed-off-by: Jiri Tomasek <jtomasek@redhat.com>


src/cim/components/helpers/versions.ts
```diff
@@ -2,13 +2,8 @@ import uniqBy from 'lodash/uniqBy';
...
 
-export const getVersionFromReleaseImage = (releaseImage = '') => {
-  const match = /.+:(.*)-/gm.exec(releaseImage);
-  if (match && match[1]) {
-    return match[1];
-  }
-  return '';
-};
+export const getVersionFromReleaseImage = (releaseImage = '') =>
+  (releaseImage.split(':')[1] || '').split('-')[0];
```
rawagner, line 6: I think it would be a bit better to take the last part of the `split` ,not the second one
```suggestion
  const releaseImageParts = releaseImage.split(':')
  return releaseImageParts[releaseImageParts.length-1] || '').split('-')[0]
```

jtomasek: Updated, thanks.

ammont82 (approved):

PR #1673, opened 2022-10-04
MGMT-11294 default network type on creation

celdrake:
https://issues.redhat.com/browse/MGMT-11294

The UI should not set the `network_type` when a cluster is created, it's the Backend's responsability based on many different rules.

- Stop sending `networkType` when a cluster is created
- Do not modify `networkType` for the user, unless the existing choice is not valid for the changes they are doing (eg. SDN cannot be used for DualStack clusters)
- NetworkType has been move outside the "advanced settings" section. The current behaviour of the block is to reset the settings to their default values if the section is unselected, meaning the UI would need to know the default networkType for the cluster.
This avoid duplicating the backend logic for the default networkType also in the UI.

NetworkType moved outside "advanced settings"
network-type-outside

PR for e2e tets adaptations: https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/912/

### DIFF
--- a/src/ocm/components/clusterConfiguration/networkConfiguration/StackTypeControl.tsx
+++ b/src/ocm/components/clusterConfiguration/networkConfiguration/StackTypeControl.tsx
@@ -8,14 +8,16 @@ import {
   HostSubnets,
   NetworkConfigurationValues,
   getFieldId,
-  getDefaultNetworkType,
   DUAL_STACK,
   IPV4_STACK,
+  NETWORK_TYPE_OVN,
+  NETWORK_TYPE_SDN,
   NO_SUBNET_SET,
   Cluster,
   ClusterNetwork,
   MachineNetwork,
   ServiceNetwork,
+  canSelectNetworkTypeSDN,
 } from '../../../../common';
 import { ConfirmationModal, PopoverIcon } from '../../../../common/components/ui';
 import { useDefaultConfiguration } from '../ClusterDefaultConfigurationContext';
@@ -66,7 +68,6 @@ export const StackTypeControlGroup = ({
 
   const setSingleStack = React.useCallback(() => {
     setFieldValue('stackType', IPV4_STACK);
-    setFieldValue('networkType', getDefaultNetworkType(isSNO, false));
 
     if (values.machineNetworks && values.machineNetworks?.length >= 2) {
       setFieldValue('machineNetworks', values.machineNetworks.slice(0, 1));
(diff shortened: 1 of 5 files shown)


ammont82 (approved):

src/ocm/components/clusterConfiguration/networkConfiguration/StackTypeControl.tsx
```diff
@@ -90,9 +90,12 @@ export const StackTypeControlGroup = ({
 
   const setDualStack = () => {
     setFieldValue('stackType', DUAL_STACK);
-    setFieldValue('networkType', getDefaultNetworkType(isSNO, true));
     setFieldValue('vipDhcpAllocation', false);
 
+    if (values.networkType === NETWORK_TYPE_SDN && !canSelectNetworkTypeSDN(isSNO, true)) {
```
jgyselov, line 95: Isn't the `&& !canSelectNetworkTypeSDN(isSNO, true)` part unnecessary?

jgyselov (approved):

src/ocm/components/clusterConfiguration/networkConfiguration/StackTypeControl.tsx
celdrake: With the current rules, yes it's unnecessary.

But I think this way is better because the code is being explicit as to why it's modifying the networkType, and also because if the rules ever change of when SDN is a valid option, then the only change needed is in `canSelectNetworkTypeSDN`.

Anyway, it can be removed so whatever you feel is better.

celdrake: The answer to the last question is no, because SDN was selectable for a previous status of the cluster (its stackType, and whether or not it's an SNO). And now we're checking if the clusters' new status (eg a change in its stackType) lets us keep the existing selection.

This is what the second part of the IFs tries to be clear about, that we're only updating networkType because now its value becomes invalid.

As I mentioned it can be removed, so it may be the best option to do so.

jgyselov: How is it relevant that SDN was selectable in the previous state if we the networking type value is currently SDN and we are changing it to OVN anyway?

jgyselov: Anyway, I was just trying to understand if there was a good reason to keep it there. But I do think it's best if this part of the condition is removed.

celdrake: I understand your point, and I'll remove it

jgyselov (approved):

ammont82 (approved):

PR #1674, opened 2022-10-05
Adds InlineInfoAlert and InlineLink components

jkilzi:


jgyselov (commented):
Do we really need additional components that just specify two prop values?

celdrake:
> Do we really need additional components that just specify two prop values?

I'm kindof the same opinion. Also, looking at the codebase, we have a mixture of inline and not inline AlertInfo, so we'd not even be able to refactor all of them.

PR #1675, opened 2022-10-05
MGMT-8762 - Non informative message in an inline alert regarding illegal hostname

jgyselov:
https://issues.redhat.com/browse/MGMT-8762

The following message will appear when some of the hosts have `hostname-valid` or `hostname-unique` validations failing.

image

In case there are any cluster validations failing (e.g. sufficient masters count), the new message will appear alongside the old one:
image


ammont82 (approved):

src/common/components/clusterWizard/ClusterWizardStepValidationsAlert.tsx
```diff
@@ -72,20 +89,29 @@ const ClusterWizardStepValidationsAlert = <ClusterWizardStepsType extends string
...
+                      {(failedClusterValidations.filter(Boolean) as Validation[]).map(
+                        (validation) => (
+                          <ListItem key={validation.id}>{validation.message}</ListItem>
+                        ),
+                      )}
+                    </List>
+                  </FlexItem>
+                </>
+              )}
+              {!!failedHostValidations.length && (
```
celdrake, line 107: This doesn't scale well if we need to adapt a different message.
Let's try to make it generic, still iterating over `failedHostValidations`, and printing all its messages.

Then we adapt specific validationIds to specific messages. It can still be done in this component.

celdrake (changes requested):

src/common/components/clusterWizard/ClusterWizardStepValidationsAlert.tsx
```diff
@@ -36,22 +38,37 @@ const ClusterWizardStepValidationsAlert = <ClusterWizardStepsType extends string
...
 }: ClusterWizardStepValidationsAlertProps<ClusterWizardStepsType>) => {
-  const { failedClusterValidations } = React.useMemo(() => {
+  const { failedClusterValidations, failedHostValidations } = React.useMemo(() => {
     const reducedValidationsInfo = getWizardStepClusterValidationsInfo(
       validationsInfo || {},
       currentStepId,
       wizardStepsValidationsMap,
     );
+
+    const reducedHostValidationsInfo = hosts.map((host) => {
```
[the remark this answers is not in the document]
jgyselov: I decided to use `checkHostValidation` as it is (with the boolean return type) because I'm only interested in the validation statuses (and whether they all pass or not) and not the messages.

src/common/components/clusterWizard/ClusterWizardStepValidationsAlert.tsx
```diff
@@ -36,22 +38,31 @@ const ClusterWizardStepValidationsAlert = <ClusterWizardStepsType extends string
...
+
     const flattenedValues = lodashValues(reducedValidationsInfo).flat();
+
+    const hostnameValidations = hosts.some((host) => {
+      const validations =
+        typeof host.validationsInfo === 'string'
+          ? stringToJSON<HostValidationsInfo>(host.validationsInfo)
+          : host.validationsInfo;
+
+      return checkHostValidations(validations || {}, ['hostname-valid', 'hostname-unique']);
```
celdrake, line 56: `hostnameValidations` is true if at least one host is valid, false if all the hosts are valid.
Shouldn't we be looking at the opposite, to have the message appear even if only 1 host is invalid?

Then it should be:
```suggestion
    const hasFailedHostnameValidations = hosts.some((host) => {
       ...
      return !checkHostValidations(validations || {}, ['hostname-valid', 'hostname-unique']);
    });
```

celdrake (approved):

ammont82 (approved):

PR #1678, opened 2022-10-05
MGMT-12121 Fix invalid comparison to enable ARM for OpenShift versions

celdrake:
In https://github.com/openshift-assisted/assisted-ui-lib/pull/1656, a change was introduced based on my comments in the code review. 

Turns out that the assumption was wrong, and we should always keep the `key` for the OpenshiftVersion, which can be the short form `4.11` (which is an alias to the latest patch in the same release) or the specific version form `4.11.whatever`.

Due to this change, the feature that enabled ARM architecture for certain OpenShift versions stopped working.

With this change, both ARM and setting the desired OpenShift version on cluster creation work.


ammont82 (approved):
I've tested it and it seems to work fine.

jgyselov (approved):

PR #1679, opened 2022-10-06
Mass delete hosts - delete hosts shows "pending input" status for host - clicking it opens info popup window behind the delete window

jgyselov:
https://issues.redhat.com/browse/MGMT-12081

Changes:

- By default, the popover z-index is 300. When opening from the Mass delete modal, the z-index is 500.
- The delete modal now uses `hardwareStatusColumn` instead of `statusColumn` to that the statuses are identical with the host discovery table.

image

### DIFF
--- a/src/ocm/components/hosts/HardwareStatus.tsx
+++ b/src/ocm/components/hosts/HardwareStatus.tsx
@@ -1,12 +1,16 @@
+import { sortable } from '@patternfly/react-table';
 import React from 'react';
 import {
   areOnlySoftValidationsOfWizardStepFailing,
   getWizardStepHostStatus,
   getWizardStepHostValidationsInfo,
   Host,
+  HostsTableActions,
   hostStatus,
   HostStatus,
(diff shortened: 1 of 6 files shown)


src/ocm/components/hosts/HardwareStatus.tsx
```diff
@@ -15,9 +19,10 @@ type HardwareStatusProps = {
   host: Host;
   validationsInfo: ValidationsInfo;
   onEditHostname?: () => void;
+  zIndex?: number;
```
celdrake, line 22: I'd prefer to have a prop such as `isWithinModal` and then convert it to whatever the zIndex needs to be in the displaying component.

celdrake:
Couldn't push to @jgyselov 's remote so I created a follow-up PR with a small change
https://github.com/openshift-assisted/assisted-ui-lib/pull/1704

Tested that the popup appears correctly on top of the other window.

PR #1694, opened 2022-10-18
MGMT-12275 Fix logic to determine platform integration

celdrake:
https://issues.redhat.com/browse/MGMT-12275

The logic to determine when the platform integrations are possible was wrong because it was expecting a different API response.

For a cluster with Vsphere hosts, we are receiving 
```
"supported-platforms": ['vsphere', 'none', 'metal']
```
So we should filter out `none` and `metal` and apply the same algorithm as before with those reduced values.


ammont82 (approved):

jkilzi (approved):

jkilzi:
don't forget to port it to 2.11 :)

PR #1698, opened 2022-10-18
MGMT-12094: Remove unnecessary calls to the get hosts api

jkilzi:
https://issues.redhat.com/browse/MGMT-12094

Replaces #1652

### DIFF
--- a/src/ocm/components/AddHosts/AddHosts.tsx
+++ b/src/ocm/components/AddHosts/AddHosts.tsx
@@ -38,26 +38,19 @@ const AddHosts: React.FC = () => {
   const [isSubmitting, setSubmitting] = React.useState(false);
   const clusterVarieties = useClusterStatusVarieties(cluster);
 
-  if (!cluster || !resetCluster) {
-    return null;
-  }
-
-  const handleHostsInstall = async () => {
+  const handleHostsInstall = React.useCallback(async () => {
     setSubmitting(true);
     try {
-      if (!cluster.hosts) {
+      if (!cluster || !resetCluster || !cluster.hosts) {
         return;
       }
 
-      const canAllHostsBeInstalled = cluster.hosts.every((host) =>
+      const hostsToBeInstalled = cluster.hosts.filter((host) =>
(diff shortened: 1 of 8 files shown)


ammont82 (changes requested):
For Day2, if you want to add more hosts, is not working in this way.
Maybe something like :
const hostsToBeInstalled = cluster.hosts.filter((host) =>
        canInstallHost(cluster, host.status),
      );
      if (hostsToBeInstalled && hostsToBeInstalled.length > 0) {
        await HostsService.installAll(hostsToBeInstalled);
        void resetCluster();
      } else {
        throw new Error(`Not all hosts are ready to be installed`);
      }

src/ocm/components/AddHosts/AddHosts.tsx
```diff
@@ -44,8 +45,19 @@ const AddHosts: React.FC = () => {
   const handleHostsInstall = async () => {
     setSubmitting(true);
     try {
-      await HostsService.installAll(cluster);
-      void resetCluster();
+      if (!cluster.hosts) {
```
ammont82, line 48: For Day2, if you want to add more hosts, is not working in this way. (AddHosts class)
Maybe something like :
const hostsToBeInstalled = cluster.hosts.filter((host) =>
        canInstallHost(cluster, host.status),
      );
      if (hostsToBeInstalled && hostsToBeInstalled.length > 0) {
        await HostsService.installAll(hostsToBeInstalled);
        void resetCluster();
      } else {
        throw new Error(`Not all hosts are ready to be installed`);
      }

ammont82 (changes requested):

src/ocm/components/AddHosts/AddHosts.tsx
jkilzi: Yes I agree, this part fall in between the gaps while cherry picking the commits.
I'm fixing it now. Just bear in mind the error message you put there is not quite accurate :)

celdrake (approved):

PR #1700, opened 2022-10-19
MGMT-11877 Multi-architecture hosts clusters

celdrake:
https://issues.redhat.com/browse/MGMT-11877

Adds support for clusters using hosts of different CPU architecture.
Currently supported Architectures are: `x86`, `ARM`.

- Day1 stays the same, the user select the base CPU architecture and the generated ISO for host discovery will be associated to that CPU architecture.

- On Day2, if the OpenShift version supports multi-arch, then they can select the CPU architecture for which they are generating the ISO and they can discover hosts from the supported architectures.

Needs https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/3638


ammont82 (approved):

jkilzi (approved):

celdrake:
Moving development to https://github.com/openshift-assisted/assisted-ui-lib/pull/1705
which will include the base multi-arch functionality + Static IP configuration

PR #1705, opened 2022-10-26
MGMT-11877 Day2 support for Multi-arch and Static IP configuration

celdrake:
Unified PR for https://issues.redhat.com/browse/MGMT-11877 and https://issues.redhat.com/browse/MGMT-9617

Includes the following changes for Day2 flow:
- Adds support for clusters using hosts of different CPU architecture. Currently supported Architectures are: `x86`, `ARM`.

1. Day1 stays the same, the user select the base CPU architecture and the generated ISO for host discovery will be associated to that CPU architecture.

2. On Day2, if the OpenShift version supports multi-arch, then they can select the CPU architecture for which they are generating the ISO and they can discover hosts from the supported architectures.

- Adds support to define the Static IP configuration to the Day2 hosts.

Needs https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/3638


celdrake:
Moving to the main repository to allow for other contributors
https://github.com/openshift-assisted/assisted-ui-lib/pull/1706

PR #1724, opened 2022-11-07
MGMT-12335 - Next and back buttons should be aligned with cluster status in readonly

jgyselov:
https://issues.redhat.com/browse/MGMT-12335

**Before:**
image

**After:**
image


celdrake:
@jgyselov So we want the "read-only" user to have the same kind of restrictions that the people with write privileges do?
That means that for example, if a host is disconnected, and the "read-only" user would like to look at the network configuration, we would be blocked since they can't fix the validations issues, and they wouldn't be allowed to proceed to next step.

Anyway, if that's the behaviour that we want, then I think other changes are needed. For example, in `handleOnNext` in  `ClusterDetailsForm`, (and I'm not sure if in other places too).

src/ocm/components/clusterWizard/ClusterWizardFooter.tsx
```diff
@@ -116,7 +113,7 @@ const ClusterWizardFooter = ({
       leftExtraActions={additionalActions}
       cluster={cluster}
       onFetchEvents={onFetchEvents}
-      isNextDisabled={isNextDisabled && !isViewerMode}
+      isNextDisabled={isNextDisabled}
```
celdrake, line 116: The prop can be removed as it would be part of `rest`

celdrake (changes requested):

jgyselov:
> So we want the "read-only" user to have the same kind of restrictions that the people with write privileges do? That means that for example, if a host is disconnected, and the "read-only" user would like to look at the network configuration, they would be blocked since they can't fix the validations issues, and they wouldn't be allowed to proceed to next step.

Yes, that's what the ticket is about. The question, I guess, is what use to the users is the information in the steps they would be blocked from accessing? And if a user should be able to see that information in read-only mode and not otherwise?

> Anyway, if that's the behaviour that we want, then I think other changes are needed. For example, in `handleOnNext` in `ClusterDetailsForm`, (and I'm not sure if in other places too).

```
const handleOnNext = (
    dirty: boolean,
    submitForm: FormikHelpers<unknown>['submitForm'],
    cluster?: Cluster,
  ): (() => Promise<void> | void) => {
    if (isViewerMode || (!dirty && !isUndefined(cluster) && canNextClusterDetails({ cluster }))) {
      return moveNext;
    }
    return submitForm;
  };
```

Why would that need to change? We still don't want to submit the form when you press Next in read only? Or am I missing something?

@celdrake

celdrake:
> 

 Why would that need to change? We still don't want to submit the form when you press Next in read only? Or am I missing something?
> 
> @celdrake

@jgyselov You are right, I got confused. Looks like the control is in the `ClusterWizardFooter` for all cases, so the change you did is enough.

celdrake (approved):

ammont82 (approved):

PR #1726, opened 2022-11-07
MGMT-11369 - Hosts table overflows from its container's space

jgyselov:
https://issues.redhat.com/browse/MGMT-11369

A scrollbar will appear under the table if the parent container is too small. Applies to all tables.

**Before:**

https://user-images.githubusercontent.com/87187179/200327018-c3b2a01a-ef70-4e4e-b553-1cb190def5c7.mp4

After:

https://user-images.githubusercontent.com/87187179/200327047-f28c10c9-717a-46c1-ba8b-4bf6d4e3ca2b.mp4

### DIFF
--- a/src/common/components/hosts/AITable.tsx
+++ b/src/common/components/hosts/AITable.tsx
@@ -15,6 +15,7 @@ import {
   IAction,
   ISeparator,
   TableProps,
+  InnerScrollContainer,
 } from '@patternfly/react-table';
 import { ExtraParamsType } from '@patternfly/react-table/dist/js/components/Table/base';
 import xor from 'lodash/xor';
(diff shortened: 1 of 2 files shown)


ammont82 (approved):

src/common/components/hosts/AITable.tsx
```diff
@@ -49,23 +50,25 @@ const TableMemo: React.FC<WithTestID & TableMemoProps> = React.memo(
...
-        onSort={onSort}
-        rowWrapper={rowWrapper}
-        data-testid={testId}
-        actionResolver={actionResolver ? tableActionResolver : undefined}
-        {...newProps}
-      >
-        <TableHeader />
-        <TableBody rowKey={rowKey} />
-      </Table>
+      <InnerScrollContainer>
```
celdrake, line 53: This component becomes "invisible" when switching to less wide screen sizes.
Is there something that could be done to avoid this? Maybe setting some CSS to have minimum height or something?

I know we don't commit to non-desktop viewports, but let's see if we can adjust it a bit better.

celdrake (changes requested):

src/common/components/hosts/AITable.tsx
jgyselov: Fixed

https://user-images.githubusercontent.com/87187179/200569560-e1ac48c0-51ab-4d5c-b4e7-5436360d1ecd.mp4

ammont82 (approved):

celdrake (approved):

PR #1742, opened 2022-11-10
MGMT-12075 block wizard if cluster polling fails

celdrake:
https://issues.redhat.com/browse/MGMT-12075

When the cluster polling fails due to Network / API down issues, we may be showing outdated cluster information to the user.
To ensure the cluster data integrity, we want to stop the user from interacting with the UI until the connectivity has been restored.

We need to show a new modal to the user when the cluster data may be out of date, due to network error. But we don't want to show it in other cases, such as when we abort a polling request during some update operations.

Therefore, we need to handle error types differently:
1. Error 40x --> custom error state (I don't think this worked before as we were not pass the `code` field to the `errorDetail` object)
2. Aborted request -->These errors are ignored, as the cluster may be updated from the write operation, or if not, it will be in the next polling.
3. Network errors, etc --> the `uiState` is set to  `ERROR`. We don't interrupt the polling if we already had cluster data, which means that the moment the connectivity is restored, the modal will close automatically.

### DIFF
--- a/src/ocm/api/utils.ts
+++ b/src/ocm/api/utils.ts
@@ -1,9 +1,13 @@
+import Axios, { AxiosError } from 'axios';
 import { Severity } from '@sentry/browser';
-import Axios from 'axios';
 import pick from 'lodash/pick';
 import { captureException } from '../sentry';
 import { isApiError } from './types';
 import { getErrorMessage } from '../../common/utils';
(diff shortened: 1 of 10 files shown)


ammont82:
New PR opened: https://github.com/openshift-assisted/assisted-ui-lib/pull/1763

PR #1759, opened 2022-11-23
MGMT-12578: Role selection is cut because of an inner scroll bar

jgyselov:
https://issues.redhat.com/browse/MGMT-12578

Before:
image

After:
image
image

But the bug still prevails for the action kebab menu:
image

- [ ] Open an issue with PatternFly about the action menu


celdrake (approved):

celdrake:
@jgyselov Should open a bug in our side to track the issue about the kebab menu?

jgyselov:
> @jgyselov Should open a bug in our side to track the issue about the kebab menu?

That's a good idea, I'll do that. :slightly_smiling_face:

PR #1760, opened 2022-11-23
MGMT-12382 - Nutanix cluster - Trying to enable DHCP for vips returns "The DHCP server failed to allocate the IP"

jgyselov:
https://issues.redhat.com/browse/MGMT-12382


src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -58,6 +58,10 @@ const NetworkConfigurationForm: React.FC<{
   const isAutoSaveRunning = useFormikAutoSave();
   const errorFields = getFormikErrorFields(errors, touched);
 
+  // DHCP allocation is currently not supported with Nutanix
+  // https://issues.redhat.com/browse/MGMT-12382
+  const isVipDhcpAllocationDisabled = cluster.platform?.type === 'nutanix';
```
celdrake, line 63: Should we maybe not disable it in case we have a Nutanix cluster which has DCHP ON for whatever reason (eg by using the API previously)?

That would let the user switch it OFF and then set the IPs.

jgyselov: I think it would be better if set DHCP allocation off in that case. Because with Nutanix, the auto allocation will fail.

If the IPs are already allocated, they should stay (in the input fields). But I think the checkbox should be hidden anyway.

celdrake: Makes sense :+1:

celdrake (approved):

src/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfigurationForm.tsx
```diff
@@ -54,10 +54,21 @@ const NetworkConfigurationForm: React.FC<{
...
-  const { errors, touched, isSubmitting, isValid } = useFormikContext<NetworkConfigurationValues>();
+  const { errors, touched, isSubmitting, isValid, setFieldValue, values } =
+    useFormikContext<NetworkConfigurationValues>();
   const isAutoSaveRunning = useFormikAutoSave();
   const errorFields = getFormikErrorFields(errors, touched);
 
+  // DHCP allocation is currently not supported with Nutanix
+  // https://issues.redhat.com/browse/MGMT-12382
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const isVipDhcpAllocationDisabled = React.useMemo(() => cluster.platform?.type === 'nutanix', []);
```
jkilzi, line 65:
```suggestion
  const isClusterPlatformTypeNutanix = React.useMemo(() => cluster.platform?.type === 'nutanix', []);
```

jkilzi (dismissed):

celdrake (approved):

ammont82 (approved):

PR #1768, opened 2022-11-29
Update types to enable correct typechecking on CIM

jtomasek:
Required by: https://github.com/stolostron/console/pull/2264


rawagner (approved):

src/cim/components/Hypershift/types.ts
```diff
@@ -1,12 +1,32 @@
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
 
+export type Condition = {
```
mareklibra, line 3: There is already a type for Condition in src/cim/types/k8s/shared.ts

jtomasek: Fixed

rawagner (approved):

PR #1773, opened 2022-12-01
MGMT-12323 - In UI - installation starts and after ~1 min UI "jumps" back to operators page

jgyselov:
https://issues.redhat.com/browse/MGMT-12323

This bug initially appeared as fixed by #1701 but after further investigation I learned that it still reproduces if the user stays on the original tab where the cluster was created (which QE/their tests were doing and I wasn't) because we still had `"new"` inside the location state.

https://github.com/openshift-assisted/assisted-ui-lib/blob/2fb10afc9b23a58627476eb00bb33f78c221b87a/src/ocm/components/clusterWizard/wizardTransition.ts#L40-L42

So in this PR, I'm removing the `"new"` location state when the users click Next on the "Operators" step. With that, the initial step they land on after reloading the page will not be "Operators" but one of "Host discovery", "Networking", etc.

If we want a different condition for "is the first step "Operators" or not", I'm happy to hold a discussion.

### DIFF
--- a/src/ocm/components/clusterWizard/Operators.tsx
+++ b/src/ocm/components/clusterWizard/Operators.tsx
@@ -23,6 +23,7 @@ import { updateCluster } from '../../reducers/clusters';
 import { handleApiError } from '../../api';
 import { canNextOperators } from './wizardTransition';
 import { getErrorMessage } from '../../../common/utils';
+import { useHistory, useLocation } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


celdrake (approved):
Looks good to me. We wanted to ensure that users would see the Operators page at least once. If they click next in that page, we know for sure they did.

ammont82 (approved):

PR #1781, opened 2022-12-07
MGMT-9677 - Hosts are not sorted in discovery hosts

jgyselov:
https://issues.redhat.com/browse/MGMT-9677

Hosts are generally sorted by `requested_hostname`. But let's pre-sort them by discovery time just in case.

### DIFF
--- a/src/common/components/hosts/HostsTable.tsx
+++ b/src/common/components/hosts/HostsTable.tsx
@@ -19,10 +19,10 @@ type HostsTableEmptyStateProps = {
   isSNO?: boolean;
 };
 
-export const HostsTableEmptyState: React.FC<HostsTableEmptyStateProps> = ({
+export const HostsTableEmptyState = ({
   setDiscoveryHintModalOpen,
   isSNO = false,
-}) => {
+}: HostsTableEmptyStateProps) => {
   const { t } = useTranslation();
   return (
     <EmptyState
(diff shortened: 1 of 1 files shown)


jkilzi (approved):

ammont82 (changes requested):
In other parts of the code (ex: /src/common/components/hosts/tableUtils), we use getDateTimeCell to sort date values:

export const discoveredAtColumn: TableRow<Host> = {
  header: {
    title: 'Discovered on',
    props: {
      id: 'col-header-discoveredat',
    },
    transforms: [sortable],
  },
  cell: (host) => {
    const { createdAt } = host;
    const dateTimeCell = getDateTimeCell(createdAt);
    return {
      title: dateTimeCell.title,
      props: { 'data-testid': 'host-discovered-time' },
      sortableValue: dateTimeCell.sortableValue,
    };
  },
};

jgyselov:
> In other parts of the code (ex: /src/common/components/hosts/tableUtils), we use getDateTimeCell to show/sort date values

@ammont82  That's for when the users want to sort the rows by discovery time. Whereas this PR is for the situation where we are sorting by hostname (which is the default) but the hosts have the same names. (The order in which they are sent from the BE can change at any time.) We don't want to affect the actual sorting in the table, we just don't want the rows to change their order randomly.

PR #1782, opened 2022-12-07
Adding translations for ja, zh, ko

rh-tchuang:
Translations for Japanese, Simplified Chinese and Korean is completed.


ammont82 (approved):

ammont82 (commented):
The files need to be renamed to translation.json.
All the files need to have the same name

rh-tchuang:
will be opening another PR

PR #1783, opened 2022-12-07
MGMT-12842 - Validation text from BE causing the popup box to get out of the window border

jgyselov:
https://issues.redhat.com/browse/MGMT-12842

> 1. Popovers (and modals) should always have a “max height” value which look reasonable
> 2. If the content doesn’t fit, we should show a vertical scrollbar
> 3. We should pay attention if this is an edge case or common issue
> 4. If this happens a lot, it means that we have too much content (text) and we should address this problem.

I added a wrapper around the popover content with a set maximum height. If a scrollbar appears it should be around the inner content and the close button should stay visible. ~~`25rem`~~ `33vh` seemed like a reasonable height, but I'll be happy to change it if you think otherwise.

**Before:**
image

**After:**
image


ammont82 (dismissed):

src/common/components/hosts/HostStatus.tsx
```diff
@@ -193,7 +193,11 @@ type WithHostStatusPopoverProps = AdditionNtpSourcePropsType &
 const WithHostStatusPopover: React.FC<WithHostStatusPopoverProps> = (props) => (
   <Popover
     headerContent={<div>{props.title}</div>}
-    bodyContent={<HostStatusPopoverContent {...props} />}
+    bodyContent={
+      <div style={{ maxHeight: '25rem', overflow: 'auto' }}>
```
celdrake, line 197: It's hard to say if an absolute value is going to work in all situations.
Could it be more adaptable if we used a size related to the viewport height?

jgyselov: Changed to `33vh`.

celdrake (approved):

ammont82 (approved):

PR #1799, opened 2022-12-13
MGMT-12590 Use Jira for customer bug reports

mareklibra:


src/common/config/constants.ts
```diff
@@ -14,8 +14,8 @@ export const ODF_REQUIREMENTS_LINK =
   'https://access.redhat.com/documentation/en-us/red_hat_openshift_data_foundation/4.9/html/planning_your_deployment/infrastructure-requirements_rhodf#resource-requirements_rhodf';
 export const VSPHERE_CONFIG_LINK = 'https://access.redhat.com/solutions/6677901';
 
-export const getBugzillaLink = (version = '') =>
-  `https://bugzilla.redhat.com/enter_bug.cgi?product=OpenShift%20Container%20Platform&Component=OpenShift%20Container%20Platform&component=assisted-installer&version=${version}`;
+export const getBugzillaLink = () =>
```
celdrake, line 17: Is the function used outside `assisted-ui-lib`?

If not, I suggest renaming to `getReportIssueLink` or something similar.

celdrake (approved):

src/common/config/constants.ts
celdrake: Sure, takes sense to keep the function but we now have a generic name :)

PR #1801, opened 2022-12-14
MGMT-12386 - Can create Nutanix 4.10 cluster with "Integrate with platform" enabled and checked

jgyselov:
https://issues.redhat.com/browse/MGMT-12386

Follow up on  #1708

BE doesn't seem to take feature support levels into account for `/cluster/${id}/supported-platforms`. I don't know if it should, but in any case, let's check the feature support levels before we enable the Platform integration button in the UI.

### DIFF
--- a/src/ocm/components/clusterConfiguration/platformIntegration/PlatformIntegration.tsx
+++ b/src/ocm/components/clusterConfiguration/platformIntegration/PlatformIntegration.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { OcmSwitchField } from '../../ui/OcmFormFields';
-import { Cluster, PopoverIcon } from '../../../../common';
+import { Cluster, PopoverIcon, useFeatureSupportLevel } from '../../../../common';
 import useClusterSupportedPlatforms, {
   SupportedPlatformIntegrationType,
 } from '../../../hooks/useClusterSupportedPlatforms';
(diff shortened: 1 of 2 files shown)


src/ocm/components/clusterConfiguration/platformIntegration/PlatformIntegration.tsx
```diff
@@ -43,17 +43,32 @@ const PlatformIntegrationLabel = ({
...
+  openshiftVersion,
+}: {
+  clusterId: Cluster['id'];
+  openshiftVersion: Cluster['openshiftVersion'];
+}) => {
   const { isPlatformIntegrationSupported, supportedPlatformIntegration } =
     useClusterSupportedPlatforms(clusterId);
 
+  const featureSupportLevels = useFeatureSupportLevel();
+  const isNutanixSupported = featureSupportLevels.isFeatureSupported(
```
celdrake, line 57:
```suggestion
  const isNutanixFeatureSupported = featureSupportLevels.isFeatureSupported(
```

I suggest the change to have more difference between `isPlatformIntegrationSupported` and `isNutanixSupported`, otherwise they can be confused.

jgyselov: Done :+1:

celdrake (approved):

ammont82 (approved):

PR #1802, opened 2022-12-14
MGMT-12917: Fix listing of versions from clusterimageset releaseImage

jtomasek:
- update getVersionFromReleaseImage to use the same regex as other ACM wizards, return undefined if there is no match
- getOcpVersions uses clusterImageSet name if there is no match ^

Fixes MGMT-12917


src/cim/components/helpers/versions.ts
```diff
@@ -8,8 +8,10 @@ import {
 import { OpenshiftVersionOptionType, OpenshiftVersion } from '../../../common';
 
 export const getVersionFromReleaseImage = (releaseImage = '') => {
-  const releaseImageParts = releaseImage.split(':');
-  return (releaseImageParts[releaseImageParts.length - 1] || '').split('-')[0];
+  const match = /.+:(.*)-/gm.exec(releaseImage);
```
mareklibra, line 11: We are dealing with image SHA-based tags  (i.e. `registry.ci.openshift.org/ocp/release@sha256:5f9ac79c9c257c28b8b51e79a0382de285d6e4c4e4537710b7c117601d320293`). 

I think we are doing workaround for dev/testing only, they should not be present in the prod setup.

mareklibra (approved):

src/cim/components/helpers/versions.ts
jtomasek: Right, but in general this fix is an improvement which correctly sets the clusterimageset name as a version incase when releaseImage has a non standard name which does not allow us to get a version from it.

PR #1786, opened 2022-12-07
Bug: 2120954 Don't allow using dots in cluster name

jtomasek:
BZ2120954 https://bugzilla.redhat.com/show_bug.cgi?id=2120954

Signed-off-by: Jiri Tomasek <jtomasek@redhat.com>

PR #1814, opened 2022-12-16
MGMT-12756 cluster name cannot use dot

celdrake:
https://issues.redhat.com/browse/MGMT-12756

Prevent cluster name to contain "dot" character
Similar to https://github.com/openshift-assisted/assisted-ui-lib/pull/1786/files

plus some Eslint fixes

### DIFF
--- a/src/common/components/ui/formik/validationSchemas.ts
+++ b/src/common/components/ui/formik/validationSchemas.ts
@@ -98,7 +98,7 @@ export const nameValidationSchema = (
 export const sshPublicKeyValidationSchema = Yup.string().test(
   'ssh-public-key',
   'SSH public key must consist of "[TYPE] key [[EMAIL]]", supported types are: ssh-rsa, ssh-ed25519, ecdsa-[VARIANT]. A single key can be provided only.',
-  (value) => {
+  (value: string) => {
     if (!value) {
       return true;
     }
@@ -117,11 +117,13 @@ export const sshPublicKeyValidationSchema = Yup.string().test(
 export const pullSecretValidationSchema = Yup.string().test(
   'is-well-formed-json',
   "Invalid pull secret format. You must use your Red Hat account's pull secret.",
-  (value) => {
+  (value: string) => {
     const isValid = true;
     if (!value) return isValid;
     try {
-      const pullSecret = JSON.parse(value);
+      const pullSecret = JSON.parse(value) as {
+        auths: string;
+      };
       return (
         pullSecret.constructor.name === 'Object' &&
         !!pullSecret?.auths &&
@@ -247,7 +249,7 @@ export const ipBlockValidationSchema = (reservedCidrs: string | string[] | undef
     .test(
       'cidr-is-not-unspecified',
       'The specified CIDR is invalid because its resulting routing prefix matches the unspecified address.',
-      (cidr = '') => {
+      (cidr: string) => {
         const ip = getSubnet(cidr);
         if (ip === null) {
           return false;
@@ -262,7 +264,7 @@ export const ipBlockValidationSchema = (reservedCidrs: string | string[] | undef
     .test(
       'valid-cidr-base-address',
       ({ value }) => `${value as string} is not a valid CIDR`,
-      (cidr = '') => {
+      (cidr: string) => {
         const ip = getSubnet(cidr);
         if (ip === null) {
           return false;
@@ -275,7 +277,7 @@ export const ipBlockValidationSchema = (reservedCidrs: string | string[] | undef
         return result;
       },
     )
-    .test('cidrs-can-not-overlap', 'Provided CIDRs can not overlap.', (cidr = '') => {
+    .test('cidrs-can-not-overlap', 'Provided CIDRs can not overlap.', (cidr: string) => {
       if (cidr && reservedCidrs && reservedCidrs.length > 0) {
         return !overlap(cidr, reservedCidrs);
       }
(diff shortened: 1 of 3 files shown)


ammont82:
This validation affects CIM side,not? Maybe we need a review from @jtomasek

ammont82 (dismissed):

celdrake:
> This validation affects CIM side,not? Maybe we need a review from @jtomasek

It shouldn't, but you are right that it could. I have asked Jirka for review.

mareklibra:
There was similar work done recently for CIM, I think we should unify: https://github.com/openshift-assisted/assisted-ui-lib/pull/1786

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -27,7 +27,7 @@ const NAME_CHARS_REGEX = /^[a-z0-9-.]*$/;
 const CLUSTER_NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
 // NOTE: based on https://github.com/openshift/assisted-service/blob/master/internal/cluster/validations/validations.go#L32
 const CLUSTER_NAME_REGEX =
-  /^[.-a-z0-9]?(([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)*)[.-a-z0-9]?$/;
+  /^[-a-z0-9]?(([a-z0-9]([-a-z0-9]*[a-z0-9])?([a-z0-9]([-a-z0-9]*[a-z0-9])?)*)*)[-a-z0-9]?$/;
 const CLUSTER_NAME_NO_DOT_REGEX = /^[a-z0-9-]*$/;
```
mareklibra, line 31: Isn't `CLUSTER_NAME_REGEX` what should be used?

celdrake: The interactive checker in OCM has more restrictions than what CIM allows now.
I'm guessing the regular expression could be simplified further but it was hard to understand the parts.

checker

mareklibra: I don't think `-test` is either a valid name or beginning of an URL.

I think both CIM and OCM should be unified - none of them use `dots`. And `dash (-)` is allowed in both cases. Max length is equal.

In both cases we end up with a k8s resource named by that string and later on building an URL out of it.

celdrake: @mareklibra I think you are right, in fact the `nameValidationSchema` uses the validation that the first and last characters are letters for both CIM and OCM.

PTAL, I set the same Regexp for both, one which doesn't allow the `dash` to be either at the beginning or end of the name.

ammont82 (approved):

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -26,9 +26,8 @@ const NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
 const NAME_CHARS_REGEX = /^[a-z0-9-.]*$/;
 const CLUSTER_NAME_START_END_REGEX = /^[a-z0-9](.*[a-z0-9])?$/;
 // NOTE: based on https://github.com/openshift/assisted-service/blob/master/internal/cluster/validations/validations.go#L32
-const CLUSTER_NAME_REGEX =
-  /^[.-a-z0-9]?(([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)*)[.-a-z0-9]?$/;
-const CLUSTER_NAME_NO_DOT_REGEX = /^[a-z0-9-]*$/;
+const CLUSTER_NAME_NO_DOT_REGEX =
+  /^[a-z0-9]?(([a-z0-9]([-a-z0-9]*[a-z0-9])?([a-z0-9]([-a-z0-9]*[a-z0-9])?)*)*)[a-z0-9]?$/;
```
mareklibra, line 30: When the dots are removed, this expression is way too complicated.

Without dots, we can check for kind of following one (not tested!!):
`/^[a-z0-9]?[a-z0-9-]*$/`

(Must start at an alphanumeric, followed by any count of alphanumerics or dashes. Can end at dash.)

celdrake: I agree it looks too complicated, let me see if I can refine it.
It should not be allowed to end in a dash, as there is a separate check for it (via `CLUSTER_NAME_START_END_REGEX`).
name

mareklibra: ok, then. I wasn't sure. Should I play with putting the expression together?

celdrake: @mareklibra This last comment made me think, we don't need the main RegExp to cover all cases, given that we're using `CLUSTER_NAME_START_END_REGEX` to identify the start and end of the name.

Therefore the "main" RegExp only cares about defining the allowed characters. I put back the simple RegExp expression that @jtomasek had inially written, only I renamed it for clarity.

See how it works now:
https://user-images.githubusercontent.com/829045/208850959-9dcd59f7-723b-4547-b122-bd1096fcb1f6.mp4

mareklibra: You are right, this is better when we already have it split.

jtomasek (approved):

mareklibra (approved):

PR #1817, opened 2022-12-21
Hostnames over 64 characters are not allowed

jgyselov:
https://issues.redhat.com/browse/MGMT-12661


ammont82 (approved):

package.json
```diff
@@ -37,7 +37,7 @@
     "prepare": "install-peers",
     "test": "run-s test:circular lint format:check",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 ./src/index.ts",
-    "lint": "eslint . --max-warnings=203 --color",
+    "lint": "eslint . --max-warnings=199 --color",
```
celdrake, line 40: We have a winner :1st_place_medal:

src/common/components/hosts/EditHostForm.tsx
```diff
@@ -99,22 +100,11 @@ const EditHostForm = ({
...
-                  title={status.error.title}
-                  actionClose={
-                    <AlertActionCloseButton onClose={() => setStatus({ error: null })} />
-                  }
-                  isInline
-                >
-                  {status.error.message}
-                </Alert>
-              )}
+              <AlertFormikError status={status as StatusErrorType} />
```
celdrake, line 107: You need to pass the function for closing the alert.

package.json
ammont82: Bravo @jgyselov !

src/common/components/ui/formik/constants.tsx
```diff
@@ -29,11 +29,10 @@ export const acmClusterNameValidationMessages = (t: TFunction) => ({
 });
 
 export const nameValidationMessages = (t: TFunction) => ({
-  INVALID_LENGTH: t('ai:1-253 characters'),
+  INVALID_LENGTH: t('ai:1-63 characters'),
```
celdrake, line 32: We'll need to check with CIM too because the `richNameValidationSchema` is being used in a few places, looks like not only for host names.

celdrake (changes requested):

package.json
jgyselov: Good team effort! Bravo to all of us! :fireworks:

ammont82 (dismissed):

src/common/components/ui/formik/constants.tsx
jtomasek: How about keeping this function intact and introducing a specific one for a hostname `hostnameValidationMessages`? CIM uses this as a general validation for a resource name.

src/common/components/ui/formik/validationSchemas.ts
```diff
@@ -333,7 +332,7 @@ export const richNameValidationSchema = (t: TFunction, usedNames: string[], orig
   const nameValidationMessagesList = nameValidationMessages(t);
   return Yup.string()
     .min(1, nameValidationMessagesList.INVALID_LENGTH)
-    .max(253, nameValidationMessagesList.INVALID_LENGTH)
+    .max(63, nameValidationMessagesList.INVALID_LENGTH)
```
jtomasek, line 335: Similar as with `nameValidationMessages` function, can we add a dedicated function for hostname `richHostnameValidationSchema`?

jtomasek (dismissed):

src/common/components/ui/formik/constants.tsx
jgyselov: We already had `hostnameValidationMessages` use by both OCM and CIM. I've renamed it to `BMCValidationMessages` (for you to use in CIM) and created new `hostnameValidationMessages` to go with the new `richHostnameValidationSchema` for our purposes in ocm/common.

These is likely a lot of duplicate code now, but 
1. CIM side should be unaffected
2. dealing with different validations based on different conditions in Yup schemas is a pain in the...

batzionb:
@jtomasek referred me to this PR, because I'm also dealing with name validations
I noticed [this validation](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/components/ui/formik/constants.tsx#L36) regarding hostnames: 
Number of characters between dots (.) must be 1-63
In this PR seems like maximum of 63 is the whole hostname and not the parts between the dots
So this validation can be removed? Or is relevant to something else?

celdrake:
> @jtomasek referred me to this PR, because I'm also dealing with name validations I noticed [this validation](https://github.com/openshift-assisted/assisted-ui-lib/blob/master/src/common/components/ui/formik/constants.tsx#L36) regarding hostnames: Number of characters between dots (.) must be 1-63 In this PR seems like maximum of 63 is the whole hostname and not the parts between the dots So this validation can be removed? Or is relevant to something else?
> 

I don't think the validation can be removed. Julie's changes in this PR were split so they only affect hostnames (via `richHostnameValidationSchema`), at least for OCM. 

The validation that requires max characters between dots to be 63 seems to be still used via `nameValidationMessages` function which is used by ACM for InfraEnv , BMC.

batzionb:
I'm just curious about the difference between the validations...
@jtomasek is this validation (Number of characters between dots (.) must be 1-63) needed for CIM?
Maybe in CIM it should be max 63 for hostnames, and max 253 for CRs?

celdrake (approved):

ammont82 (approved):

PR #1824, opened 2022-12-22
MGMT-13028 - Show only enabled operators in the Review and create summary

jgyselov:
https://issues.redhat.com/browse/MGMT-13028

(The changed test ids shouldn't be an issue - the tests weren't using them anyway.)


src/ocm/components/clusterConfiguration/review/ReviewOperatorsTable.tsx
```diff
@@ -15,43 +15,26 @@ export const ReviewOperatorsTable = ({ cluster }: { cluster: Cluster }) => {
   const { t } = useTranslation();
   const operatorNames = operatorLabels(t);
 
+  const operators = React.useMemo(
+    () => [OPERATOR_NAME_CNV, OPERATOR_NAME_ODF, OPERATOR_NAME_LVM],
```
celdrake, line 19: Ideally I'd want to have a type defined in a similar way to `SupportedPlatformType` that we can define in a single place and drives all operators. 

Therefore if we are missing an entry for an operator in such type, we'd get a TS error and we don't need to actively think about all places where we need to change when a new operator is added.

jgyselov: I tried my best :slightly_smiling_face:

src/common/config/constants.ts
```diff
@@ -285,21 +285,14 @@ export const OPERATOR_NAME_CNV = 'cnv';
 export const OPERATOR_NAME_LSO = 'lso';
 export const OPERATOR_NAME_ODF = 'odf';
 export const OPERATOR_NAME_LVM = 'lvm';
-export const OPERATOR_NAME_CVO = 'cvo';
```
jgyselov, line 288: I removed the ones (with the corresponding translation stuff) that were unused in the codebase.

src/common/config/constants.ts
```diff
@@ -285,21 +285,14 @@ export const OPERATOR_NAME_CNV = 'cnv';
...
-export const OPERATOR_NAME_CONSOLE = 'console';
```
jgyselov, line 289: I removed the ones (with the corresponding translation stuff) that were unused in the codebase.

src/common/config/constants.ts
```diff
@@ -285,21 +285,14 @@ export const OPERATOR_NAME_CNV = 'cnv';
...
-  | typeof OPERATOR_NAME_CVO
-  | typeof OPERATOR_NAME_CONSOLE;
-
-export const operatorLabels = (t: TFunction): { [key in OperatorName]: string } => ({
-  [OPERATOR_NAME_CONSOLE]: t('ai:OpenShift Console'),
-  [OPERATOR_NAME_CVO]: t('ai:OpenShift Cluster Version Operator'),
-  [OPERATOR_NAME_LSO]: t('ai:OpenShift Local Storage'),
+
+const OperatorNames = [OPERATOR_NAME_CNV, OPERATOR_NAME_LSO, OPERATOR_NAME_ODF, OPERATOR_NAME_LVM];
+export const ExposedOperatorNames = [OPERATOR_NAME_CNV, OPERATOR_NAME_ODF, OPERATOR_NAME_LVM];
```
jgyselov, line 290: If at any time we expose LSO to the users (or stop using `OPERATOR_NAME_LSO` in the code), we can remove the distinction between operators and exposed operators.

src/common/config/constants.ts
celdrake: We should check with CIM as I know there are projects that have references to `assisted-ui-lib`

src/ocm/components/clusterConfiguration/review/ReviewOperatorsTable.tsx
```diff
@@ -16,42 +13,18 @@ export const ReviewOperatorsTable = ({ cluster }: { cluster: Cluster }) => {
...
-              ? 'Enabled'
-              : 'Disabled',
-            props: { 'data-testid': 'openshift-data-foundation' },
-          },
-        ],
-      },
-    ];
-  }, [cluster, operatorNames]);
+    return ExposedOperatorNames.filter((operator) =>
+      hasEnabledOperators(cluster.monitoredOperators, operator),
```
celdrake, line 17: You have fixed some Eslint warnings by properly typing the data, please update `package.json`

src/common/config/constants.ts
```diff
@@ -285,21 +285,14 @@ export const OPERATOR_NAME_CNV = 'cnv';
...
+
+export type OperatorName = typeof OperatorNames[number];
+export type ExposedOperatorNameType = typeof ExposedOperatorNames[number];
```
celdrake, line 293: either both are `(Exposed)OperatorName` or `(Exposed)OperatorNameType`

celdrake (changes requested):
Looks good! 
We won't need to remember to update the `ReviewOperatorsTable` when a new operator is added, and we have a more robust code and less warnings :)

celdrake (approved):
All good, only remaining question is whether it's safe to delete the unused operators

ammont82 (approved):

celdrake (approved):

PR #1825, opened 2022-12-22
Update axios version according to OCM

mareklibra:


ammont82 (approved):

package.json
```diff
@@ -53,7 +53,7 @@
     "@patternfly/react-tokens": "^4.58.2",
     "@reduxjs/toolkit": "^1.5",
     "@sentry/browser": "^5.9 || ^6.0",
-    "axios": ">=0.19.2 <1.0.0",
+    "axios": ">=0.22.0 <1.0.0",
```
jkilzi, line 56: ## tl;dr
```suggestion
    "axios": ">=0.23.0 <1.0.0",
```

## Why?
We'll need also to bump `axios-case-converter`. Currently we depend on v0.8.1 which [depends on axios >=0.21.1 at least in this version](https://github.com/mpyw/axios-case-converter/blob/ebcbc417a512347928d27ca88354932a73531dca/package.json#L74)   
In version 0.9.0 they require [at least axios >= 0.23.0](https://github.com/mpyw/axios-case-converter/blob/f72cfadf04e19c90fcc3163456b0abfa6fdb51c0/package.json#L74)  
It will be best if we could at least anchor the lower-bound on our side as well to be at least `axios: >= 0.23.0 <1.0.0` and bump `axios-case-converter` to `^0.9.0`.

jkilzi (changes requested):

package.json
mareklibra: good point, thanks, updating

mareklibra: With `axios-case-converter`, we can go to the latest. Based on release notes, there shouldn't be breaking changes affecting us: https://github.com/mpyw/axios-case-converter/releases

And we will be till in axios range `=>0.23.0`

mareklibra: Recently the OCM has `axios@0.27.2`, so we are safe here

jkilzi (approved):

PR #1826, opened 2022-12-22
MGMT-13060 Display "Platform Integration" in new "Review & Create" page

celdrake:
https://issues.redhat.com/browse/MGMT-13060

- Adding the note about Platform integration to the new Review & Create page, in the "Host inventory" block
platform

Note: The section "Host inventory" was previously "Inventory", I renamed it based on the latest mockup.

### DIFF
--- a/src/ocm/components/clusterConfiguration/review/ReviewSummary.tsx
+++ b/src/ocm/components/clusterConfiguration/review/ReviewSummary.tsx
@@ -2,11 +2,16 @@ import React from 'react';
 import { ExpandableSection } from '@patternfly/react-core';
 import {
   Cluster,
+  DetailItem,
+  DetailList,
   hasEnabledOperators,
+  isClusterPlatformTypeVM,
   OPERATOR_NAME_CNV,
   OPERATOR_NAME_LVM,
   OPERATOR_NAME_ODF,
   ReviewHostsInventory,
+  SupportedPlatformType,
+  RenderIf,
 } from '../../../../common';
 import {
   ReviewClusterDetailTable,
@@ -14,6 +19,7 @@ import {
   ReviewOperatorsTable,
   TableSummaryExpandable,
 } from '.';
+import PlatformIntegrationNote from '../platformIntegration/PlatformIntegrationNote';
 
 const ReviewSummary = ({ cluster }: { cluster: Cluster }) => {
   const showOperatorsSummary =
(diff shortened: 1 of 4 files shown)


ammont82 (dismissed):

ammont82 (commented):
ERROR in this line:

/home/runner/work/assisted-ui-lib/assisted-ui-lib/src/ocm/components/clusterConfiguration/review/ReviewPreflightChecks.tsx
  54:10  error  'GridItemProps' is defined but never used  @typescript-eslint/no-unused-vars

celdrake:
> ERROR in this line:
> 
> /home/runner/work/assisted-ui-lib/assisted-ui-lib/src/ocm/components/clusterConfiguration/review/ReviewPreflightChecks.tsx 54:10 error 'GridItemProps' is defined but never used @typescript-eslint/no-unused-vars

Thanks, should be fixed now :sweat_smile:

ammont82 (dismissed):

jgyselov (commented):
image
image

I'm not a big fan of having the "Platform integration" section in the second row of the collapsed component. I think it would look better all in one row, something like this, perhaps:

image

Here I modified the grid to span 10 columns instead of the default 12 and every grid item to span two columns. (It didn't seem to work using PatternFly props, so I had to override the CSS - but maybe I was doing something wrong...)

image

Then we could get rid of the offsets and margins...

@celdrake @parvathyvr WDYT?

celdrake:
> image image
> 
> I'm not a big fan of having the "Platform integration" section in the second row of the collapsed component. I think it would look better all in one row, something like this, perhaps:
> 
> image
> 
> Here I modified the grid to span 10 columns instead of the default 12 and every grid item to span two columns. (It didn't seem to work using PatternFly props, so I had to override the CSS - but maybe I was doing something wrong...)
> 
> image
> 
> Then we could get rid of the offsets and margins...
> 
> @celdrake @parvathyvr WDYT?

I think with two rows it behaves better for not so large screens. 

https://user-images.githubusercontent.com/829045/209168633-5637c002-e95f-4c97-9b74-3c6560aec194.mp4

parvathyvr:
@jgyselov I agree that it looks better if we could have all 4 in 1 row.
 But not sure if we should worry too much of having them in all in a single row since 
1. More preflight checks might get added in the future. 
2. I think Platform integration is not going to be always present.But the other 3 will always be there.
3. For smaller screen sizes, 3 column layout would be easier to scan and read. 
@nirfarkas any thoughts on this?

jgyselov:
On a slightly different note - does the platform integration part absolutely need to go in the Preflight checks section? Could we perhaps put it in the Cluster summary instead? I think the Cluster summary tables deal better with adding new information - we could add a new row with "Plaform integration" and the hosts' platform and maybe do something similar to when the cluster has an unsupported OpenShift version:

image

Just throwing this idea out there... @parvathyvr @nirfarkas WDYT?

ammont82 (approved):

src/ocm/components/clusterConfiguration/review/ReviewSummary.tsx
```diff
@@ -38,8 +44,24 @@ const ReviewSummary = ({ cluster }: { cluster: Cluster }) => {
         </TableSummaryExpandable>
       )}
 
-      <TableSummaryExpandable title={'Inventory'}>
-        <ReviewHostsInventory hosts={cluster.hosts} />
+      <TableSummaryExpandable title={'Host inventory'}>
+        <>
+          <ReviewHostsInventory hosts={cluster.hosts} />
+          <RenderIf condition={isClusterPlatformTypeVM(cluster)}>
```
jkilzi, line 50: Please remind me, is platform integration with Nutanix also possible?
If it is and there is no extra setup needed with them then perhaps we shouldn't display anything in that case?

celdrake: I'm checking that, I was unaware of it.

celdrake: @jkilzi Confirmed that Nutanix also has extra configuration steps, which we show in the link ( https://access.redhat.com/solutions/6983944 for Nutanix)

jkilzi (approved):

PR #1831, opened 2022-12-22
Don't run lint and format before building the project

jkilzi:
These scripts are executed anyway on every PR, there is no need to execute them again when building the project.


ammont82 (approved):

.github/workflows/on-pull-request.yaml
```diff
@@ -38,5 +38,14 @@ jobs:
           ./scripts/validatei18n_files.sh
         shell: bash
 
-      - name: Build code (includes tests)
+      - name: Check for circular dependencies
```
celdrake, line 41: Why not use the shortcut `yarn test` that does all checks ? So if we add a new check, it will likely be added there vs. having to add it to this list.

Since we are at it, I think the name `test` is not appropriate anymore, maybe rename it to something like `validate` or `run-checks` ...

jkilzi: Absolutely agree with the semantics of `test`... this is not testing anything!
However, I don't agree with having an all-in-one command, the main reason is because it makes the job log harder to read.
By separating these steps you can easily find at  what step there is a failure and fix it swiftly.

ammont82 (approved):

.github/workflows/on-pull-request.yaml
jkilzi: sure

jkilzi:
Closing this PR for now, these changes will be incorporated into the monorepo PR

PR #1850, opened 2023-01-07
TypeScript fixes

mareklibra:


src/ocm/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -60,6 +60,18 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
   const canAddHosts =
     (!isSNO(cluster) || isSNOExpansionAllowed) && cluster.status === 'installed' && !ocmClient;
 
+  const onAddHosts = () => {
```
jkilzi, line 63: `useCallback`?

src/common/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -179,7 +176,7 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
 
   const onDeleteChip: ToolbarFilterProps['deleteChip'] = (type, chip) => {
     if (type) {
-      const id = chip['key'] || chip;
+      const id = (chip['key'] || chip) as string;
```
jkilzi, line 179: do we have any guarantee either `chip.key` or `chip` will be defined or will be a string?

src/common/components/ui/formik/MultiSelectField.tsx
```diff
@@ -29,11 +29,11 @@ const MultiSelectField: React.FC<MultiSelectFieldProps> = ({
   ...props
 }) => {
   const [isOpen, setOpen] = React.useState(false);
-  const [field, { touched, error }, { setValue }] = useField(props.name);
+  const [field, { touched, error }, { setValue }] = useField<string[]>(props.name);
   const fieldId = getFieldId(props.name, 'multiinput', idPostfix);
   const isValid = !(touched && error);
   const errorMessage = !isValid ? error : '';
-  const hText = getHelperText ? getHelperText(field.value) : helperText;
+  const hText = getHelperText ? getHelperText(field.value?.join(',')) : helperText;
```
jkilzi, line 36: `field.value?.join(',')` will evaluate to `undefined` when there is no `value`. Maybe appending `?? ''`  will help?  
Also what I find interesting is w/o peeking into `getHelperText` definition, how much responsibility this function has.
Looks like it's fine accepting undefined values, and if its type is `field.value: string[]` then it should know what to do with that array already. So perhaps there is no need to call `.join(',')`???

jkilzi (changes requested):
Marek kudos for lowering the number to 125!
I'm leaving a couple of questions, PTAL

rawagner (approved):

src/common/components/ui/ClusterEventsToolbar.tsx
mareklibra: yes, we do

src/ocm/components/clusterWizard/ClusterDetailsForm.tsx
```diff
@@ -148,7 +148,7 @@ const ClusterDetailsForm = (props: ClusterDetailsFormProps) => {
                 (dirty || (cluster && canNextClusterDetails({ cluster })))
               )
             }
-            onNext={handleOnNext(dirty, submitForm, cluster)}
+            onNext={void handleOnNext(dirty, submitForm, cluster)}
```
celdrake, line 151: It may not be due to this change, but in OCM the  "Next" button is missing.
I wanted to check if it worked fine because I think in this case the return value is needed.
next-button

src/common/components/ui/formik/MultiSelectField.tsx
mareklibra: done

src/ocm/components/clusterDetail/ClusterDetail.tsx
mareklibra: done

celdrake (changes requested):

src/ocm/components/clusterWizard/ClusterDetailsForm.tsx
mareklibra: I don't think the return value is used. Plus it is triggered synchronously without waiting on result.

TypeScript:
```export type WizardFooterGenericProps = {
  onNext?: () => void;
  ```
  
  Used as-it-is for Button's `onClick`

Comment (celdrake) [reply src/ocm/components/clusterWizard/ClusterDetailsForm.tsx]: Looks right, I didn't check but there is definitely a problem with the Next button. Could you see what triggered it?

File: src/ocm/components/clusterDetail/CancelInstallationModal.tsx [hunk 5]
```diff
@@ -66,7 +66,9 @@ const CancelInstallationModal: React.FC = () => {
         <Button
           key="submit"
           variant={ButtonVariant.danger}
-          onClick={handleSubmit}
+          onClick={() => {
+            void handleSubmit();
```
celdrake, line 70: I'd revert this unless we test this case specifically

src/ocm/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -160,7 +165,7 @@ const ClusterDetail: React.FC<ClusterDetailProps> = ({ cluster }) => {
             <ToolbarButton
               id="cluster-installation-logs-button"
               variant={ButtonVariant.link}
-              onClick={() => downloadClusterInstallationLogs(addAlert, cluster.id)}
+              onClick={() => void downloadClusterInstallationLogs(addAlert, cluster.id)}
```
celdrake, line 168: Same here, we can fix this issues in the future. It's in multiple places, so it's safer to revert all of them and test each case individualy.

src/ocm/components/clusterWizard/ClusterDetailsForm.tsx
```diff
@@ -148,7 +148,7 @@ const ClusterDetailsForm = (props: ClusterDetailsFormProps) => {
                 (dirty || (cluster && canNextClusterDetails({ cluster })))
               )
             }
-            onNext={handleOnNext(dirty, submitForm, cluster)}
+            onNext={() => void handleOnNext(dirty, submitForm, cluster)}
```
celdrake, line 151: Let's revert for now and fix later

celdrake (changes requested):

src/ocm/components/clusterWizard/ClusterDetailsForm.tsx
mareklibra: You are right. Reverting this change. Let's fix separately. 

For future reference:
The issue is in `handleOnNext` returning `() => Promise<void> | void`.
The `onNext` requires its result but `handleOnNext` should be changed to return only `() => void`.

src/ocm/components/clusterDetail/ClusterDetail.tsx
mareklibra: reverting here

src/ocm/components/clusterDetail/CancelInstallationModal.tsx
mareklibra: This change is safe, imo

celdrake (approved):
Nice job! Massive drop in the warning number :)

rawagner (approved):

celdrake (approved):

src/common/components/ui/ClusterEventsToolbar.tsx
```diff
@@ -177,9 +174,9 @@ const ClusterEventsToolbar: React.FC<ClustersListToolbarProps> = ({
     });
   };
 
-  const onDeleteChip: ToolbarFilterProps['deleteChip'] = (type, chip) => {
+  const onDeleteChip: ToolbarFilterProps['deleteChip'] = (type, chip: ToolbarChip | string) => {
     if (type) {
-      const id = chip['key'] || chip;
+      const id = (chip['key'] || chip || '') as string;
```
jkilzi, line 179: BTW, you can use `??` instead of `||`. The former will use the RHS when the LHS is either `null` or `undefined`. It doesn't coerce the LHS into a boolean.

jkilzi (approved):

rawagner (approved):

PR #1855, opened 2023-01-10
MGMT-12825 Use new Github actions updated to Node16

celdrake:
https://issues.redhat.com/browse/MGMT-12825

Updated Github actions to remove deprecation warnings for Node 12.
New versions have been migrated to Node 16.

See:
https://github.com/redhat-actions/push-to-registry/releases/tag/v2.7
https://github.com/redhat-actions/buildah-build/releases/tag/v2.11

### DIFF
--- a/.github/workflows/on-pull-request.yaml
+++ b/.github/workflows/on-pull-request.yaml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           cache: yarn
-          node-version: '12'
+          node-version: 16
 
       - name: Install dependencies
         # --frozen-lockfile: doesn't generate a yarn.lock file, fails if an update is needed.
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -94,7 +94,7 @@ jobs:
           context: ./assisted-ui
 
       - name: Push the assisted-ui image to Quay.io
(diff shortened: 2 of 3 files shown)


celdrake:
I only changed the version for push-to-registry because for buildah, it was already using the latest version.
We can also change `push-to-registry` to the open v2 so it get updates. WDYT?

See https://github.com/openshift-assisted/assisted-ui-lib/actions/runs/3875094028/jobs/6607154355

jkilzi (approved):

jkilzi:
If the actions require now Node 16 then we should reflect that [here](https://github.com/openshift-assisted/assisted-ui-lib/pull/1855/files#diff-c366d7d7cb2829080753d85f60c4f8e1fdbb301d20d707e0e7872822ead6ae88R35) too

ammont82 (dismissed):

celdrake:
> If the actions require now Node 16 then we should reflect that [here](https://github.com/openshift-assisted/assisted-ui-lib/pull/1855/files#diff-c366d7d7cb2829080753d85f60c4f8e1fdbb301d20d707e0e7872822ead6ae88R35) too

I'll see to change all of the node versions to 16 instead of 12 while we're at it.

ammont82 (approved):

jkilzi (approved):

PR #1856, opened 2023-01-10
MGMT-13109: Cluster update - send only allowed fields

celdrake:
https://issues.redhat.com/browse/MGMT-13109

After a change in the API, the UI can only send a payload with accepted fields in PATCH requests on a Cluster.
I created an "allowed list" instead of a "ignored list" of fields to be in control of what's being send to the API.

I reviewed all calls to Cluster PATCH API and I couldn't see other problems. Also CI from QE team seems to confirm this.


jkilzi:
What about the others APIs? InfrEnvs for example?
Honestly, this is not the way I expected to resolve the issue. Think about how much code we'll need to maintain when a new endpoint is created.

celdrake:
> What about the other APIs? InfrEnvs for example? Honestly, this is not the way I expected to resolve the issue. Think about how much code we'll need to maintain when a new endpoint is created.

For clarification, I've created a PR with 2 commits.


1. https://github.com/openshift-assisted/assisted-ui-lib/pull/1857/commits/7403d3d745d6641ccfcd7e2be65f8378ddde2299 is the part about the bugfix. It changes the type of the object being passed to `handleClusterUpdate` from the type of a cluster creation, to the type of cluster edition. From that object, we cherry-pick only the fields that are actually editable in that screen.

2. https://github.com/openshift-assisted/assisted-ui-lib/pull/1857/commits/d46a273583c992d3b4ff8b2f90a8aede9e9db31e This a larger refactor. The purpose is that, in the ClusterDetail screen, about 90% of the fields are not editable. So it doesn't really make sense to use `V2ClusterUpdateParams` as its type. So I defined a type which represents only the editable fields in that step.

Let me know what you think.

celdrake:
@jkilzi Updated the PR to avoid using pick or omit.
The real problem in this case, is that the Type `V2ClusterUpdateParams` has props which are all present in `ClusterCreateParams`, and this second Type has more properties, which are ignored for TS.

Now we're not mixing the types between the creation and edition of a cluster.

ammont82 (approved):

jkilzi (approved):

jkilzi:
needs port to r/2.14

PR #1860, opened 2023-01-11
MGMT-13241: Use infraEnv to detect the clusters Cpu Architecture

celdrake:
https://issues.redhat.com/browse/MGMT-13241

The Cluster's `cpuArchitecture` doesn't have the expected values if it uses a multi-architecture image.
Therefore, to detect the underlying Cpu Architecture of a cluster, we must look at its associated InfraEnv.

On the other hand, to display if a given OpenshiftVersion is capable of multi-arch, we must look at the Cluster's Cpu Architecture.

I tried to make this distinction as clear as possible.
fix


jkilzi (approved):

jkilzi:
needs port to r/2.14

ammont82 (approved):

PR #1870, opened 2023-01-16
Fix TS error

celdrake:
- Correct resolution for react types. Fixes problems with React.ReactNode etc

- Included some particular TS errors. Down to about 70 errors.


jkilzi:
Approving and merging this PR will introduce the entire monorepo. That shouldn't be the scope/intention right now. At least before that, we should have a good description of the new folder structure and how to get started so everybody can remain productive. Also, even if all the fixes to the TS files are OK, we'll still need to get the GH actions adapted because they are still executing the old workflow which no longer applies to this setup.

@celdrake I'll cherry-pick those commits you made on top of my dev branch and I'll prepare a new PR that will introduce properly all the changes we made to the repo.

@jtomasek is working also on a watch script for the ui-lib. That script is also a prerequisite.

celdrake:
@jkilzi yes, I couldn't push to your fork hence couldn't do a proper PR targeting your branch.

The first commit solves the generic issues with React types. This should be applied to your branch.

The second commit SHOULDN'T be applied to your branch. I started fixing TS errors, but these need to be fixed in the current `assisted-ui-lib`. Because of the different file structure, we can't cherry-pick commits, so I'll use the PR as a reference to port the fixes to `master` branch.

jkilzi:
> @jkilzi yes, I couldn't push to your fork hence couldn't do a proper PR targeting your branch.
> 
> The first commit solves the generic issues with React types. This should be applied to your branch.
> 
> The second commit SHOULDN'T be applied to your branch. I started fixing TS errors, but these need to be fixed in the current `assisted-ui-lib`. Because of the different file structure, we can't cherry-pick commits, so I'll use the PR as a reference to port the fixes to `master` branch.
> 
> The second commit is now in this stand-alone PR targeting master branch: #1871

10x, OK then let's close this PR. I'll cherry-pick the first commit as you said.

PR #1876, opened 2023-01-18
Fix typescript 4 errors part 3

celdrake:
- More Typescript v4 fixes.

Added a typed equivalent to `Object.keys` which returns a type based on the object.
This makes the casting when accessing the values unnecessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "prepare": "install-peers",
     "test": "run-s test:circular lint format:check",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 ./src/index.ts",
-    "lint": "eslint . --max-warnings=115 --color",
+    "lint": "eslint . --max-warnings=114 --color",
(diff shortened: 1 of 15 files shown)


jkilzi:
For the record, [this thread|https://github.com/microsoft/TypeScript/issues/38520] seems to be relevant

jkilzi:
How can we enforce `getKeys` will be used instead of `Object.keys`?  
Otherwise, we'll keep fixing this issue over and over again...

celdrake:
> How can we enforce `getKeys` will be used instead of `Object.keys`? Otherwise, we'll keep fixing this issue over and over again...

With a quick search, I found this
https://github.com/remithomas/eslint-plugin-ban

Maybe it makes more sense to include it in the mono-repo instead of the master branch?

jkilzi (approved):

ammont82 (approved):

PR #1879, opened 2023-01-22
AGENT-513: Single cluster support

mareklibra:
Changes needed for the new use-case of single-cluster assisted UI flow (Billi), mostly focused on the disconnected environment.

There is new application consuming the ai-ui-lib, so far hosted here: https://github.com/mareklibra/assisted-ui-single .
The project is about to be transferred to another organization (wip recently)


src/ocm/components/clusterConfiguration/OcmClusterDetailsFormFields.tsx
```diff
@@ -68,6 +69,7 @@ export const OcmClusterDetailsFormFields = ({
     nameInputRef.current?.focus();
   }, []);
 
+  const isSingleClusterFeatureEnabled = useFeature('ASSISTED_INSTALLER_SINGLE_CLUSTER_FEATURE');
```
celdrake, line 72: The only thing I can think of to make the UI more flexible is not to use directly `useFeature`, but instead create a similar function which can tell us which UI features are enabled or not. 

For example
const { hasNetworkConfigurationControlGroup } = getUiFeatureConfig();

This new function would centralise the logic about what UI features are enabled in each scenario. 
If it only depends on data we have on the cluster, it can be feasible.

mareklibra: Thank you. Let's take this offline. I understand your point but it seems to me like wrapping a wrapper.
It's subjective if that brings value, I am afraid that this is the a parallel to having 5 opinions per 5 people asked.

src/ocm/components/clusters/ClusterPage.tsx
```diff
@@ -37,8 +37,13 @@ type MatchParams = {
   clusterId: string;
 };
 
-const ClusterPage: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
-  const { clusterId } = match.params;
+const ClusterPageGeneric: React.FC<{ clusterId: string; showBreadcrumbs: boolean }> = ({
```
jkilzi, line 40: I don't see any changes in the Router. Is this change safe?

jkilzi (changes requested):

src/ocm/components/clusters/ClusterPage.tsx
mareklibra: Yes, see line 188:
```
export const ClusterPage
```

mareklibra: and 
https://github.com/openshift-assisted/assisted-ui-lib/pull/1879/files/6b08f09391dbb60f105282501c9f61a409835430#diff-405056fa39d9616ee42a0921614834f7e59ea052d79936ca0b82781049dfc3a2R6

celdrake (approved):
Looks good to me

src/ocm/components/clusters/ClusterPage.tsx
jtomasek: `showBreadcrumbs` can be optional with a false default value

jtomasek (approved):
Looks good.
As discussed some time ago, if the single cluster mode should further diverge from the OCM flow it would be much preferable to use composition instead of conditionals.

src/ocm/components/clusterDetail/RebootNodeZeroModal.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { Button, Checkbox, Modal, ModalVariant, Text, TextContent } from '@patternfly/react-core';
+
+import { useClusterStatusVarieties } from './ClusterDetailStatusVarieties';
+import { Cluster, KubeconfigDownload, ClusterCredentials } from '../../../common';
+
+import './RebootNodeZeroModal.css';
+
+// TODO(mlibra): Following will be reimplemented based on future decisions in https://issues.redhat.com/browse/AGENT-522
```
jkilzi, line 9: Here are my thoughts on this part:  
1. This component won't be seen in OCM, TTBOMK there's no feature requiring this modal however, lives as part of the OCM-part of the lib... (but you see were this is going in the general case)
2. The day the mentioned decision behind AGENT-522 happens we'll need to release a new version of the ui-lib.

So it seems like we will need to release versions of the "ocm-lib" because either they are required for ABI and/or OCM. I think we should gather and brainstorm around this solution or rethink what is "shared/common code" before we get ourselves into a maintenance nightmare.

celdrake (changes requested):
Based on @jkilzi comments, I agree that it may be difficult to maintain. Let's discuss it further first

src/ocm/components/clusterDetail/RebootNodeZeroModal.tsx
jkilzi: We had a meeting today where we agreed on the following steps:
1. Merge this PR as a WIP step.
2. A project redesign is required to ensure parts of the lib can be handed over in the future.
3. Open a long-term epic in order to begin decomposing the lib/s into smaller chunks.
4. Use the mono-repo as the hosting infrastructure for this refactor.

jkilzi (approved):

PR #1880, opened 2023-01-23
WIP ACM-3002: Fix conflicting IngressControllers for AWS NLB

mareklibra:
To do:
- [ ] file documentation bug, suggest changes, update linked URL in this PR
- [ ] update IngressControlles page link


mareklibra:
Screenshot from 2023-01-24 10-52-56

wolfeallison:
@mareklibra It looks good! A couple of comments:

- "If there are many clusters, use higher values for the storage fields." Could we be more specific on what "many" is here? Do we think the user knows what is a high value for the storage field? 
- "Storage sizes cannot be changed once you configure" - change "once" to "after" because it can be mistranslated to "one time"
- I'd change "List of IngressController resources" to "View IngressController resources" so it starts with a verb to match the "learn more" beside it

jtomasek (approved):

PR #1887, opened 2023-01-25
MGMT-13134 - Cluster name box echo all characters when overlimit

jgyselov:
https://issues.redhat.com/browse/MGMT-13134

`maxLength` won't allow to input more characters that 54.


src/ocm/components/clusterConfiguration/OcmClusterDetailsFormFields.tsx
```diff
@@ -82,6 +82,7 @@ export const OcmClusterDetailsFormFields = ({
             ? uniqueOcmClusterNameValidationMessages(t)
             : ocmClusterNameValidationMessages(t)
         }
+        maxLength={54}
```
celdrake, line 85: It would be better to introduce a new constant for the maxLength we allow and to use it here, and also in `nameValidationSchema`

If it ever changes, we won't need to remember to keep the two in sync or to find all references which need to be updated.

celdrake (changes requested):

celdrake (approved):

ammont82 (approved):

PR #1892, opened 2023-01-27
MGMT-13470 - Allow support for both LVMO and LVMS depending on Openshift version

jgyselov:
https://issues.redhat.com/browse/MGMT-13470
2.14 Backport: #1893 

### Operators step

- LVM(S) always enabled by default
- `featureSupportLevels` are used to determine if we display LVM (for version 4.11 and level `limited`) or LVMS (for versions 4.12+ and level `supported`)
- in < 4.11 LVM is hidden (feature support level `unsupported`)

https://user-images.githubusercontent.com/87187179/215061058-f2b0e839-0e50-439c-a60b-01c3379e2369.mp4

- in 4.12+ LVMS disabled and checked when CNV is selected

https://user-images.githubusercontent.com/87187179/215061088-0e9a4df2-93f7-44e4-8780-870b9f7718dc.mp4

### Review step

- 4.12+:

image

- 4.11:

image

Edit: returned back the Dev-Preview badge:

image


ammont82 (dismissed):

src/ocm/components/clusterConfiguration/operators/utils.tsx
```diff
@@ -0,0 +1,26 @@
+import {
+  OPERATOR_NAME_LVM,
+  Cluster,
+  FeatureSupportLevelData,
+  ExposedOperatorName,
+  OPERATOR_NAME_LVMS,
+} from '../../../../common';
+
+export const handleLVMS = ({
```
celdrake, line 9: The name is not clear, I would rename it to something like

`getActualLVMOperatorName`

src/ocm/components/clusterConfiguration/operators/LvmCheckbox.tsx
```diff
@@ -1,35 +1,43 @@
...
   const { t } = useTranslation();
-  const operatorName = operatorLabels(t)[OPERATOR_NAME_LVM];
+  const [operator, setOperator] = React.useState('');
+
+  const featureSupportLevel = useFeatureSupportLevel();
+  React.useEffect(() => {
+    setOperator(handleLVMS({ openshiftVersion, featureSupportLevel }));
+  }, [featureSupportLevel, openshiftVersion]);
+
+  const operatorName = operatorLabels(t)[operator];
```
celdrake, line 31:
```suggestion
  const operatorLabel = operatorLabels(t)[operator];
```
as `operator` is in reality an "operator name"

src/ocm/components/clusterConfiguration/operators/LvmCheckbox.tsx
```diff
@@ -1,35 +1,43 @@
...
   return (
     <>
       Install {operatorName}{' '}
       <PopoverIcon
         component={'a'}
         headerContent="Additional Requirements"
-        bodyContent={<LvmHostRequirements clusterId={props.clusterId} />}
+        bodyContent={<LvmHostRequirements clusterId={clusterId} />}
       />
-      <FeatureSupportLevelBadge featureId="LVM" openshiftVersion={props.openshiftVersion} />
```
celdrake, line 32: LVM in 4.11 is "dev-preview", shouldn't we still display it?
As is an operator that cannot be upgraded etc

src/ocm/components/featureSupportLevels/featureStateUtils.ts
```diff
@@ -92,16 +94,6 @@ const getOdfDisabledReason = (cluster: Cluster | undefined, isSupported: boolean
   return undefined;
 };
 
-const getLvmDisabledReason = (cluster: Cluster | undefined, isSupported: boolean) => {
```
celdrake, line 95: Removing this condition seems wrong.
For clusters using < 4.11, LVM should be shown, but should be disabled as it's unsupported.
We must also put back the double if in `LvmCheckbox`

celdrake (changes requested):
There are now some conflicts due to the changes I merged about ARM.

We must ensure these are compatible - for any version, if using ARM, operators should be disabled (LVM Or LVMS). This change should not be ported to v2.14 though.

src/ocm/components/clusterConfiguration/operators/LvmCheckbox.tsx
jgyselov: Added back-

src/ocm/components/featureSupportLevels/featureStateUtils.ts
jgyselov: I'll keep this removed. For <4.11 the checkbox will be hidden based on `/feature-support-levels` and this function will not be used anywhere.

src/ocm/components/clusterConfiguration/operators/utils.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const getActualLVMOperatorName = ({
+  openshiftVersion,
+  featureSupportLevel,
+  operator = OPERATOR_NAME_LVM,
+}: {
+  openshiftVersion: Cluster['openshiftVersion'];
+  featureSupportLevel: FeatureSupportLevelData;
+  operator?: ExposedOperatorName;
+}) => {
+  if (
```
celdrake, line 18: Since we are using this function in cases where `operator` is not LVM, the code has a case for which it doesn't work.

If `operator !== OPERATOR_NAME_LVM`, it needs to return that same operator name.

celdrake (changes requested):

ammont82 (approved):

celdrake (approved):

PR #1895, opened 2023-01-27
Fix useInfraEnv() for missing clusterId (new cluster flow)

mareklibra:
In the new cluster flow, the clusterId does not exist. Called from `FeatureSupportLevelProvider`.


celdrake (commented):
Thanks for fixing this!

However, I'm wondering if it would be better to move this to `useInfraEnvId` since it's where we try to fetch the infraEnv by looking at the `clusterId`.

mareklibra:
@celdrake , np, I moved it there.

celdrake (approved):

jkilzi (approved):

PR #1900, opened 2023-01-31
MGMT-11073: Block hostname field while action is in progress

celdrake:
https://issues.redhat.com/browse/MGMT-11073

After clicking on "change", the field for the host name was still editable, so users could keep typing on it even though it would have no effect.

We now disable the field while an action is in progress.

https://user-images.githubusercontent.com/829045/215818835-595ff6eb-e630-4e50-b414-bb8757f039e4.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "prepare": "install-peers",
     "test": "run-s test:circular lint format:check",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 ./src/index.ts",
-    "lint": "eslint . --max-warnings=102 --color",
+    "lint": "eslint . --max-warnings=98 --color",
     "lint:fix": "yarn run lint --fix",
     "format": "prettier --write '**/*.{json,md,scss,yaml,yml}'",
     "format:check": "prettier --check '**/*.{json,md,scss,yaml,yml}'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "prepare": "install-peers",
     "test": "run-s test:circular lint format:check",
     "test:circular": "dpdm --warning false --tree false --exit-code circular:1 ./src/index.ts",
-    "lint": "eslint . --max-warnings=102 --color",
+    "lint": "eslint . --max-warnings=98 --color",
```
celdrake, line 40: This is by removing the error alert and changing it to  `AlertFormikError`.
I will see separately if there are more cases that we can remove like that.

src/common/components/hosts/EditHostForm.tsx
```diff
@@ -56,6 +57,7 @@ const EditHostForm = ({
   getEditErrorMessage,
 }: EditHostFormProps) => {
   const hostnameInputRef = React.useRef<HTMLInputElement>();
+  const [isEditInProgress, setIsEditInProgress] = React.useState<boolean>(false);
```
jgyselov, line 60: Could we rename to `isSaveInProgress`, `setIsSaveInProgress`? With the _edit_ word I think "the user is editing the form" but in this case it seems to be "the user's changes being applied" for which the word _save_ seems more fitting.

jgyselov (changes requested):

src/common/components/hosts/EditHostForm.tsx
celdrake: Makes sense, I'll rename it

jgyselov (approved):

ammont82 (approved):

PR #1901, opened 2023-02-01
MGMT-13226 - Tooltip should not appear when user hovers over the informational question mark

jgyselov:
https://issues.redhat.com/browse/MGMT-13226

Before:
image

After:

image
image

Network and stack type controls were handled similarly. However controls with helper texts are a bit trickier, so for now I'm leaving them as is.
image

### DIFF
--- a/src/common/components/clusterWizard/networkingSteps/NetworkTypeControlGroup.tsx
+++ b/src/common/components/clusterWizard/networkingSteps/NetworkTypeControlGroup.tsx
@@ -18,8 +18,8 @@ export const NetworkTypeControlGroup = ({
 }: NetworkTypeControlGroupProps) => {
   const { t } = useTranslation();
   return (
-    <FormGroup fieldId={GROUP_NAME} label="Network type">
-      <Split hasGutter>
+    <FormGroup isInline fieldId={GROUP_NAME} label="Network type">
+      <Split>
         <SplitItem>
           <Tooltip
             hidden={isSDNSelectable}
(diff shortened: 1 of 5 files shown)


src/common/components/clusterWizard/networkingSteps/NetworkTypeControlGroup.tsx
```diff
@@ -32,36 +32,33 @@ export const NetworkTypeControlGroup = ({
...
-              }
+              label={<>{t('ai:Software-Defined Networking (SDN)')}&nbsp;</>}
             />
           </Tooltip>
         </SplitItem>
-        <SplitItem />
+        <SplitItem>
+          <PopoverIcon
+            bodyContent={'The classic bullet-proof networking type'}
+            buttonStyle={{ top: '-4px' }}
```
celdrake, line 42: Is it possible that we don't need `buttonStyle` if we keep `noVerticalAlign` on the `PopoverIcon`?

Plus, I see later instead of `top: -4px` we are using `marginTop: '4px'`. I didn't check if `noVerticalAlign` would work there, but I'd try to keep everything as consistent as possible.

jgyselov: I wish it was, but `noVerticalAlign` doesn't really work in this case. At least if we want the icon to stay aligned with the text just as it was before.

Also `top: -4px` and `marginTop: 4px` are fixing two different problems - in one case, the icon is too high, and in the other, it's too low.

celdrake: Understood, if it's not possible to unify them let's keep your changes. Thanks!

celdrake (approved):

ammont82 (approved):

PR #1902, opened 2023-02-01
NO-ISSUE: Fix table console errors

celdrake:
Fixes the console errors and warnings we were receiving from the `Table` component due to the missing `aria-label` at the table level, and the `rowKey` at the row level.

### DIFF
--- a/src/ocm/components/clusterConfiguration/review/ReviewOperatorsTable.tsx
+++ b/src/ocm/components/clusterConfiguration/review/ReviewOperatorsTable.tsx
@@ -27,8 +27,8 @@ export const ReviewOperatorsTable = ({ cluster }: { cluster: Cluster }) => {
           operator,
         }),
       )
-      .map((operator, index) => ({
-        rowId: `operator-${operator}-${index}`,
+      .map((operator) => ({
+        rowId: `operator-${operator}`,
         cells: [
           operatorNames[operator],
           {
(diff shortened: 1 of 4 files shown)


src/ocm/components/clusterConfiguration/review/ReviewOperatorsTable.tsx
```diff
@@ -26,7 +27,8 @@ export const ReviewOperatorsTable = ({ cluster }: { cluster: Cluster }) => {
           operator,
         }),
       )
-      .map((operator) => ({
+      .map((operator, index) => ({
+        rowId: `operator-${operator}-${index}`,
```
jgyselov, line 31: Will we have duplicate operator ids? If not then using the index seems obsolete.

celdrake: Ahh, it did in my tests for LVM(S) as I mocked the data and acted as if all existing operators where active.
In reality that situation cannot happen, as only one of LVM / LVMS would be active at the same time.

I'll remove the index.

ammont82 (approved):
This PR is very helpful because console errors were very annoying. Good job!

jgyselov (approved):

jgyselov (approved):

ammont82 (approved):

PR #1905, opened 2023-02-03
MGMT-13564: Change CPU architecture selection to a Dropdown

celdrake:
https://issues.redhat.com/browse/MGMT-13564

In preparation to support different CPU architectures that the ones we currently support, we need to change the selection of CPU architecture in the Cluster Details step to a Dropdown.

The fields are now also reorganized, so that CpuArchitecture is directly below OpenShift version (As changing the version can impact the available CPU architectures), and because those are both Dropdowns.

See the new behaviour here:
https://user-images.githubusercontent.com/829045/216600290-0ef329db-dcfe-462d-b36a-8b5904b7dd37.mp4


src/common/components/clusterWizard/clusterDetailsValidation.ts
```diff
@@ -74,7 +72,7 @@ export const getClusterDetailsInitialValues = ({
     diskEncryptionMode: cluster?.diskEncryption?.mode ?? 'tpmv2',
     diskEncryptionTangServers: parseTangServers(cluster?.diskEncryption?.tangServers),
     diskEncryption: cluster?.diskEncryption ?? {},
-    cpuArchitecture: cluster?.cpuArchitecture || CpuArchitecture.x86,
+    cpuArchitecture: cluster?.cpuArchitecture || getDefaultCpuArchitecture(),
```
celdrake, line 75: Instead of a hardcoded default cpuArchitecture, we now have a place where we can configure it so that everywhere will use that default.
This needs to be implemented later, for now it's just about preparing the code for that change.

jgyselov (dismissed):

ammont82 (approved):

ammont82 (approved):

jgyselov (approved):

PR #1909, opened 2023-02-07
MGMT-13244 - Add host ssh keys does not allow space or \n for next key

jgyselov:
https://issues.redhat.com/browse/MGMT-13244

- SSH upload field now allows spaces and \n
- Spaces and empty lines get trimmed on blur
- Newlines are used as separators between SSH keys
- SSH public keys must be unique

These changes shouldn't affect uses of the UploadSSH component outside `/ocm`.


ammont82 (approved):

celdrake (approved):
I tested it again, and for some reason the error I saw earlier is not happening anymore.
So all good!

PR #1916, opened 2023-02-13
MGMT-13266: Show missing feature support levels error in R&C

celdrake:
https://issues.redhat.com/browse/MGMT-13334

For clusters using an OpenShift version which does not include feature-support-level details, the wizard would thrown an error in two places: Networking page or Review & Create. The error was not clear, since this error is a misconfiguration of the OpenShift version in the environment.

- The error is now clearer
- Now the error page will only be shown in the Review & Create page, to prevent the user from installing a cluster with an unknown set of unsupported features. In the Networking page, it was only used to display the warning about "only using VMs".

review-error

### DIFF
--- a/src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
+++ b/src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
@@ -1,7 +1,14 @@
 import React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { ActionListItem, Button, ButtonVariant, Grid, GridItem } from '@patternfly/react-core';
-import { Cluster, ClusterWizardStepHeader, useAlerts, ClusterWizardStep } from '../../../../common';
+import {
+  Cluster,
+  ClusterWizardStepHeader,
+  useAlerts,
+  ClusterWizardStep,
+  ErrorState,
+  useFeatureSupportLevel,
+} from '../../../../common';
 import { useClusterWizardContext } from '../../clusterWizard/ClusterWizardContext';
 import { getApiErrorMessage, handleApiError } from '../../../api';
 import { updateCluster } from '../../../reducers/clusters';
@@ -12,6 +19,9 @@ import { useStateSafely } from '../../../../common/hooks';
 import { selectCurrentClusterPermissionsState } from '../../../selectors';
 import ReviewPreflightChecks from './ReviewPreflightChecks';
 import ReviewSummary from './ReviewSummary';
+import { getLimitedFeatureSupportLevels } from '../../../../common/components/featureSupportLevels/utils';
+import { useTranslation } from '../../../../common/hooks/use-translation-wrapper';
+
 import './ReviewCluster.css';
 
 const ReviewStep = ({ cluster }: { cluster: Cluster }) => {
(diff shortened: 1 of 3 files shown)


jgyselov (dismissed):

ammont82 (approved):

celdrake:
Had to rebase because by dropping the errors there was 1 more in master and the pipeline was failing.

ammont82 (approved):

jgyselov (dismissed):

ammont82 (dismissed):

src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
```diff
@@ -12,15 +19,29 @@ import { useStateSafely } from '../../../../common/hooks';
...
+  const { t } = useTranslation();
   const dispatch = useDispatch();
 
+  const hasSupportLevel = React.useMemo<boolean>(() => {
+    try {
+      getLimitedFeatureSupportLevels(cluster, featureSupportLevelData, t);
+      return true;
+    } catch (e) {
+      return false;
+    }
```
jkilzi, line 42: Why do we need 2 return statements?

```suggestion
    let theAnswer = false;
    try {
      theAnswer = getLimitedFeatureSupportLevels(cluster, featureSupportLevelData, t);
    } catch {
      // pass...
    }
    
    return theAnswer;
```

jkilzi (changes requested):

src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
```diff
@@ -40,6 +61,17 @@ const ReviewStep = ({ cluster }: { cluster: Cluster }) => {
     }
   };
 
+  if (!hasSupportLevel) {
+    return (
+      <ErrorState
+        title={'Feature support levels not available'}
+        content={`OpenShift version ${
+          cluster.openshiftVersion || ''
+        } does not currently include feature support levels.`}
```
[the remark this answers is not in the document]
celdrake: You are right, this message should not be seen by users.
In any case, we tried to make it as clear as possible in case some developer has this issue due to some problem in their environment configuration. The message was checked with Allison.

src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
celdrake: Fixed it

src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
```diff
@@ -12,15 +19,30 @@ import { useStateSafely } from '../../../../common/hooks';
...
   const { isViewerMode } = useSelector(selectCurrentClusterPermissionsState);
   const [isStartingInstallation, setIsStartingInstallation] = useStateSafely(false);
+  const featureSupportLevelData = useFeatureSupportLevel();
+  const { t } = useTranslation();
   const dispatch = useDispatch();
 
+  const hasSupportLevel = React.useMemo<boolean>(() => {
+    let hasSupportLevels = false;
+    try {
+      hasSupportLevels = !!getLimitedFeatureSupportLevels(cluster, featureSupportLevelData, t);
```
jkilzi, line 39: Do you still need the try-catch after adding the !!
?

celdrake: Yes, the current implementation of `getLimitedFeatureSupportLevels` is that it can throw errors in a few cases and in fact, it will throw one in the case we're handling here (misconfiguration of the dev environment and feature-support-levels missing).

The casting etc doesn't prevent the error from being thrown.

ammont82 (approved):

src/ocm/components/clusterConfiguration/review/ReviewStep.tsx
```diff
@@ -12,15 +19,30 @@ import { useStateSafely } from '../../../../common/hooks';
...
+    } catch (e) {
+      // Missing support levels
```
jkilzi, line 41:
```suggestion
      Sentry.captureException(e);
```
Would this work?

jkilzi (changes requested):

celdrake:
It's now moved to https://github.com/openshift-assisted/assisted-ui-lib/pull/1976

PR #1925, opened 2023-02-15
MGMT-13470 - Allow support for both LVMO and LVMS depending on Openshift version

jgyselov:
https://issues.redhat.com/browse/MGMT-13470

(add support in Operator progress items)

**Before:**
image

**After:**
image


ammont82 (dismissed):

src/common/components/clusterDetail/OperatorsProgressItem.tsx
```diff
@@ -72,25 +74,32 @@ export function getOperatorsIcon(status: OperatorStatus | 'pending') {
...
       bodyContent={
         <List className="operators-progress-item__operators-list">
-          {operators.map((operator) => {
+          {operators.map((operator: MonitoredOperator) => {
             let status = operator.status ?? 'pending';
             if (operator.status === 'available') {
               status = 'installed';
             }
-            const operatorName = operator.name as OperatorName;
+            const operatorName = getActualLVMOperatorName({
```
celdrake, line 98: I see this fixes the problem in the cluster installation page, however there are other places where we are using `operatorLabels` and we may not have adapted the operator name.

So I'm thinking that it would be better to move the logic to adapt the operator name to the `operatorLabels`, and not have to remember to call the `getActualLVMOperatorName` everytime we want the operator name.

jgyselov: Good idea. :+1:

src/common/config/constants.ts
```diff
@@ -304,12 +305,22 @@ export const ExposedOperatorNames = [
 export type OperatorName = typeof OperatorNames[number];
 export type ExposedOperatorName = typeof ExposedOperatorNames[number];
 
-export const operatorLabels = (t: TFunction): { [key in ExposedOperatorName]: string } => ({
-  [OPERATOR_NAME_ODF]: t('ai:OpenShift Data Foundation'),
-  [OPERATOR_NAME_LVM]: t('ai:Logical Volume Manager'),
-  [OPERATOR_NAME_LVMS]: t('ai:Logical Volume Manager Storage'),
-  [OPERATOR_NAME_CNV]: t('ai:OpenShift Virtualization'),
-});
+export const operatorLabels = (
```
celdrake, line 308: There are still some old usages that are not passing the new parameters, please fix those too.

I checked and apparently `FailedOperatorsWarnings` is not being used anymore, so we can remove this file. HOwever, `ProgressBarAlerts` should be refactored.

celdrake (changes requested):

src/common/config/constants.ts
jgyselov: `FailedOperatorsWarnings` file removed. `ProgressBarAlerts` fixed. So now all usages of  the `operatorLabels` functions should be covered.

celdrake (approved):

ammont82 (approved):

PR #1877, opened 2023-01-19
Fix typescript 4 errors part 4

celdrake:
More errors for Typescript v4

PR #1927, opened 2023-02-15
MGMT-13677: Fix non-production warning only for appropriate versions

celdrake:
https://issues.redhat.com/browse/MGMT-13677

The warning about the version not being production-ready has stopped being displayed.

The previous code relied on the matching of the labels, but for Beta version, it wasn't finding a version that matched, because the label had some extra text ("Developer Preview release"). 

What finally broke it is this change https://github.com/openshift-assisted/assisted-ui-lib/pull/1877/files#diff-e9b056fd8eab90b0a2096e8e64661a8cb4106723abb935e099cb9faa23b18117R43
but IMO if we don't find the match, we should not show the warning. There should always be a match, as the selected version must be on of the dropdown values.

Now we are matching on the value which is always the same, no matter if we append some text to the rendered item.

### DIFF
--- a/src/common/components/ui/OpenShiftVersionDropdown.tsx
+++ b/src/common/components/ui/OpenShiftVersionDropdown.tsx
@@ -24,12 +24,10 @@ export type HelperTextType = (
 
 type OpenShiftVersionDropdownProps = {
   name: string;
-  defaultValue: string | undefined;
   items: {
     label: string;
     value: string;
   }[];
-  isDisabled: boolean;
   versions: OpenshiftVersionOptionType[];
   getHelperText: HelperTextType;
   showReleasesLink: boolean;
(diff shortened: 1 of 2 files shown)


src/common/components/ui/OpenShiftVersionDropdown.tsx
```diff
@@ -24,32 +24,39 @@ export type HelperTextType = (
...
+  const { defaultLabel, defaultValue } = React.useMemo(() => {
+    const defaultVersion = versions.find((item) => item.default);
+    return {
+      defaultLabel: defaultVersion?.label || '',
+      defaultValue: defaultVersion?.value || '',
+    };
+  }, [versions]);
+
+  const [helperText, setHelperText] = React.useState(getHelperText(versions, defaultValue, t));
+  const [current, setCurrent] = React.useState<string>(defaultLabel);
```
jkilzi, line 58: `current` what? ;-P

jkilzi (approved):

jkilzi:
@celdrake millestone?

src/common/components/ui/OpenShiftVersionDropdown.tsx
celdrake: Agree, it should be `currentLabel` but it's like that in possibly all Dropdowns and I would rename it for all at a later point.

celdrake:
> @celdrake millestone?

This one is for master (v2.16), but I need to create a backport for `v2.15` too

ammont82 (approved):

PR #1932, opened 2023-02-20
Fix no-unsafe-argument errors

celdrake:
Fixing errors of type `no-unsafe-argument`


src/common/components/ui/formik/types.ts
```diff
@@ -101,7 +101,6 @@ export interface UploadFieldProps extends FieldProps {
   dropzoneProps?: Omit<DropzoneProps, 'onDropRejected'> & {
     onDropRejected?: (helpers: FieldHelperProps<string>) => DropFileEventHandler;
   };
-  transformValue?: (key: string) => void;
```
celdrake, line 104: This prop had only been passed from OCM's UploadSSH field, but it was recently removed.

jgyselov (dismissed):

jgyselov (approved):

ammont82 (approved):

PR #1956, opened 2023-03-02
MGMT-12331 - Remove Format column in the Host discovery page

jgyselov:
https://issues.redhat.com/browse/MGMT-12331

**Before:**
image

**After:**
image

### DIFF
--- a/src/common/components/storage/DisksTable.tsx
+++ b/src/common/components/storage/DisksTable.tsx
@@ -29,11 +29,11 @@ interface DisksTableProps extends WithTestID {
   updateDiskSkipFormatting?: DiskFormattingType;
 }
 
-const diskColumns = [
+const diskColumns = (showFormat: boolean) => [
   { title: 'Name' },
   { title: 'Role' },
   { title: 'Limitations' },
-  { title: 'Format?' },
+  showFormat ? { title: 'Format?' } : '',
   { title: 'Drive type' },
   { title: 'Size' },
   { title: 'Serial' },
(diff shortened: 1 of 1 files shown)


ammont82 (approved):

src/common/components/storage/DisksTable.tsx
```diff
@@ -29,11 +29,11 @@ interface DisksTableProps extends WithTestID {
   updateDiskSkipFormatting?: DiskFormattingType;
 }
 
-const diskColumns = [
+const diskColumns = (showFormat: boolean) => [
```
jkilzi, line 32:
```suggestion
const diskColumns = (showFormat: boolean) => [
  { title: 'Name' },
  { title: 'Role' },
  { title: 'Limitations' },
  showFormat ? { title: 'Format?' } : false,
  { title: 'Drive type' },
  { title: 'Size' },
  { title: 'Serial' },
  { title: 'Model' },
  { title: 'WWN' },
].filter(Boolean);
```

jkilzi (changes requested):

src/common/components/storage/DisksTable.tsx
jgyselov: Maybe not... 
image

`.filter(Boolean)` changes the type of the array and PT table doesn't like that. The current version works and I would keep it as is.

jkilzi (approved):

ammont82 (approved):

PR #1966, opened 2023-03-07
MGMT-9166 - Help text disappears once there is an error

jgyselov:
https://issues.redhat.com/browse/MGMT-9166

**Some examples of input fields with helper and error texts:**

image
image
image
image
image
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openshift-assisted-ui-lib",
-  "version": "2.16.3",
+  "version": "0.0.0",
   "description": "React component library for the Bare Metal Installer",
   "license": "Apache-2.0",
   "repository": "openshift-assisted/assisted-ui-lib",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openshift-assisted-ui-lib",
-  "version": "2.16.0",
+  "version": "2.16.1",
   "description": "React component library for the Bare Metal Installer",
   "license": "Apache-2.0",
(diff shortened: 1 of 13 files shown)


jgyselov:
@nirfarkas PTAL

ammont82 (approved):

ammont82 (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "openshift-assisted-ui-lib",
-  "version": "0.0.0",
+  "version": "2.16.3",
```
rawagner, line 3: changing the version is a mistake, right ?

src/common/components/ui/formik/InputField.tsx
```diff
@@ -24,7 +32,7 @@ const InputField: React.FC<
       inputError,
       description,
       labelInfo,
-      showErrorMessage = true,
+      // showErrorMessage = true,
```
rawagner, line 35: there is couple of places where this prop is still used. I think if we change
```
{errorMessage && (
  <HelperText fieldId={fieldId} isError>
    {errorMessage}
  </HelperText>
)}
```

to
```
{errorMessage && showErrorMessage && (
   <HelperText fieldId={fieldId} isError>
     {errorMessage}
   </HelperText>
)}
```
we should be good

rawagner (changes requested):

package.json
jgyselov: Yes.

ammont82 (approved):

rawagner (approved):

PR #2041, opened 2023-04-20
MGMT-14307 - Day2 - In add hosts - setting "control plane" role to host but UI marks it as control plane + worker

jgyselov:
https://issues.redhat.com/browse/MGMT-14307

Day1 should remain unaffected.

**Before:**

image

**After:**

image

### DIFF
--- a/src/common/components/hosts/tableUtils.tsx
+++ b/src/common/components/hosts/tableUtils.tsx
@@ -89,6 +89,7 @@ export const hostnameColumn = (
 
 export const roleColumn = (
   t: TFunction,
+  clusterKind: Cluster['kind'],
   canEditRole?: HostsTableActions['canEditRole'],
   onEditRole?: HostsTableActions['onEditRole'],
   schedulableMasters?: boolean,
@@ -107,7 +108,7 @@ export const roleColumn = (
         ? (role: HostUpdateParams['hostRole']) => onEditRole(host, role)
         : undefined;
       const isRoleEditable = canEditRole?.(host);
-      const hostRole = getHostRole(host, t, schedulableMasters);
+      const hostRole = getHostRole(host, t, clusterKind, schedulableMasters);
       return {
         title: (
           <RoleCell
--- a/src/common/components/hosts/tableUtils.tsx
+++ b/src/common/components/hosts/tableUtils.tsx
@@ -23,7 +23,6 @@ import {
 } from './utils';
 import { selectMachineNetworkCIDR } from '../../selectors/clusterSelectors';
 import { hostStatus } from './status';
-import { DropdownProps } from '@patternfly/react-core';
 import { TFunction } from 'i18next';
 
 export const getSelectedNic = (nics: Interface[], currentSubnet: Address4 | Address6) => {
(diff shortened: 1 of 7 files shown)


jkilzi (approved):

src/common/components/hosts/utils.ts
```diff
@@ -120,15 +120,23 @@ export const canHostnameBeChanged = (hostStatus: Host['status']) =>
...
   }`;
 
   if (schedulableMasters && host.role === 'master') {
-    roleLabel = t('ai:Control plane node, Worker');
+    roleLabel =
+      clusterKind === 'Cluster' ? t('ai:Control plane node, Worker') : t('ai:Control plane node');
   }
-  return `${roleLabel}${host.bootstrap ? ' (bootstrap)' : ''}`;
+  const role = `${roleLabel}${host.bootstrap ? ' (bootstrap)' : ''}`;
+  console.log(host, schedulableMasters, roleLabel, role);
```
jkilzi, line 138:
```suggestion
```
;-p

jgyselov: omggg

jgyselov: Anyway, I don't know what you are talking about. This never happened.

ammont82 (dismissed):

jkilzi (approved):

ammont82 (approved):

src/common/components/hosts/utils.ts
```diff
@@ -120,13 +120,19 @@ export const canHostnameBeChanged = (hostStatus: Host['status']) =>
     'pending-for-input',
   ].includes(hostStatus);
 
-export const getHostRole = (host: Host, t: TFunction, schedulableMasters?: boolean): string => {
+export const getHostRole = (
```
rawagner, line 123: the function is called from `StorageUtils.tsx` which is not passing the new `clusterKind` prop

src/common/components/hosts/tableUtils.tsx
```diff
@@ -83,6 +83,7 @@ export const hostnameColumn = (
 
 export const roleColumn = (
   t: TFunction,
+  clusterKind: Cluster['kind'],
```
rawagner, line 86: there is couple of places all over the source code, where the new `clusterKind` is not passed

rawagner (changes requested):
But Im unsure about the OCM part, i dont know what the `clusterKind` `"Cluster" | "AddHostsCluster"` means. Somebody OCM-aware needs to review :)

src/common/components/hosts/tableUtils.tsx
jgyselov: In the end, I just switched the position on the parameter, made it optional and changed the condition in `getHostRole`. Things should be working now even for other usages of these functions.

ammont82 (dismissed):

ammont82 (dismissed):

src/common/components/hosts/tableUtils.tsx
jgyselov: @celdrake I removed the `position` parameter because it was never passed to `roleColumn`. It's optional in `RoleCell`, `RollDropdown` and further on. 

I suppose it could be removed from those as well, since we use the default position anyway.

celdrake: If that's the case, then yes, we should be cleaning-up the unused property. Otherwise there are left-overs and in the future it's harder to track what's being used and what isn't

celdrake (dismissed):
Thanks for cleaning up the unused props!

ammont82 (dismissed):

rawagner (approved):

PR #2051, opened 2023-04-25
MGMT-14131 - When AI service is down, AI cluster list is hang with loading sign

jgyselov:
https://issues.redhat.com/browse/MGMT-14131

If the error happens on load, we show an in-page error:

image

If the error happens on reload, we show a modal:

image

### DIFF
--- a/src/ocm/reducers/clusters/clustersSlice.ts
+++ b/src/ocm/reducers/clusters/clustersSlice.ts
@@ -12,15 +12,13 @@ import { handleApiError, ocmClient } from '../../api';
 
 export const fetchClustersAsync = createAsyncThunk<Cluster[] | void>(
   'clusters/fetchClustersAsync',
-  async () => {
+  async (value, { rejectWithValue }) => {
     try {
       const { data } = await ClustersAPI.list();
       const isOcm = !!ocmClient;
       return data.filter((cluster) => (isOcm ? cluster.kind === 'Cluster' : true));
     } catch (e) {
-      handleApiError(e, () => {
-        void Promise.reject('Failed to fetch clusters.');
-      });
+      return handleApiError(e, () => rejectWithValue(value));
     }
   },
 );
--- a/src/ocm/components/clusterDetail/ClusterPollingErrorModal.tsx
+++ b/src/ocm/components/clusterDetail/ClusterPollingErrorModal.tsx
@@ -1,10 +1,21 @@
 import React from 'react';
 import { Button, ButtonVariant, Modal, ModalVariant } from '@patternfly/react-core';
 
-const ClusterPollingErrorModal = () => {
+const defaultTitle = 'Unable to fetch the latest data';
+
+const defaultContent =
+  "There was an error fetching the latest data from the API. We'll try again, but if the data does not update in a few minutes, try refreshing the page.";
+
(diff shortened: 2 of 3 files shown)


ammont82 (dismissed):

ammont82 (approved):

ammont82 (approved):

src/ocm/reducers/clusters/clustersSlice.ts
```diff
@@ -12,7 +12,7 @@ import { handleApiError, ocmClient } from '../../api';
 
 export const fetchClustersAsync = createAsyncThunk<Cluster[] | void>(
   'clusters/fetchClustersAsync',
-  async () => {
+  async (_, { rejectWithValue }) => {
```
rawagner, line 15: lets rename to something meaningful..ie `value`
```suggestion
  async (value, { rejectWithValue }) => {
```

rawagner (dismissed):

src/ocm/reducers/clusters/clustersSlice.ts
jgyselov: Done.

ammont82 (approved):

src/ocm/reducers/clusters/clustersSlice.ts
```diff
@@ -21,6 +21,7 @@ export const fetchClustersAsync = createAsyncThunk<Cluster[] | void>(
       handleApiError(e, () => {
         void Promise.reject('Failed to fetch clusters.');
       });
+      return rejectWithValue(value);
```
celdrake, line 24: Would it work removing this line and updating?
```

 return handleApiError(e, () => {
        return Promise.reject(value);
      });
});
```
Or alternatively, with the code:
```
handleApiError(e);
return rejectWithValue(value);
```

either way, I think it's best not to duplicate the logic for rejecting the promise, while making sure it will be rejected, or the `POLLING_ERROR` will not be registered.

ammont82: @celdrake WDYT? 
In ocm/reducers/clusters/currentClusterSlice.ts, the code is:

```
return handleApiError(e, () => {
      return rejectWithValue({
        code: getApiErrorCode(e as Error | AxiosError),
        message: getApiErrorMessage(e),
      });
    });

Comment (celdrake) [reply src/ocm/reducers/clusters/clustersSlice.ts:24]: It's fine with me if it works as it should. The main idea is to reject the promise only once.
I can't say what the effect is of rejecting the the `code/message` object instead of `value`. I thought `value` was the message we wanted to display.

Comment (ammont82) [reply src/ocm/reducers/clusters/clustersSlice.ts:24]: @jgyselov can you try it?

Comment (jgyselov) [reply src/ocm/reducers/clusters/clustersSlice.ts]: I've settled on 
```
return handleApiError(e, () => rejectWithValue(value));
```
(we also don't really care what `value` is).

ammont82 (approved):

rawagner (approved):

src/ocm/reducers/clusters/clustersSlice.ts
celdrake: Great, thanks!

PR #2069, opened 2023-04-28
MGMT-12656: CIM: Append table actions to content-scrollable element

rawagner:
This will work only on OCP 4.13 as we need latest pf/react-table


jtomasek (approved):

rawagner:
already fixed in master by https://github.com/openshift-assisted/assisted-installer-ui/pull/1919. Closing

PR #2094, opened 2023-05-05
MGMT-14508 - Slowness in search text

jgyselov:
https://issues.redhat.com/browse/MGMT-14508

- An unnecessary dependency list item was causing the UI to make the requests twice. I couldn't reproduce the "slowness in text search" issue that was reported, but this fix should improve the experience, too.
- Moved the loader from the modal contents to the header
- When two requests are made close to each other, the first one gets aborted (removes flashing)

**New video:**

https://github.com/openshift-assisted/assisted-installer-ui/assets/87187179/33134c20-f452-4902-9b77-73a81b1d512c


ammont82 (approved):

ammont82 (dismissed):

libs/ui-lib/lib/common/components/ui/eventsModal.tsx
```diff
@@ -80,9 +89,17 @@ export const EventsModal = ({
   fallbackEventsURL,
 }: EventsModalProps) => {
   const { t } = useTranslation();
+  const [isLoading, setLoading] = useStateSafely(true);
+
   return (
     <Modal
-      title={title}
+      header={
+        <>
```
jkilzi, line 97: Do we need the fragment here?

jgyselov: Probably not.

jkilzi (approved):

ammont82 (approved):

PR #2095, opened 2023-05-05
Change the API alerts position in the "Add hosts" modal

jgyselov:
https://issues.redhat.com/browse/MGMT-14177

**Before:**

image

**After:**

image

I added a dismiss button because the SSH key error can be quite lengthy:

image


nirfarkas:
@jgyselov I don't think this should be dismissible. We will need to address the content of the errors though.

PR #2103, opened 2023-05-08
Adds the latest translations for ko,ja,zh

jkilzi:


jtomasek:
Let's hold this one, we're still syncing 2.8-cim branch with master so this PR may not be needed.

PR #2114, opened 2023-05-11
MGMT-14574: CIM: Allow empty proxy fields

rawagner:


ammont82 (dismissed):

jtomasek (dismissed):

ammont82 (dismissed):

libs/ui-lib/lib/common/components/clusterConfiguration/DiscoveryImageConfigForm.tsx
```diff
@@ -51,14 +51,15 @@ export const StaticIPInfo: React.FC = () => {
...
 
-const validationSchema = Yup.lazy<DiscoveryImageFormValues>((values) =>
-  Yup.object<DiscoveryImageFormValues>().shape({
-    sshPublicKey: sshPublicKeyValidationSchema,
-    httpProxy: httpProxyValidationSchema(values, 'httpsProxy'),
-    httpsProxy: httpProxyValidationSchema(values, 'httpProxy'), // share the schema, httpS is currently not supported
-    noProxy: noProxyValidationSchema,
-  }),
-);
+const validationSchema = (allowEmpty: boolean) =>
```
mareklibra, line 54: nit:
```suggestion
const getValidationSchema = (allowEmpty: boolean) =>
```

libs/ui-lib/lib/common/components/ui/formik/validationSchemas.ts
```diff
@@ -507,14 +507,12 @@ const httpProxyValidationMessage = 'Provide a valid HTTP URL.';
...
   pairValueName: 'httpProxy' | 'httpsProxy',
-) =>
-  Yup.string()
-    .test(
-      'http-proxy-no-empty-validation',
-      'At least one of the HTTP or HTTPS proxy URLs is required.',
-      (value) => Boolean(!values.enableProxy || value || values[pairValueName]),
-    )
-    .test('http-proxy-validation', httpProxyValidationMessage, (value: string) => {
+  allowEmpty?: boolean,
```
mareklibra, line 510: About codestyle only: having an optional boolean param is not good for the maintenance since we can not add more params later and it is not a big deal to pass `false` here now.
In addition, it makes sense to review other usages of that and decide about the value there.

libs/ui-lib/lib/common/components/clusterConfiguration/DiscoveryImageConfigForm.tsx
rawagner: fixed

libs/ui-lib/lib/common/components/ui/formik/validationSchemas.ts
rawagner: I've changed the input params to be an object. Which still enables us to have optional params & add more later :)
All other usages of the `httpProxyValidationSchema` were reviewed.

mareklibra (approved):

PR #2116, opened 2023-05-11
MGMT-14577: CIM: Set 'online' to true by default

rawagner:


nirfarkas:
@rawagner what happens if I check this "online" box?

ammont82 (approved):

libs/ui-lib/lib/cim/components/Agent/BMCForm.tsx
```diff
@@ -183,7 +184,7 @@ const emptyValues: AddBmcValues = {
   password: '',
   bootMACAddress: '',
   disableCertificateVerification: true, // TODO(mlibra)
-  online: false, // TODO(mlibra)
+  online: false,
```
jtomasek, line 187: Should it be 'true' by default?

jtomasek: IIUC the bug is caused by the fact that the BMH is created with `online: false` which causes the host to never go through discovery because it does not get booted. That is why I think it should be `online: true` by default. Is there ever an use case in CIM when user would want the host to be in `online: false`?

jtomasek (approved):

ammont82 (approved):

PR #2123, opened 2023-05-17
MGMT-14640: InfraEnvs cache return type hotfix

jgyselov:
`InfraEnvCache.getInfraEnvId()` is now returning `Error` instead of `null`, the condition in `InfraEnvsService` needs to be changed accordingly.

Otherwise the UI is stuck loading in case the infra env hasn't been cached previously.


jkilzi (approved):

jkilzi:
We just released 2.20, please back port this PR to the releases/v2.20.
I'll create a patch (2.20.1) for it + add a JIRA issue and link this PR to it, fix version 2.20.1

ammont82 (approved):

PR #2127, opened 2023-05-17
MGMT-14325: Avoid visual jump when going through the "Add hosts" modal steps

jgyselov:
https://issues.redhat.com/browse/MGMT-14325

**Before:**

https://github.com/openshift-assisted/assisted-installer-ui/assets/87187179/9718b4bc-00f4-408a-ad6e-8e1a30880e6a

**After:**

https://github.com/openshift-assisted/assisted-installer-ui/assets/87187179/506d5e7c-89eb-4574-adae-48b31bdfd79c


ammont82 (approved):

nirfarkas:
Looks great!
/approve

libs/ui-lib/lib/ocm/components/clusterConfiguration/DiscoveryImageModal.css
```diff
@@ -0,0 +1,3 @@
+#generate-discovery-iso-modal {
+  height: 791px;
+}
```
jkilzi, line 3: I'm not sure this is the way to go, can you try this instead?https://www.patternfly.org/v4/components/modal/#scrollable-modals

jkilzi (changes requested):

libs/ui-lib/lib/ocm/components/clusterConfiguration/DiscoveryImageModal.css
jgyselov: I can add `tabIndex={0}` to the Modal, no problem. But doing that alone doesn't fix our issue with the changing height of the modal - the CSS is needed here.

jkilzi: ok, what about using one of these classes: https://www.patternfly.org/v4/utilities/sizing

jkilzi: Try removing `hasNoBodyWrapper` then play with the sizing classes ^. I managed to get it working by applying: `.pf-u-min-height` + `style="--pf-u-min-height--MinHeight: 60ch;"` on the body.

jgyselov: I tried with the `minHeight` as you suggested... and I'm not convinced.

It can look fine...

https://github.com/openshift-assisted/assisted-installer-ui/assets/87187179/42c40d59-8f61-42db-b1e1-e437fcc28add

But again there is the problem that we cannot predict the modal height without knowing the data, so we again get that jump. Less jarring, less annoying, but it's still there.

https://github.com/openshift-assisted/assisted-installer-ui/assets/87187179/a2966bc3-e48b-43f5-b7c6-fb62635e1665

---

I mean it depends on what we are trying to achieve here. If the goal is to remove any jumps then we need to go with my solution and set the `height`. But we are okay with some jumps then we can just increase the minimum height of the modal.

Let me know your thoughts.

cc @nirfarkas

jkilzi: Maybe we are misusing the modal, maybe we should seek for a "wizard-like" solution.
The problem here is that we render a modal and then in the same modal we render a complete different content. Maybe the this is not what PF want us to do with it...

jkilzi: I'm just trying to reason about the need to tweak the component's CSS in order to achieve this behavior, this tweak is not scalable and will look awkward in different screen sizes.

jkilzi (approved):

PR #2149, opened 2023-05-23
Quick edit II

cadawson:
I did a quick edit of the text in this doc and proposed updates. The first entry of a line, I left intact, in case that was the variable name. I updated the second entry on the line with the updates. Let me know if you have questions. There is also another PR that should cover the first 100 lines.


rawagner (dismissed):

jkilzi (approved):

jkilzi:
@rawagner PTAL

libs/locales/lib/en/translation.json
```diff
@@ -788,11 +788,12 @@
...
-  "ai:You can either wait or investigate. A common issue can be misconfigured storage. Once you fix the issue, you can delete AgentServiceConfig to try again.": "You can either wait or investigate. A common issue can be misconfigured storage. Once you fix the issue, you can delete AgentServiceConfig to try again.",
+  "ai:You are approving multiple hosts. All hosts listed below will be approved to join the infrastructure environment if you continue. Make sure that you expect and recognize the hosts before approving.": "You are approving multiple hosts. All hosts listed will be approved to join the infrastructure environment, if you continue. Make sure that you expect and recognize the hosts before approving.",
+  "ai:You are deleting multiple resources. All resources listed below will be deleted if you continue. This action cannot be undone.": "You are deleting multiple resources. All resources listed will be deleted if you continue. This action cannot be undone.",
```
rawagner, line 792: not sure where this came from, but it needs to be removed

jkilzi: Which line? Or both, 786+787?

rawagner: 787

jkilzi:
```suggestion
```

jkilzi (changes requested):

jkilzi (approved):

libs/locales/lib/en/translation.json
jkilzi: Addressed in 5c3f45f499e59b691a6ec8e6e56e2a81a7e1bde7

jkilzi: Addressed in 5c3f45f499e59b691a6ec8e6e56e2a81a7e1bde7

jkilzi (approved):

PR #2153, opened 2023-05-24
MGMT-14197: Fix catastrophic backtracking in no proxy dns regex

rh-gvincent:
PROXY_DNS_REGEX regex validation timeout.
The . (dot) in the group 1 of the old regex introduce an extremely large number of permutations. Moving the dot to the second group in the regex.

The wildcard regex is also not needed, because we already returned true if the value equals '*'.


jkilzi (approved):

jkilzi:
@rh-gvincent can you please port this change to the `releases/v2.20` branch?

PR #2157, opened 2023-05-25
CIM: Fix hypershift scaling and mapping of AgentMachines to HostedClu…

rawagner:
…ster


libs/ui-lib/lib/cim/components/Hypershift/modals/ManageHostsModal.tsx
```diff
@@ -57,11 +57,6 @@ const ManageHostsModal = ({
         path: '/spec/replicas',
       });
     }
-    patches.push({
```
rawagner, line 60: this patch is not needed as the dialog does not allow chaing the agent labels.

ammont82 (approved):

PR #2171, opened 2023-05-30
MGMT-14703: Limit and verify type when allowing to upload files / yaml to day2

jgyselov:
https://issues.redhat.com/browse/MGMT-14703

image


jkilzi (approved):

jkilzi:
The validation message in the screenshot above is a bit awkward since the validation failure has nothing to do with the file extension but with the file content. I've updated it and now it reads: **"Unsupported file type. Please provide a valid YAML file."**

nirfarkas:
I think we're supporting YAML, YML and JSON file types

PR #2185, opened 2023-06-06
MGMT-14658: Generate iso modal - not able to drag and drop ssh key [BACKPORT]

jgyselov:
https://issues.redhat.com/browse/MGMT-14658

### DIFF
--- a/apps/assisted-ui/package.json
+++ b/apps/assisted-ui/package.json
@@ -39,6 +39,11 @@
     "vite": "^4.0.0",
     "vite-plugin-environment": "^1.1.3"
   },
+  "overrides": {
+    "@patternfly/react-core": {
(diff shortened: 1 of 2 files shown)


libs/ui-lib/package.json
```diff
@@ -101,6 +101,11 @@
     "uuid": "^8.1",
     "yup": "^0.28.3"
   },
+  "overrides": {
+    "@patternfly/react-core": {
+      "attr-accept": "2.2.2"
+    }
+  },
```
jkilzi, line 108: I don't think we need them here we are no longer installing peer dependencies.
Deps in this section get installed by CIM, assisted-ui or uhc-portal

jkilzi (approved):

ammont82 (approved):

PR #2211, opened 2023-06-19
MGMT-14986: External partners remains selected after OCP < v4.14 is selected

jkilzi:


ammont82 (approved):

jkilzi:
Converting it to draft in order to retrieve the checkbox status from the support-level API instead from the hard-coded constant

PR #2226, opened 2023-06-26
Introduce vitest for unit-testing

mareklibra:
[Vitest](https://vitest.dev/api/) is introduced. More tests will be added in follow-ups.

To run existing unit tests, call
```
yarn test:unit
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -48,7 +48,8 @@
       "extends": ["@openshift-assisted/eslint-config-assisted-ui"],
       "files": ["libs/ui-lib/lib/**/*.{ts,tsx}"],
       "parserOptions": {
-        "tsconfigRootDir": "libs/ui-lib"
+        "tsconfigRootDir": "libs/ui-lib",
+        "project": "tsconfig.eslint.json"
       },
       "rules": {
         "no-restricted-imports": [
--- a/libs/ui-lib/lib/common/components/ui/formik/validationSchemas.test.ts
+++ b/libs/ui-lib/lib/common/components/ui/formik/validationSchemas.test.ts
@@ -1,0 +1,6 @@
+import { test } from 'vitest';
+import { noProxyValidationSchema } from './validationSchemas';
+
+test('noProxyValidationSchema runs', async () => {
+  await noProxyValidationSchema.validate('.y.com');
+});
--- a/libs/ui-lib/tsconfig.eslint.json
+++ b/libs/ui-lib/tsconfig.eslint.json
@@ -1,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": []
+}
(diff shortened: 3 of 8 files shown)


ammont82 (approved):

libs/ui-lib/tsconfig.eslint.json
```diff
@@ -0,0 +1,4 @@
+{
```
jkilzi, line 1: Why do we need this? What was your intent?

libs/ui-lib/lib/common/components/ui/formik/validationSchemas.test.ts
```diff
@@ -0,0 +1,6 @@
+import { test } from 'vitest';
+import { noProxyValidationSchema } from './validationSchemas';
+
+test('noProxyValidationSchema runs', async () => {
+  await noProxyValidationSchema.validate('.y.com');
```
jkilzi, line 5: What should anyone looking at this test understand about this test?
What's the expected value/result?
Under what circumstance the test should pass/fail?

jkilzi (changes requested):

libs/ui-lib/lib/common/components/ui/formik/validationSchemas.test.ts
mareklibra: It's a stub, I have a pending PR actually implementing that particular test, it will follow right after we agree on the infra here. I was just lazy to do `Hello, world`.

libs/ui-lib/tsconfig.eslint.json
mareklibra: We `do not` want the test files in the build. We `do` want to lint the test files.

The linter needs to have all files matching for linting to be included in the tsconfig selection. One way how to tackle it is creating a new tsconfig for linter while reusing the "main" tsconfig and just easing the `exclude` section.

.eslintrc.json
```diff
@@ -48,7 +48,8 @@
       "extends": ["@openshift-assisted/eslint-config-assisted-ui"],
       "files": ["libs/ui-lib/lib/**/*.{ts,tsx}"],
       "parserOptions": {
-        "tsconfigRootDir": "libs/ui-lib"
+        "tsconfigRootDir": "libs/ui-lib",
+        "project": "tsconfig.eslint.json"
```
mareklibra, line 52: Here

libs/ui-lib/tsconfig.eslint.json
mareklibra: https://github.com/openshift-assisted/assisted-installer-ui/pull/2226/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR52

https://github.com/openshift-assisted/assisted-installer-ui/pull/2226/files#diff-f98ebae61e31cb1a9fff4b3bb9f3475a7a7b44b5a1e8332eca456a9e0e85f833R23

jkilzi (approved):

PR #2230, opened 2023-06-27
MGMT-14576: Fix noProxy validation

mareklibra:


jgyselov (approved):

ammont82 (commented):
@mareklibra I've changed this validation to avoid catastrophic backtracking problems https://github.com/openshift-assisted/assisted-installer-ui/pull/2180. Have you tested if your new validation avoids these problems?

jkilzi:
> @mareklibra I've changed this validation to avoid catastrophic backtracking problems #2180. Have you tested if your new validation avoids these problems?

@ammont82 this is a different regex, I ran the one here through https://devina.io/redos-checker, and seems safe.
I think it would be better to search for a library we can use in order to perform these kinds of validations as part of the unit-tests. So every time we add a new regex we could write a test that uses this library in order to verify the pattern is safe...

jkilzi (approved):

PR #2232, opened 2023-06-28
Day2 Integration tests

celdrake:
Added basic Integration tests for the Day2 flow.

Created a new page, exposed in the Router, that can only be used in a mocked environment.
- When it's accessed and it's not mocked, it displays a message.
non-mocked

 - When it's accesses and it's mocked, it will show two buttons, that represent the tab for Adding hosts. 
  One button represents when it's clicked and the cluster does not have metrics yet, while the other when it already has them.
mocked

Included some small refactorings too, like not having to use the entire `ocmClient` to determine if the application is in OCM or not.


libs/ui-lib-tests/cypress/fixtures/infra-envs/openshift-versions.ts
```diff
@@ -42,17 +47,20 @@ const versions = {
   },
 };
 
+expect(
```
celdrake, line 50: There were two versions which had the "default: true" flag

ammont82 (approved):

libs/ui-lib/lib/ocm/api/axiosClient.ts
```diff
@@ -15,22 +15,25 @@ const axiosCaseConverterOptions = {
...
-  } else {
-    return applyCaseMiddleware(client, axiosCaseConverterOptions);
-  }
+  return client;
 };
 
-let client: AxiosInstance = getDefaultClient();
-let clientWithoutConverter: AxiosInstance = getDefaultClient(true);
+let client = applyCaseMiddleware(getDefaultClient(), axiosCaseConverterOptions);
+let clientWithoutConverter = getDefaultClient();
```
jkilzi, line 34: @ammont82 please remind me why do we need an Axios instance without the case converter?

libs/ui-lib/lib/ocm/components/AddHosts/AddHosts.tsx
```diff
@@ -72,7 +72,7 @@ export const AddHosts = () => {
 
   return (
     <ModalDialogsContextProvider>
-      <Card>
+      <Card label="add-hosts-tab-content">
```
jkilzi, line 75: What is the `label` attribute used for?

celdrake: It was not used, I removed it

libs/ui-lib/lib/ocm/api/axiosClient.ts
```diff
@@ -15,22 +15,25 @@ const axiosCaseConverterOptions = {
   },
 };
 
-const getDefaultClient = (withoutConverter = false) => {
+const token = '';
```
jkilzi, line 18: Danger zone!
Is this really necessary?

libs/ui-lib/lib/ocm/api/axiosClient.ts
ammont82: Yes, it'ss used in OCM.

ammont82 (approved):

libs/ui-lib-tests/cypress/fixtures/infra-envs/feature-support.ts
jgyselov, line 1: We should probably create a task for adapting the tests to the new feature support levels API.

jgyselov (changes requested):

libs/ui-lib/lib/ocm/components/HostsClusterDetailTab/HostsClusterDetailTabContentErrors.tsx
```diff
@@ -0,0 +1,66 @@
...
+      We're waiting for your recently installed cluster to report its metrics.{' '}
+      <Button variant={'link'} isInline onClick={onTryAgain}>
+        Try again
+      </Button>{' '}
+      in a few minutes.
+    </>
+  );
+};
+
+const UnsupportedVersion = ({ version, onTryAgain }: { version: string } & TryAgainProps) => {
```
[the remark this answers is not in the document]
celdrake: That's a good point, I'll rename to keep the names consistent.

libs/ui-lib-tests/cypress/support/interceptors.ts
```diff
@@ -1,28 +1,37 @@
...
-import createDualStackFixtureMapping from '../fixtures/dualstack';
-import createStaticIpFixtureMapping from '../fixtures/static-ip';
 import { HttpRequestInterceptor } from 'cypress/types/net-stubbing';
-import createCustomManifestsFixtureMapping from '../fixtures/custom-manifests';
-import { getEventHeaders, getEvents } from '../fixtures/cluster/events';
+
+import { hasWizardSignal, setLastWizardSignal } from './utils';
+import * as allFixtures from '../fixtures';
+
+const { day2FlowIds } = allFixtures;
```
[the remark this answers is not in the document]
celdrake: Yes, that works. I can rename to `fixtures`

libs/ui-lib/lib/ocm/api/axiosClient.ts
celdrake: I'll revert the file and only add the `isInOcm` to avoid passing the entire `ocmClient` in many different components.

libs/ui-lib/lib/ocm/api/axiosClient.ts
celdrake: I'll revert, I did a small refactor while I was looking for the way to make the Cypress tests trigger the `isInOcm` flag.

libs/ui-lib-tests/cypress/integration/create-multinode/2-networking.cy.ts
```diff
@@ -1,36 +1,37 @@
...
 import { networkingPage } from '../../views/networkingPage';
-import * as utils from '../../support/utils';
-import { NetworkingRequest } from '../../fixtures/create-mn/requests';
+import { getNetworkingRequest } from '../../fixtures/create-mn/requests';
 
 describe(`Assisted Installer Multinode Networking`, () => {
   before(() => {
     cy.loadAiAPIIntercepts({
-      activeSignal: 'HOST_RENAMED_1',
+      activeSignal: 'READY_TO_INSTALL',
```
jgyselov, line 8: I spent a lot of time in my PR (https://github.com/openshift-assisted/assisted-installer-ui/pull/2233) making changes to the validations info fixtures to make the mocks closer resemble the actual life-cycle of a cluster. This feels like a step back.

When the user first accesses the Networking page, the cluster doesn't have `machineNetworks`, `apiVips`, and `ingressVips` defined. These values are not provided automatically by the BE, the users need to enter them manually. Until then, there are a lot of pending validations that prevent the user from continuing past the Networking page.

Maybe we could do something like this here:

```

describe(`Assisted Installer Multinode Networking`, () => {
  describe('Before finishing networking configuration', () => {
    before(() => {
      cy.loadAiAPIIntercepts({
        activeSignal: 'HOST_RENAMED_3',
        activeScenario: 'AI_CREATE_MULTINODE',
      });
    });

    beforeEach(() => {
      cy.loadAiAPIIntercepts(null);
      commonActions.visitClusterDetailsPage();
      // because of the failing networking validations, a cluster at this stage will start on the Host discovery page
      commonActions.moveNextSteps(['Host discovery', 'Storage']);
    });

    it('Should submit updated network information', () => {
     ...
    });
  });

  describe('After finishing networking configuration', () => {
    before(() => {
      cy.loadAiAPIIntercepts({
        activeSignal: 'READY_TO_INSTALL',
        activeScenario: 'AI_CREATE_MULTINODE',
      });
    });

    beforeEach(() => {
      cy.loadAiAPIIntercepts(null);
      commonActions.visitClusterDetailsPage();
      commonActions.startAtWizardStep('Networking');
    });

    it('Should see the Ready Host inventory status', () => {
     ...
    });
    ...
  });
});
```

jgyselov (changes requested):

libs/ui-lib-tests/cypress/integration/create-multinode/2-networking.cy.ts
celdrake: You are right, my changes did change the approach.
I have updated the tests with two describe as you suggest, and added some additional assertions to make sure we're testing what the tests say. PTAL.

libs/ui-lib-tests/cypress/fixtures/create-mn/requests.ts
```diff
@@ -24,11 +24,11 @@ const UMNetworkingRequest = {
   network_type: 'OVNKubernetes',
 };
 
-const NetworkingRequest = {
-  api_vips: [{ ip: '192.168.122.10', cluster_id: fakeClusterId }],
-  api_vip: '192.168.122.10',
-  ingress_vips: [{ ip: '192.168.122.110', cluster_id: fakeClusterId }],
-  ingress_vip: '192.168.122.110',
+const getNetworkingRequest = (apiVip: string, ingressVip: string) => ({
```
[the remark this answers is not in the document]
celdrake: I had changed it because since I signalled `READY_TO_INSTALL`, those Vips were equal to the default ones, so the test had to verify that the entered Vips were submitted.
Since I have changed the tests to use two describe blocks, I've also reverted this change.

libs/ui-lib-tests/cypress/integration/create-multinode/2-networking.cy.ts
jgyselov: LGTM

jgyselov (approved):

ammont82 (approved):

PR #2240, opened 2023-07-03
[MGMT-15152] Fix jump bug and adapt tests

celdrake:
https://issues.redhat.com/browse/MGMT-15152

- While writing Integration tests for Day2, I noticed some intermitent test failures.
Sometimes, the test would click on the nav wizard and it would not move to the expected step. This is due to the effect that loads the custom manifests, and sometimes it would jump to the customManifests / review step even if the user had selected a previous step.

- I also noted that some requests were not correctly intercepted (eg. delete of a custom manifest includes the folder and filename, and the POST was not mocked.

- Finally, did some changes in the custom manifests tests to align them to use the "lastWizardSignal" as the other tests do.

### DIFF
--- a/libs/ui-lib-tests/cypress/integration/custom-manifests/2-modifying-existing-custom-manifest.cy.ts
+++ b/libs/ui-lib-tests/cypress/integration/custom-manifests/2-modifying-existing-custom-manifest.cy.ts
@@ -1,36 +1,53 @@
 import { commonActions } from '../../views/common';
+import { setLastWizardSignal } from '../../support/utils';
 import { customManifestsPage } from '../../views/customManifestsPage';
 const ACTIVE_NAV_ITEM_CLASS = 'pf-m-current';
 
 describe(`Assisted Installer Custom manifests step`, () => {
   before(() => {
     cy.loadAiAPIIntercepts({
-      activeSignal: 'READY_TO_INSTALL',
+      activeSignal: 'ONLY_DUMMY_CUSTOM_MANIFEST_ADDED',
       activeScenario: 'AI_CREATE_CUSTOM_MANIFESTS',
     });
   });
 
   beforeEach(() => {
     cy.loadAiAPIIntercepts(null);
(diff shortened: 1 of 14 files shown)


libs/ui-lib-tests/cypress/integration/custom-manifests/2-modifying-existing-custom-manifest.cy.ts
```diff
@@ -1,36 +1,53 @@
...
 
   beforeEach(() => {
     cy.loadAiAPIIntercepts(null);
+    setLastWizardSignal('ONLY_DUMMY_CUSTOM_MANIFEST_ADDED');
     commonActions.visitClusterDetailsPage();
   });
 
-  describe('Modifying existing Custom Manifests', () => {
-    it('Can configure custom manifests step and next button is disabled', () => {
+  describe('Initial step for Custom Manifests', () => {
```
celdrake, line 20: This "describe" blocks handles the bug.

libs/ui-lib-tests/cypress/support/interceptors.ts
```diff
@@ -110,16 +110,13 @@ const mockInfraEnvResponse: HttpRequestInterceptor = (req) => {
 
 const mockCustomManifestResponse: HttpRequestInterceptor = (req) => {
   const fixtureMapping = getScenarioFixtureMapping();
-  if (fixtureMapping?.manifests) {
-    req.reply(fixtureMapping.manifests);
-  }
+  req.reply(fixtureMapping?.manifests || []);
```
celdrake, line 113: The API should always respond, even if it's an empty array of data.

ammont82 (approved):

ammont82:
@celdrake I've tested the custom manifests workflow and it works well. 
Thanks for your patch.

jkilzi (approved):

jkilzi (approved):

PR #2243, opened 2023-07-03
MGMT-14908: Cleanup the ui-lib-tests project

jkilzi:
Basically deletes a lot of unused code that was mostly inherited from kni-auto.


libs/ui-lib-tests/cypress/support/interceptors.ts
```diff
@@ -169,7 +169,7 @@ const addClusterListIntercepts = () => {
   cy.intercept('GET', allClustersApiPath, (req) => {
     const fixture = hasWizardSignal('CLUSTER_CREATED') ? updatedList() : initialList;
     req.reply(fixture);
-  });
+  }).as('list-clusters');
```
celdrake, line 172: I have done quite a big refactoring in my branch for Day2 to `interceptors.ts`.
Since you're only adding aliases, I think it's best to revert this file in your branch and we can add the missing aliases later (I added some there too, but maybe not all of them)

celdrake (approved):

ammont82 (approved):

PR #2258, opened 2023-07-14
MGMT-15410: Add unit tests for OCM wizard transitions

mareklibra:


ammont82 (dismissed):

libs/ui-lib/lib/ocm/components/clusterWizard/wizardTransition.test.ts
```diff
@@ -0,0 +1,472 @@
...
+    // based on clusterDetailsStepValidationsMap
+    const cluster: Cluster = cloneDeep(clusterBase);
+    const validationInfo = cloneDeep(validationInfoClusterDetails);
+
+    expect(canNextClusterDetails({ cluster })).toBe(false);
+
+    cluster.status = 'ready';
+    expect(canNextClusterDetails({ cluster })).toBe(true);
+
+    cluster.status = clusterBase.status;
```
celdrake, line 265: I would rather extract each case into its own "test" unit.

This would allow:
 - To have a self-explanatory title of what each test is verifying
 - To make sure each test is independent (you could have a `beforeEach` that defines the initial cluster and validation info, therefore nothing from a previous test can affect a subsequent test)
 - To be able to see at a glance what's being covered in each "describe" block and what may be missing. Currently, you would need to read the entire test to understand that.

rawagner (approved):

libs/ui-lib/lib/ocm/components/clusterWizard/wizardTransition.test.ts
rawagner: I believe this is done. Resolving

jgyselov (approved):

PR #2269, opened 2023-07-20
MGMT-12926: Grouping disks by holders

jgyselov:
https://issues.redhat.com/browse/MGMT-12926

- Disks with the drive type "LVM" are hidden.
- The disks are first sorted by name and then by holders (or name again if the disk is a holder itself).
- Disks with **one** defined holder are indented. This should result in a visual hierarchy:

image

- If a disk with holders is selected as the installation disk, we show a warning icon and a popover. The content depends on the holder's drive type (LVM, RAID, multipath).
image

- Under the storage table, we show a new alert in case there are limitations for the selected installation disk.
image

- In case multiple hosts have installation disk limitations, we show the warnings in a list.
image


jkilzi (approved):

ammont82 (approved):

libs/ui-lib-tests/cypress/integration/storage/storage-step-disk-holders.cy.ts
```diff
@@ -0,0 +1,92 @@
...
+    setTestStartSignal('READY_TO_INSTALL');
+  });
+
+  describe(`Host storage`, () => {
+    beforeEach(() => {
+      setTestStartSignal('READY_TO_INSTALL');
+      commonActions.visitClusterDetailsPage();
+      commonActions.startAtWizardStep('Storage');
+
+      storagePage = new StorageForm();
```
jkilzi, line 42:
```suggestion
      storageForm = new StorageForm();
```

jkilzi (changes requested):

jgyselov:
The `/views` part of the tests will need a refactor. But for the time being, it does the job.

libs/ui-lib/lib/common/components/storage/DisksTable.tsx
```diff
@@ -53,6 +60,77 @@ const SkipFormattingDisk = () => (
...
+    return `The installation disk ${diskName as string} selected for host ${
+      hostName as string
+    }, is part of a software RAID that will be deleted during the installation.`;
+  } else if (holder?.driveType === 'Multipath') {
+    return `The installation disk ${diskName as string} selected for host ${
+      hostName as string
+    } is managed by multipath. We strongly recommend using the multipath device ${
+      holder.name as string
+    } to improve reliability.`;
+  }
```
jkilzi, line 82: If `holder` is not defined the code falls through the conditions.
I would recommend first checking if the `holders` value is truthy and then using a switch block on the driveType in order to improve readability.

libs/ui-lib/lib/common/components/storage/DisksTable.tsx
```diff
@@ -53,6 +60,80 @@ const SkipFormattingDisk = () => (
...
+      {'   '}
+      {disk.bootable ? `${disk.name || ''} (bootable)` : disk.name}
+      {diskLimitations && (
+        <>
+          {'    '}
+          <Popover
+            headerContent="Warning"
+            bodyContent={<Alert variant={AlertVariant.warning} isInline title={diskLimitations} />}
+            minWidth="20rem"
+            maxWidth="30rem"
```
jkilzi, line 126:
```suggestion
            minWidth="20rem"
            maxWidth="30rem"
```
Please try using one of the `pf-u-*` classes instead

jkilzi (approved):

libs/ui-lib/lib/common/components/storage/DisksTable.tsx
jgyselov: PatternFly provides these props, it should be ok to use them.

Also I think the `pf-u-*` classes are for percentages which isn't really what we need here.

ammont82 (approved):

PR #2284, opened 2023-08-01
MGMT-15414: When "No partner integration" is selected allow the BE to determine the platform type

jgyselov:
https://issues.redhat.com/browse/MGMT-15414


libs/ui-lib/lib/ocm/services/ClusterDetailsService.ts
```diff
@@ -57,6 +57,10 @@ const ClusterDetailsService = {
       params.tags = AI_UI_TAG;
     }
 
+    if (params.platform?.type === 'none') {
+      delete params.platform;
+    }
+
```
jkilzi, line 63: This won't be enough. The initialValues should also be refactored

jkilzi (changes requested):

libs/ui-lib-tests/cypress/integration/create-multinode/0-create-cluster.cy.ts
```diff
@@ -28,10 +28,13 @@ describe(`Assisted Installer Multinode Cluster Installation`, () => {
 
       commonActions.getInfoAlert().should('not.exist');
       commonActions.toNextStepAfter('Cluster details');
-      cy.wait('@create-cluster');
+
+      cy.wait('@create-cluster').then(({ request }) => {
+        expect(Object.keys(request.body)).not.to.contain('platform');
+      });
```
jkilzi, line 34: Following [this discussion](https://redhat-internal.slack.com/archives/C05KWUC1E2X/p1690974339940609), we should assert that if the platform is set and the type is 'baremetal' then user_managed_networking = false, maybe this is the right place to do it.

jgyselov: This is definitely not the right place to do that. (`with-user-managed-networking.cy.ts` might be better)

But following that discussion it isn't 100% that we will be setting baremetal+UMN from the cluster details. So let's not rush this.

jkilzi (approved):

jkilzi (dismissed):

libs/ui-lib/lib/ocm/components/clusterConfiguration/platformIntegration/constants.tsx
```diff
@@ -0,0 +1,19 @@
+import { NUTANIX_CONFIG_LINK, VSPHERE_CONFIG_LINK } from '../../../../common';
+import { PlatformType } from '../../../api';
+
+export const ExternalPlatformLabels: { [key in PlatformType]: string } = {
```
celdrake, line 4: I like that you extracted this data into a `constants` module.

Due to that, now the data in `ClusterPlatformIntegrationHint` should be unnecessary, and since the component is not used anywhere, it should be deleted entirely.

Usages of `integrationPlatformLinks` should refer to this new file.

celdrake (changes requested):
Looks good :100: Just a small extra clean-up!

celdrake (approved):

jkilzi (approved):

PR #2286, opened 2023-08-04
MGMT-15054: FCP devices (multipath) do not show correct size

jgyselov:
https://issues.redhat.com/browse/MGMT-15054

If a disk has defined holders, it is not included in the total storage size calculations.


celdrake (commented):
The change looks good and the tests changes too.

However, I have noticed that the function `createHostInventoryWithDiskHolders` does not really create fixtures with the expected `master/workerDisk`. It overwrites the `disks` with fixed data defined as `disksWithHolders`.

I suggest to fix it, perhaps by changing the function `createHostInventoryWithDiskHolders` (removing the declared `diskSpace` param, and send a dummy `diskSpace` value (eg. 0) to `createHostInventory`.). Or make `diskHolders` dynamic and use the received diskSpace.

jgyselov:
> However, I have noticed that the function `createHostInventoryWithDiskHolders` does not really create fixtures with the expected `master/workerDisk`. It overwrites the `disks` with fixed data defined as `disksWithHolders`.
>
> I suggest to fix it, perhaps by changing the function `createHostInventoryWithDiskHolders` (removing the declared `diskSpace` param, and send a dummy `diskSpace` value (eg. 0) to `createHostInventory`.). Or make `diskHolders` dynamic and use the received diskSpace.

I decided to remove the `diskSpace` parameter. I don't have the best understanding of how exactly the host inventory data works, I'm just using mock data provided [here](https://issues.redhat.com/browse/MGMT-12926) for the disks with holders.

ammont82 (approved):

jkilzi (approved):

PR #2289, opened 2023-08-12
MGMT-15251: Link to infra env page instead of networking in case of n…

rawagner:
…tp failure


ammont82 (dismissed):

libs/ui-lib/lib/cim/components/ClusterDeployment/wizardTransition.ts
```diff
@@ -1,3 +1,4 @@
+import { Host } from '../../../common';
```
jkilzi, line 1:
```suggestion
import { Host } from "@openshift-assisted/types/assisted-installer-service";
```

jkilzi (approved):

ammont82 (approved):

PR #2299, opened 2023-08-21
MGMT-11177: Update UX of Dropdowns to match Patternfly style

jgyselov:
https://issues.redhat.com/browse/MGMT-11177


ammont82 (dismissed):

jgyselov:
Base domain had the last usage of the SelectField in the OCM part.

SelectField is still used [here](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/cim/components/Hypershift/HostedClusterWizard/HostsStep/HostsForm.tsx#L46) in /cim and [here](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/common/components/clusterConfiguration/OpenShiftVersionSelect.tsx#L71) and [here](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx#L92) in /common (none of which are consumed by OCM). I've decided against doing any changes to these files because I don't have a way of testing them.

jkilzi (approved):

ammont82 (approved):

PR #2328, opened 2023-09-06
MGMT-15709: OCI option must be exposed only to selected users

jkilzi:
The issue was introduced in #2326


libs/ui-lib/lib/common/features/featureGate.tsx
ammont82, line 1: ASSISTED_INSTALLER_MULTIARCH_SUPPORTED is not needed. We only need ASSISTED_INSTALLER_PLATFORM_OCI

ammont82 (approved):

libs/ui-lib/lib/common/features/featureGate.tsx
jkilzi: we still need it

PR #2374, opened 2023-09-19
Use the UI-settings API to denote that the custom manifests step should be shown

jgyselov:
https://issues.redhat.com/browse/MGMT-15659


libs/ui-lib/lib/common/api/assisted-service/UISettingsAPI.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const config = {
+  headers: {
+    'Content-Type': 'application/json',
+  },
+};
+
+const UISettingsAPI = {
+  makeBaseURI(clusterId?: Cluster['id']) {
+    return `/v2/clusters/${clusterId ? clusterId : ''}/ui-settings`;
```
jkilzi, line 12: Why is clusterId optional? You cannot call this API if you don't know the cluster ID...

libs/ui-lib/lib/ocm/components/clusterConfiguration/CustomManifestCheckbox.tsx
```diff
@@ -46,30 +46,20 @@ const CustomManifestCheckbox = ({ clusterId, isDisabled }: CustomManifestCheckbo
...
-        if (clusterId) setDeleteCustomManifestsOpen(true);
-        else {
-          cleanCustomManifests();
-        }
-      } else {
-        setValue(checked);
-        setManifestsRemoved(false);
-        clusterWizardContext.setAddCustomManifests(checked);
-      }
+      clusterWizardContext.setCustomManifestsStep(checked);
```
jkilzi, line 55: PTAL at line [77](https://github.com/openshift-assisted/assisted-installer-ui/pull/2374/commits/ba99039db3beff1789eccaa9949ac6233b374800#diff-a4b0fa62e02584dcfa2e8fedae16a7e2f632bf1dec4bc17ea797be076e8a0403R77) despite this function is inside a useCallback, there we are creating another arrow function.

ammont82:
@jgyselov I've tests and I've found that when you go to Custom manifests page the first time and the manifest is empty you can click "Next button". In theory, we needs to have all custom manifests filled to go to next page:
image

libs/ui-lib/lib/ocm/hooks/useUISettings.ts
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { Cluster } from '@openshift-assisted/types/./assisted-installer-service';
+import { UISettingService, UISettingsValues } from '../services';
+
+const useUISettings = (clusterId?: Cluster['id']) => {
```
jkilzi, line 5: I must have a `clusterId` otherwise you can't call the API

libs/ui-lib/lib/ocm/hooks/useUISettings.ts
```diff
@@ -0,0 +1,48 @@
...
+  const [uiSettings, setUISettings] = React.useState<UISettingsValues>();
+
+  const updateUISettings = React.useCallback(
+    async (newSettings: UISettingsValues) => {
+      if (clusterId) {
```
jkilzi, line 10: Making `clusterId` mandatory would make this check unnecessary

jkilzi (changes requested):

libs/ui-lib/lib/ocm/hooks/useUISettings.ts
jgyselov: The problem is that I'm using this hook in the `ClusterWizardContext` where cluster can be `undefined`. (e.g. on the New cluster page - we don't have a cluster yet, but we do have a wizard).

So I can make the `clusterId` mandatory in the service and API, but I cannot make it mandatory in this hook.

libs/ui-lib/lib/ocm/hooks/useUISettings.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const returnValue = React.useMemo(
+    () => ({ uiSettings, updateUISettings }),
+    [uiSettings, updateUISettings],
+  );
+
+  return returnValue;
+};
+
+export default useUISettings;
```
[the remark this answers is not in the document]
jgyselov: I added some error handling. PTAL.

ammont82 (approved):

jkilzi (approved):

ammont82 (approved):

jkilzi (approved):

PR #2423, opened 2023-10-24
MGMT-15365: It's possible to attempt to delete a host from infraenv for a standalone cluster created by integrated UI flow

jgyselov:
https://issues.redhat.com/browse/MGMT-15365

image

### DIFF
--- a/libs/ui-lib/lib/cim/components/Agent/tableUtils.tsx
+++ b/libs/ui-lib/lib/cim/components/Agent/tableUtils.tsx
@@ -343,6 +343,10 @@ export const canUnbindAgent = (
     }
   }
 
+  if (!agent.metadata?.labels?.hasOwnProperty(AGENT_BMH_NAME_LABEL_KEY)) {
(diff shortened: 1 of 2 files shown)


libs/ui-lib/lib/cim/components/Agent/tableUtils.tsx
```diff
@@ -343,6 +343,10 @@ export const canUnbindAgent = (
     }
   }
 
+  if (!agent.metadata?.labels?.hasOwnProperty(AGENT_BMH_NAME_LABEL_KEY)) {
```
rawagner, line 346: it looks like this will also disallow removing agents from CIM-flow cluster ?

We need to ensure this gets applied only on AI-flow cluster. To do that you need to get the `InfraEnv` of `Agent` and check if `InfraEnv.spec.clusterRef` is defined.

rawagner (approved):

ammont82 (approved):

PR #2454, opened 2023-12-15
Update setAuthInterceptor to be idempotent

rh-gvincent:
Today if we call setAuthInterceptor multiple times, it's not working because we reuse previous client and clientWithoutCaseConverter. Config url in withAssistedInstallerBasePath for example is applied multiple times.

Normally we call setAuthInterceptor only one time. But with recent module federated code, we call setAuthInterceptor every time we load the component. We also remove the job template created earlier

### DIFF
--- a/libs/ui-lib/lib/common/api/axiosClient.ts
+++ b/libs/ui-lib/lib/common/api/axiosClient.ts
@@ -21,29 +21,37 @@ const withAssistedInstallerBasePath = (client: AxiosInstance): AxiosInstance => 
   return client;
 };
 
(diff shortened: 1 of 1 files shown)


rh-gvincent:
```ts
useEffect(() => {
  auth
    .getUser()
    .then(() => auth.getToken())
    .then((token) => {
      Api.setAuthInterceptor(...);
    })
    .finally(() => setIsLoading(false));
}, []);
```

code we use in federated modules when we reuse the setAuthInterceptor

PR #2507, opened 2024-02-21
MGMT-16970: BMH status not updated in console during provisioning phase

jgyselov:
https://issues.redhat.com/browse/MGMT-16970

Adds missing argument.


libs/ui-lib/lib/cim/components/InfraEnv/InfraEnvAgentTable.tsx
```diff
@@ -162,6 +163,7 @@ const InfraEnvAgentTable: React.FC<InfraEnvAgentTableProps> = ({
       hostActions.canEditHostname,
       hostActions.canEditBMH,
       agentStatuses,
+      bmhStatuses,
```
rawagner, line 166: We should memoize the `const bmhStatuses = bmhStatus(t);` call above, as it is now used as a dependency for this `useMemo`

ammont82 (approved):

PR #2522, opened 2024-03-19
MGMT-17281: CIM UI crashes if the installation_disk_id is not correct

jgyselov:
https://issues.redhat.com/browse/MGMT-17281

### DIFF
--- a/libs/ui-lib/lib/cim/components/Agent/AgentStatus.tsx
+++ b/libs/ui-lib/lib/cim/components/Agent/AgentStatus.tsx
@@ -55,7 +55,7 @@ const AgentStatus: React.FC<AgentStatusProps> = ({
       AdditionalNTPSourcesDialogToggleComponent={AdditionalNTPSourcesDialogToggle}
       autoCSR
       additionalPopoverContent={
-        status.status.key === 'specSyncErr' ? (
+        status.status.key === 'spec-sync-err' ? (
           <StackItem>
             <SpecSyncErrAlert agent={agent} />
           </StackItem>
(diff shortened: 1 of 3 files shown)


ammont82 (dismissed):

libs/ui-lib/lib/cim/components/Agent/AgentStatus.tsx
```diff
@@ -55,7 +55,7 @@ const AgentStatus: React.FC<AgentStatusProps> = ({
       AdditionalNTPSourcesDialogToggleComponent={AdditionalNTPSourcesDialogToggle}
       autoCSR
       additionalPopoverContent={
-        status.status.key === 'specSyncErr' ? (
+        status.status.key === 'spec-sync-err' ? (
```
celdrake, line 58: IMO the following change would make the code less error prone, and it should avoid future issues like the one you're solving now.

Changing the definition in  `helpers/agentStatus.tsx` file from

```
  specSyncErr: {
    key: 'spec-sync-err',
   ...
  },
```
to 
```
  specSyncErr: {
    key: 'specSyncErr',
   ...
  },
```

With this change, the object key and `key` value for this status would match, as it happens with all the other statuses.

jgyselov: This is probably a good idea.

jgyselov: Done.

ammont82 (approved):

PR #2560, opened 2024-04-17
[OCMUI-1145] PreviewBadge: allow ReactNode content

cben:
Motivation: In OCM we have use case for showing dynamic badge content (https://issues.redhat.com/browse/OCMUI-1145) where we don't get separate "Learn More" URL but might render links in the text itself.

The existing `PreviewBadge` component supports this fine at run time, only the `popoverText: string` type was too restrictive at compile time.  
- Relaxed the type.
- Renamed `popoverText` -> `popoverContent` accordingly. Can drop this part if there is any concern?

_With the ongoing de-coupling of our codebases, I'm not sure we'll end up using this — but seems harmless enough?_

- [ ] **UNTESTED** — I don't have AI dev setup, didn't manage to run unit tests even on master (`Missing internal module 'internal/deps/./lib/client'`...)
- [x] `yarn build:all` passed.

### DIFF
--- a/libs/ui-lib/lib/common/components/ui/DeveloperPreview.tsx
+++ b/libs/ui-lib/lib/common/components/ui/DeveloperPreview.tsx
@@ -2,14 +2,14 @@ import React from 'react';
 import { useTranslation } from '../../hooks/use-translation-wrapper';
 import { PreviewBadge, PreviewBadgeProps } from './PreviewBadge';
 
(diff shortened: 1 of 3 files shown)


cben:
@ammont82 please review

ammont82:
@cben If you make these changes the styles are different:
- With your changes:
image

- In production:
image

You can see that font-size is different.

ammont82 (approved):

jgyselov (approved):

PR #2565, opened 2024-04-24
Update openshift-assisted/ui-lib to react-router-dom v6

jgyselov:
https://issues.redhat.com/browse/MGMT-17147


libs/ui-lib/lib/ocm/components/Routes.tsx
```diff
@@ -18,9 +18,9 @@ export const Routes: React.FC<{ allEnabledFeatures: FeatureListType }> = ({
     <Provider store={storeDay1}>
       <AssistedUILibVersion />
       <Switch>
```
daznaurovRH, line 20: While using CompatRouter and changing everyting into v6 Switch is no longer in use. You can use <Routes> component

libs/ui-lib/lib/ocm/components/clusters/Clusters.tsx
```diff
@@ -33,9 +33,8 @@ import {
   selectClusterTableRows,
 } from '../../store/slices/clusters/selectors';
 
-type ClustersProps = RouteComponentProps;
-
-const Clusters: React.FC<ClustersProps> = ({ history }) => {
+const Clusters = () => {
+  const history = useHistory();
```
daznaurovRH, line 37: If I remember correctly same functionality is possible to be achieved by using useLocation and useParams

jgyselov:
Closing in favour of https://github.com/openshift-assisted/assisted-installer-ui/pull/2637

PR #2626, opened 2024-06-28
MGMT-17388: Log string literal format shown with "special" characters

jgyselov:
https://issues.redhat.com/browse/MGMT-17388

image

Replacing `\n` with a space is probably the best we can do here.


ammont82 (approved):

celdrake (approved):
Did you consider splitting the message in the new line and creating a div / li per item?

jgyselov:
> Did you consider splitting the message in the new line and creating a div / li per item?

I did consider it. But given how rare `\n` appearing seems to be (I've only seen this one instance in this screenshot), I don't think it's worth it.

PR #2648, opened 2024-07-31
MGMT-16745: Add platform field to R&C and installation pages

jgyselov:
https://issues.redhat.com/browse/MGMT-16745

image


ammont82:
@jgyselov about the PlatformIntegrationNote:
- In case of Nutanix and vSphere - it's necessary
- In case of OCI - it's not necessary

PR #2655, opened 2024-08-16
NO-ISSUE: Work with OCM-UI router v6 ugprade

jgyselov:


jschuler:
Hi @jgyselov would it be possible to pick up [this change](https://github.com/openshift-assisted/assisted-installer-ui/pull/2653/files#diff-169c2383aac0ed5caeabc1debf944a63049be2e1e3b493535a9d9a003b733235L15) as well? 
it's not completely related so if you prefer can be done separately but it's a very small change :) We are changing "Clusters" to "Cluster List" in OCMUI

jgyselov:
> Hi @jgyselov would it be possible to pick up [this change](https://github.com/openshift-assisted/assisted-installer-ui/pull/2653/files#diff-169c2383aac0ed5caeabc1debf944a63049be2e1e3b493535a9d9a003b733235L15) as well? it's not completely related so if you prefer can be done separately but it's a very small change :) We are changing "Clusters" to "Cluster List" in OCMUI
> 
> Oh! also it should route to `/cluster-list` because we just changed default routing of `/` to route to the Overview page, not Cluster list anymore!
> 
> https://issues.redhat.com/browse/OCMUI-2199

Hi @jschuler, sure, no problem, I can make this change.

But are you sure about the capital L in "Cluster List"? I would have expected sentence case there.

rawagner (dismissed):

jkilzi (approved):

libs/ui-lib/lib/ocm/components/Routes.tsx
```diff
@@ -10,24 +17,38 @@ import { useFeatureDetection } from '../hooks/use-feature-detection';
 export const UILibRoutes = ({
   allEnabledFeatures,
   children,
+  history,
+  basename,
 }: {
   allEnabledFeatures: FeatureListType;
```
jschuler, line 23: since this is a required prop what should OCM-UI pass to it? or can it be made optional?

jschuler:
@jgyselov good question about the sentence casing, I would say let's move forward for now but I'll circle back with some folks and we can change if needed later on

libs/ui-lib/lib/ocm/components/Routes.tsx
jgyselov: I am not sure about that. Let's leave it as-is both here and in UHC-Portal.

I can discuss this with @ammont82 when she's back from PTO.

PR #2661, opened 2024-09-10
MGMT-18862: production via console.dev.redhat.com returns error

jgyselov:
https://issues.redhat.com/browse/MGMT-18862

UHC Portal: https://gitlab.cee.redhat.com/service/uhc-portal/-/merge_requests/5931

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterDetail/AssistedInstallerDetailCard.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterDetail/AssistedInstallerDetailCard.tsx
@@ -29,11 +29,17 @@ import { usePullSecret } from '../../hooks';
 import { Cluster } from '@openshift-assisted/types/assisted-installer-service';
 import { useFeatureDetection } from '../../hooks/use-feature-detection';
 import { OpenshiftVersionsContextProvider } from '../clusterWizard/OpenshiftVersionsContext';
-import { BrowserRouter } from 'react-router-dom-v5-compat';
+import {
+  unstable_HistoryRouter as HistoryRouter,
+  HistoryRouterProps,
+  BrowserRouter,
+} from 'react-router-dom-v5-compat';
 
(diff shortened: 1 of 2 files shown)


libs/ui-lib/lib/ocm/components/clusterDetail/AssistedInstallerDetailCard.tsx
```diff
@@ -29,11 +29,17 @@ import { usePullSecret } from '../../hooks';
 import { Cluster } from '@openshift-assisted/types/assisted-installer-service';
 import { useFeatureDetection } from '../../hooks/use-feature-detection';
 import { OpenshiftVersionsContextProvider } from '../clusterWizard/OpenshiftVersionsContext';
-import { BrowserRouter } from 'react-router-dom-v5-compat';
+import {
+  unstable_HistoryRouter as HistoryRouter,
```
celdrake, line 33: I wouldn't alias `unstable_HistoryRouter`, to signal avoiding using it when possible.

celdrake (approved):

libs/ui-lib/lib/ocm/components/clusterDetail/AssistedInstallerDetailCard.tsx
celdrake: Sorry, I hadn't checked the UHC portal PR before.

Wouldn't this approach go against the decoupling between the two applications?
Could those changes in UHC Portal be removed when we continue working with the decoupling?

jgyselov: It looks like it needs to be aliased for React to recognise it as a component.

I don't know what the plans are with regards to routing in the decoupling effort. But this is not the only place in the app where it will need to be addressed. But for now, let's just focus on making it not crash.

rawagner (approved):

PR #2668, opened 2024-09-24
ACM-13728: Missing s390x arch option in Cpu architecture in InfraEnv

jgyselov:
https://issues.redhat.com/browse/ACM-13728

Console PR: https://github.com/stolostron/console/pull/3894


libs/ui-lib/lib/cim/components/common/CpuArchitectureDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [cpuArchOpen, setCpuArchOpen] = React.useState(false);
+  const fieldId = getFieldId(name, 'input');
+
+  const onCpuArchToggle = React.useCallback(() => {
+    if (cpuArchOpen) {
+      setCpuArchOpen(false);
+    } else {
+      setCpuArchOpen(true);
+    }
+  }, [cpuArchOpen]);
```
rawagner, line 27: There is no need to memoize this function. It could also be simplified to `setCpuArchOpen(!cpuArchOpen)`

libs/ui-lib/lib/cim/components/common/CpuArchitectureDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+      if (val === CpuArchitecture.s390x) {
+        setFieldValue('userManagedNetworking', true);
+      } else {
+        setFieldValue('userManagedNetworking', false);
+      }
+      setCpuArchOpen(false);
+    },
+    [setValue, setFieldValue],
+  );
+
```
rawagner, line 43: Again, no need to memoize. 

There is also a bug in formik, where `setFieldValue` is always a new instance - we have a workaround for that in [useFormikHelpers.ts](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/common/hooks/useFormikHelpers.ts)

We can also simplify `setFieldValue('userManagedNetworking', val === CpuArchitecture.s390x);`

rawagner (changes requested):

ammont82 (approved):

rawagner (approved):

PR #2716, opened 2024-11-27
NO-ISSUE: Add OpenShift sandboxed containers operator

jhernand:
This is how the UI looks like:

image

### DIFF
--- a/libs/ui-lib/lib/common/config/docs_links.ts
+++ b/libs/ui-lib/lib/common/config/docs_links.ts
@@ -70,6 +70,9 @@ export const ODF_REQUIREMENTS_LINK =
 export const OPENSHIFT_AI_REQUIREMENTS_LINK =
   'https://docs.redhat.com/en/documentation/red_hat_openshift_ai_cloud_service/1/html/installing_the_openshift_ai_cloud_service/requirements-for-openshift-ai_install#requirements-for-openshift-ai_install';
 
+export const OSC_REQUIREMENTS_LINK =
+  'https://docs.redhat.com/en/documentation/openshift_sandboxed_containers/1.8/html/user_guide/deploying-osc-bare-metal#osc-resource-requirements_deploying-bare-metal';
+
 export const CNV_LINK = 'https://cloud.redhat.com/learn/topics/virtualization/';
 
 export const ODF_LINK = 'https://www.redhat.com/en/resources/openshift-data-foundation-datasheet';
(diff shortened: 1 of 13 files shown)


libs/ui-lib/lib/common/config/docs_links.ts
```diff
@@ -98,3 +101,6 @@ export const getCiscoIntersightLink = (downloadIsoUrl: string) =>
 
 export const MTV_LINK =
   'https://docs.redhat.com/en/documentation/migration_toolkit_for_virtualization/2.7/';
+
+export const OSC_LINK =
+  'https://www.redhat.com/en/blog/exploring-openshift-confidential-containers-solution';
```
pezhang, line 106: @jhernand  Maybe it's better to use link https://docs.redhat.com/en/documentation/openshift_sandboxed_containers which is the OSC products documentation page. Thanks.

libs/ui-lib/lib/common/config/docs_links.ts
```diff
@@ -67,6 +67,9 @@ export const HOST_REQUIREMENTS_LINK = 'https://access.redhat.com/solutions/48856
 export const ODF_REQUIREMENTS_LINK =
   'https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation';
 
+export const OSC_REQUIREMENTS_LINK =
+  'https://www.redhat.com/en/blog/red-hat-openshift-confidential-containers-deployment-solution';
+
```
pezhang, line 72: @jhernand How about using the https://docs.redhat.com/en/documentation/openshift_sandboxed_containers/1.8/html/user_guide/deploying-osc-bare-metal#osc-resource-requirements_deploying-bare-metal link here? Thanks.

jhernand: Sure, this was just an example. I will do that change and rebase.

libs/ui-lib/lib/common/config/docs_links.ts
jhernand: Done.

libs/ui-lib/lib/common/config/docs_links.ts
jhernand: Done.

pezhang:
Hello @jhernand , I have copied this PR to https://github.com/openshift-assisted/assisted-installer-ui/pull/2719.  This PR can be closed. Thanks for drafting it.

PR #2718, opened 2024-12-02
Mgmt-18987 allow iso types

liatb-rh:
Adding the ability to add minimal or full iso as host
Screenshot 2024-12-02 at 14 41 46


gamli75:
@nirfarkas @liatb-rh  should we have the same description as we have in the SaaS?
image

PR #2779, opened 2025-02-12
NO-ISSUE: Show GPUs

jhernand:
This patch adds a _GPUs_ section to the host inventory that shows the list of GPUs detected. For example, in a virtual machine with an AMD GPU and a VirtIO GPU:

gpus


nirfarkas:
Super nice!

PR #2824, opened 2025-03-03
MGMT-19323: Static IP - changing network IP range after filling host data - UI not showing the error

ElayAharoni:
https://issues.redhat.com/browse/MGMT-19323

resolved by changing the formik touched fields to include the disscussed fields, in order to show the errors correctly.
this fix also fixed the issue that on first mount the required error was not shown also.

behaviour after the fix:
https://www.loom.com/share/5e860418bb944d26b4b9453d3714a7fe?sid=eec31f75-430a-431e-8807-75c3e6b7f079


rawagner (changes requested):
Great job on finding a solution. It took me a while to get through this Static IP madness. It really is quite over-engineered form that needs some good re-thinking.

I'd suggest an easier, less error-prone, workaround.
Scrap the change that you did in `FormViewHostsFields.tsx`.
And change the component `AutosaveWithParentUpdate`

```ts
const { isSubmitting, isValid, touched, errors, values, initialValues, submitForm } =
    useFormikContext<StaticIpFormValues>();
  React.useEffect(() => {
    if (!isEqual(emptyValues, initialValues)) {
      void submitForm();
    }
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

This ensures that every time StaticIPForm is mounted and init values are not empty ones (IOW we are editing), we submit the form which forces `formik` to set all fields to touched.

rawagner (approved):

PR #2842, opened 2025-03-06
AGENT-1132: Add download kubeconfig wizard page

ElayAharoni:
https://issues.redhat.com/browse/AGENT-1132

added the new Download Kubeconfig page to the wizard only on disconnected assisted installer.
(here it shows it after the Operators step, it is only for testing reasons. the real behavior it will be after Networking and right before Review)

image


libs/ui-lib/lib/ocm/components/clusterWizard/wizardTransition.ts
```diff
@@ -245,6 +247,19 @@ const reviewStepValidationsMap: WizardStepValidationMap = {
   softValidationIds: [],
 };
 
+const kubeconfigValidationMap: WizardStepValidationMap = {
```
rawagner, line 250: I dont think we need any validations on this step
```suggestion
const kubeconfigValidationMap = buildEmptyValidationsMap();
```

libs/ui-lib/lib/ocm/components/clusterWizard/ClusterWizardContextProvider.tsx
```diff
@@ -71,6 +71,7 @@ const getWizardStepIds = (
   if (isSingleClusterFeatureEnabled) {
     // tentatively removed, proper waiting on support by backend
     stepsCopy = removeStepFromClusterWizard(stepsCopy, 'operators', 1);
```
rawagner, line 73: lets remove these so we keep the operators page (at least for now)
```suggestion
```

rawagner (changes requested):

PR #2858, opened 2025-03-11
MGMT-20083: validation blocks Next button but leave Add Hosts enabled

ElayAharoni:
https://issues.redhat.com/browse/MGMT-20083

fixed it by diasabling the add hosts and the next button whenever the is a sync error like in this case.
screenshot of the behavior after the fix: 
image


rawagner (commented):
I dont think we can fix this issue this way. The sync error could be due to some changes in networking page, which would mean if we click back, we get stuck on this one, unable to access networking to fix it.

I think we just need to add a proper support for determining version & CP count so that the sync error wont happen.

jgyselov:
> I think we just need to add a proper support for determining version & CP count so that the sync error wont happen.

That is a work in progress.

jgyselov:
Fixed in https://github.com/openshift-assisted/assisted-installer-ui/pull/2861

PR #2863, opened 2025-03-17
MGMT-20082: Deploying 4/5 nodes control plane is blocked

jgyselov:
https://issues.redhat.com/browse/MGMT-20082
https://issues.redhat.com/browse/MGMT-20083

BP of https://github.com/openshift-assisted/assisted-installer-ui/pull/2861


jgyselov:
**Control Plane Nodes select**

4 and 5 nodes are only available for

- images >=`18`
- non `-multi` images

image
image
image

**OpenShift version select**

(should be the same as in OCM)

image
image
image
image

@nirfarkas

PR #2870, opened 2025-03-20
AGENT-1142: Disable Nutanix platform integration

ElayAharoni:
https://issues.redhat.com/browse/AGENT-1142

disabled Nutanix integration for disconnected env

image


rawagner (commented):
On second thought, we should remove the option completely, not just disable.
As in disconnected, there is no configuration that user could do to make it enabled.

ElayAharoni:
after new fix: 
image

libs/ui-lib/lib/ocm/components/clusterConfiguration/platformIntegration/ExternalPlatformDropdown.tsx
```diff
@@ -142,6 +154,7 @@ export const ExternalPlatformDropdown = ({
       newFeatureSupportLevelContext,
       featureSupportLevelData,
       cpuArchitecture,
+      isSingleClusterFeatureEnabled,
```
rawagner, line 157: this should be added as a dependency to `useEffect`
I see that the eslint rule for checking the dependency array is disabled. I dont see a reason to do so, so please remove it and fix the deps too. Thanks!

rawagner (approved):

PR #2880, opened 2025-03-25
MGMT-20182: Day2 - host not shown in the UI session that generated it, but is shown in another session

jgyselov:
https://issues.redhat.com/browse/MGMT-20182

### DIFF
--- a/libs/ui-lib/lib/ocm/components/HostsClusterDetailTab/HostsClusterDetailTabContent.tsx
+++ b/libs/ui-lib/lib/ocm/components/HostsClusterDetailTab/HostsClusterDetailTabContent.tsx
@@ -33,6 +33,9 @@ export const HostsClusterDetailTabContent = ({
   const [day2Cluster, setDay2Cluster] = useStateSafely<Cluster | null>(null);
   const pullSecret = usePullSecret();
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const ocmClusterMemo = React.useMemo(() => ocmCluster, []);
+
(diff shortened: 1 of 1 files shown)


libs/ui-lib/lib/ocm/components/HostsClusterDetailTab/HostsClusterDetailTabContent.tsx
```diff
@@ -33,6 +33,9 @@ export const HostsClusterDetailTabContent = ({
   const [day2Cluster, setDay2Cluster] = useStateSafely<Cluster | null>(null);
   const pullSecret = usePullSecret();
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const ocmClusterMemo = React.useMemo(() => ocmCluster, []);
```
rawagner, line 37: is this really ok ? Can you please explain why @jgyselov @ammont82

jgyselov: The effect which creates/fetches the day2 cluster triggers twice without `ocmCluster` being memoised. That results in two day2 clusters (which is a nightmare).

The `ocmCluster` doesn't change at all between these two triggers and the information that we need from the `ocmCluster` (`external_id`, CPU architecture, OpenShift version etc.) will not change at all during runtime. The only value in the effect dependency array that changes is `isVisible` and `day2Cluster`.

It is a bit iffy, I do agree, but it should not cause us trouble.

PR #2882, opened 2025-03-27
MGMT-20217: Assisted Operators Review

ElayAharoni:
https://issues.redhat.com/browse/MGMT-20217

image

image

image

image


libs/ui-lib/lib/ocm/components/clusterConfiguration/operators/MceCheckbox.tsx
```diff
@@ -23,7 +23,7 @@ const MceLabel = ({
   return (
     <>
       <Tooltip hidden={!disabledReason} content={disabledReason}>
-        <span>Install multicluster engine </span>
+        <span>multicluster engine </span>
```
jgyselov, line 26:
```suggestion
        <span>Multicluster engine </span>
```

In general, we should be using sentence case. See https://www.patternfly.org/ux-writing/capitalization

ammont82 (approved):

PR #3004, opened 2025-06-19
MGMT-19504: Clean-up code related to highAvailabilityMode

celdrake:
Clean-up code related to `highAvailabilityMode` which is being deprecated in favor of the already used `controlPlaneCount`.

Removed from:
- Cypress tests
- OCM code

Kept at:
- SDK definition files, they are quite outdated
- CIM

I also removed `SNOControlGroup` and `SNODisclaimer` since these were not being used anymore.

Also, in  `OcmSNODisclaimer` there was a checkbox that needed to be checked when the user chose an OCP version with unsupported/dev-preview support for SNO. Currently, all the versions in the UI (4.9 to 4.20) have SNO feature as `supported`, so that was also dead code.

## Summary by CodeRabbit

- **Removed Features**
  - Removed all user interface elements and controls related to Single Node OpenShift (SNO) configuration, including checkboxes, disclaimers, and related form fields.
- **Changes**
  - Updated logic to determine Single Node OpenShift status based on the number of control plane nodes instead of high availability mode.
  - Simplified alerts and informational messages for clusters with a single control plane node.
  - Removed Single Node OpenShift related translation strings and user acknowledgments.
- **Bug Fixes**
  - Improved consistency in detecting and displaying Single Node OpenShift clusters across the application.


libs/ui-lib/lib/common/components/clusterWizard/clusterDetailsValidation.ts
```diff
@@ -123,28 +120,6 @@ export const getClusterDetailsValidationSchema = ({
         ? baseDomainValidationSchema.required('Required')
         : dnsNameValidationSchema.required('Required'),
       pullSecret: pullSecretValidationSchema.required('Required.'),
-      SNODisclaimer: Yup.boolean().when(['controlPlaneCount', 'openshiftVersion'], {
```
celdrake, line 126: Also realised that there's no SNO disclaimer checkbox in CIM, so no need for keeping this code.

PR #3015, opened 2025-06-24
MGMT-20895: Use humanized label for displaying the selected subnet

celdrake:
The  `currentDisplayValue` in  `SubnetsDropdown` was kept in an unnecessary `useState`, since it's value can be derived from other fields. On top of that, when the component first mounted, `currentDisplayValue` had the humanized label (eg. `192.168.122.0/24 (192.168.122.0 - 192.168.122.255)`), but when the user modified the selection, the display value had only the field value (eg. `192.168.122.0/24`).

Now `currentDisplayValue` is always displaying the humanized label, and it's calculated from the fields it derives from.

## Summary by CodeRabbit

- **Tests**
  - Enhanced networking UI tests to verify correct display of IPv4 and IPv6 subnet CIDR ranges during single and dual-stack configurations.  
- **Refactor**
  - Streamlined subnet dropdown label logic for more efficient and consistent display behavior.


celdrake:
@ammont82 @jgyselov I verified the fix works well in the Cypress tests, but I couldn't test the changes for real in Dual Stack clusters. 

Could you help me with how to configure for Dual Stack?  Thanks.

PR #3039, opened 2025-07-11
MGMT-20303: Do not allow renaming host after install startedOC

celdrake:
Prevents being able to rename a Day2 host after it started installing.

Note: I couldn't test the part after the Day2 host started installing, as my host was not getting Ready for installation.
From the documentation, it seems like it should behave correctly.

### DIFF
--- a/libs/ui-lib/lib/cim/components/Agent/tableUtils.tsx
+++ b/libs/ui-lib/lib/cim/components/Agent/tableUtils.tsx
@@ -40,7 +40,7 @@ export const agentHostnameColumn = (
   agents: AgentK8sResource[],
   bareMetalHosts: BareMetalHostK8sResource[],
   onEditHostname?: HostsTableActions['onEditHost'],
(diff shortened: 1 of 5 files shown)


libs/ui-lib/lib/cim/components/Agent/tableUtils.tsx
```diff
@@ -40,7 +40,7 @@ export const agentHostnameColumn = (
   agents: AgentK8sResource[],
   bareMetalHosts: BareMetalHostK8sResource[],
   onEditHostname?: HostsTableActions['onEditHost'],
-  canEditHostname?: HostsTableActions['canEditHost'],
+  canEditHostname?: HostsTableActions['canEditHostname'],
```
celdrake, line 43: The type referenced in CIM was incorrect, but because the "actions" doesn't define a `canEditHostname` function, no errors were appearing anywhere.

ammont82 (approved):

PR #3047, opened 2025-07-15
MGMT-21056: Deployment of Standalone cluster in disconnected ipv6 environment failing with error "Openshift version 4.19.3 is not supported for OpenShiftSDN" - OVN is not supported for ipv6

jgyselov:
https://issues.redhat.com/browse/MGMT-21056

If no OpenShift version has the label `visible` set to `"true"`, the `openshiftVersion` value is initially empty which results in `OpenShiftSDN` being used as the `networkType`. 

When we select a fallback OpenShift version from those with `visible` set to `"false"`, we need to reset `networkType` to a compatible value.

## Summary by CodeRabbit

* **Refactor**
  * Improved clarity and organization of the cluster details form fields, including more logical ordering of hooks and variables.
  * Enhanced naming for disk encryption indicators for easier understanding.
  * Internal updates to how network type is set when using a default OpenShift version.


celdrake (approved):
Worked well.

PR #3048, opened 2025-07-16
MGMT-21080 Enable all ocp versions for ACM 2.12

celdrake:
Takes most of the code from https://github.com/openshift-assisted/assisted-installer-ui/pull/2857 and adapts it to ACM 2.12 with PF4.

Tested the following environments:
- Create a new cluster, local-cluster has all versions having visible=undefined/false
- Create a new cluster, local-cluster has some visible versions
- Check "Cluster details" for a saved cluster Draft, the version that had been selected was a non-visible version (the version is in read-only mode)

https://github.com/user-attachments/assets/ec823e2d-8e04-405e-a3d9-7beea664d15a

### DIFF
--- a/libs/ui-lib-tests/cypress/views/forms/ClusterDetails/Fields/OpenShiftVersionField.ts
+++ b/libs/ui-lib-tests/cypress/views/forms/ClusterDetails/Fields/OpenShiftVersionField.ts
@@ -20,8 +20,11 @@ export class OpenShiftVersionField {
 
   static selectVersion(version: string) {
     OpenShiftVersionField.findDropdown().click();
-    OpenShiftVersionField.findDropdown().within(() => {
-      cy.findByRole('menuitem', { name: new RegExp(`openshift ${version}`, 'i') }).click();
-    });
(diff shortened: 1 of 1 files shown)


libs/ui-lib-tests/cypress/views/forms/ClusterDetails/Fields/OpenShiftVersionField.ts
```diff
@@ -20,8 +20,11 @@ export class OpenShiftVersionField {
 
   static selectVersion(version: string) {
     OpenShiftVersionField.findDropdown().click();
-    OpenShiftVersionField.findDropdown().within(() => {
-      cy.findByRole('menuitem', { name: new RegExp(`openshift ${version}`, 'i') }).click();
-    });
+    // Wait for dropdown to be expanded and items to be visible
```
celdrake, line 23: Seems like we have two ways to get the versions.
I adapted this one using Cursor, as the component is using PF4 and the structure is pretty different.

PR #3049, opened 2025-07-16
MGMT-21014: Impossible to choose n+1 or n-4 version - only n, n+1 and n+2 are shown in OCP version dropdown

jgyselov:
https://issues.redhat.com/browse/MGMT-21014

Let's add the OpenShift version dropdown with the "All versions" modal to the Hypershift details form for consistency.

## Summary by CodeRabbit

* **New Features**
  * Enhanced OpenShift version selection with a new dropdown supporting additional options and a modal for extended version details.
  * Support for custom OpenShift versions not present in the default list.

* **Refactor**
  * Updated form structure to use the new version dropdown and modal, replacing the previous version selection component.

* **Chores**
  * Removed deprecated OpenShift version selection components and related exports.


celdrake (approved):
The dropdown behaves as expected.

However from the bug description it appears that there may be some issue whereby "visible=true" versions don't appear in the dropdown.

The current fix would allow those versions to be selected, but they wouldn't appear on the base dropdown, only in the extended one.

PR #3065, opened 2025-07-22
OCPBUGS-59345: Use featureSet CustomNoUpgrade

bfournie:
Setting the featureSet to TechPreviewNoUpgrade enabled all TechPreview features which caused the installation to fail (see details in bug). Instead CustomNoUpgrade is used with a featureGate specific to OVE UI.

## Summary by CodeRabbit

* **New Features**
  * Updated network configuration to support custom feature sets and additional feature gates for cluster operations.


ammont82:
@bfournie we have some format issues. Can you solve it?

bfournie:
> @bfournie we have some format issues. Can you solve it?

Yes, @ammont82 I ran `prettier` against the file, it should be all set now.

PR #3073, opened 2025-07-24
MGMT-21259: Do not animate scroll when chat is reopened

celdrake:
When users open an existing chat, do not animate scrolling to the bottom.
We still scroll to the bottom in case they want to continue the conversation.

https://github.com/user-attachments/assets/c54f1ec9-bdab-4de2-83aa-808d692d301f

## Summary by CodeRabbit

* **Enhancements**
  * Improved chat window scrolling for a smoother user experience: the window now scrolls instantly to the latest message on initial load and smoothly for new incoming messages.


celdrake:
@ammont82 I'd like to update this PR with the changes from https://github.com/openshift-assisted/assisted-installer-ui/pull/3074/files so that when users reopen the chatbot, we also focus the "Send message" input so they can type new questions if needed.

So please review #3074 first and I'll rebase and update this one later.

PR #3122, opened 2025-08-18

asmasarw:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prefix Length in the Static IP (Network-Wide) form is now required and must be numeric; null/empty values are rejected.

* **New Features**
  * Added a disconnected cluster wizard flow with two new steps: Basic information and Review & Download ISO (includes ISO download and summary).
  * Cluster wizard gains a toggle to switch between connected and disconnected flows.
  * Wizard now renders steps dynamically; Cluster Details form obtains navigation from context.
  * OpenShift version display supports optional supplementary content.

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/FormViewNetworkWide/FormViewNetworkWideFields.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/FormViewNetworkWide/FormViewNetworkWideFields.tsx
@@ -76,20 +76,6 @@ const MachineNetwork: React.FC<{ fieldName: string; protocolVersion: ProtocolVer
     const cidr = getMachineNetworkCidr(value);
     return getHumanizedSubnetRange(getAddressObject(cidr, protocolVersion));
   }, [value, protocolVersion, errorMessage]);
-
-  const handleValidatePrefixLength = (value: string) => {
-    if (!value || value === '') {
-      return 'Prefix length is required';
-    }
-    if (Number(value) < MIN_PREFIX_LENGTH) {
-      return `Prefix length must be at least ${MIN_PREFIX_LENGTH}`;
-    }
-    if (Number(value) > MAX_PREFIX_LENGTH[protocolVersion]) {
-      return `Prefix length must be at most ${MAX_PREFIX_LENGTH[protocolVersion]}`;
-    }
-    return undefined;
-  }
-
   const fieldId = getFieldId(`${fieldName}`, 'input');
   return (
     <FormGroup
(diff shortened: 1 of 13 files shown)


libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/FormViewNetworkWide/FormViewNetworkWideFields.tsx
```diff
@@ -76,6 +76,20 @@ const MachineNetwork: React.FC<{ fieldName: string; protocolVersion: ProtocolVer
     const cidr = getMachineNetworkCidr(value);
     return getHumanizedSubnetRange(getAddressObject(cidr, protocolVersion));
   }, [value, protocolVersion, errorMessage]);
+
+  const handleValidatePrefixLength = (value: string) => {
```
jgyselov, line 80: While this technically does the job, I don't think it's the best solution here.

In the Assisted Installer UI project, we use [Yup](https://github.com/jquense/yup) and [Formik](https://formik.org/docs/overview) for form handling and validation. And you can generally assume that if there's a form, there is a Yup validation schema for it somewhere.

In the case of the static IP form (more specifically, the "form view", "network wide" static IP form), the Yup schema is defined in [here](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/FormViewNetworkWide/formViewNetworkWideValidationSchema.tsx#L99-L124) and attached to the form via [here](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/FormViewNetworkWide/FormViewNetworkWide.tsx#L25).

From the `networkWideValidationSchema` object, you can navigate to [getMachineNetworkValidationSchema](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/FormViewNetworkWide/formViewNetworkWideValidationSchema.tsx#L59-L71) where we already have a validation schema set up for the subnet prefix length.

If you play around with the prefix length input field on the master branch, you'll see that the validation for the minimum and maximum is working. So the question is, why does it not work for the `required` test?

My immediate theory is that `required` doesn't handle the `prefixLength: number | ''` type (defined in [here](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/data/dataTypes.ts#L15-L18)) the way we wanted it to, so `required` might not be the best way to go here.

You will need to tweak the validation schema so that it handles the `''` value the way we want it to. You should be able to do it all with Yup functions. Look at [.when()](https://github.com/jquense/yup?tab=readme-ov-file#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema) and [.test()](https://github.com/jquense/yup?tab=readme-ov-file#schematestname-string-message-string--function--any-test-function-schema) and see if they are helpful.

jgyselov (changes requested):
Please, also rename the PR to follow the `MGMT-xxxx: [jira title]` format. It makes navigation between PRs and Jira tickets easier and allows the @openshift-ci-robot to trigger a procedure that links this PR in the corresponding Jira ticket (so you don't have to do that manually -- it's nice!).

PR #3125, opened 2025-08-19

asmasarw:
[MGMT-20743](https://issues.redhat.com//browse/MGMT-20743) | [Staging] [UI] - Machine network CIDR value should be mandatory


asmasarw:

PR #3126, opened 2025-08-19

asmasarw:


jgyselov (approved):
Nice! The solution was simpler than I expected 😆

PR #3143, opened 2025-08-27
Remove operators from ABI that are not yet included

rawagner:
## Summary by CodeRabbit

- Changes
  - Removed Cluster Observability, NUMA Resources, OADP, and MetalLB from the Single-Cluster Operators list in the UI.
  - These operators will no longer appear in operator selection, filtering, or related listings for single-cluster setups.
  - Any UI references to these operators in single-cluster contexts have been removed to reflect the updated availability.


celdrake:
@rawagner Could we mark this as a revert of https://github.com/openshift-assisted/assisted-installer-ui/pull/3101?
And to update the JIra to reflect that it's being reverted.

PR #3150, opened 2025-09-01
MGMT-21647: Add delete conversation option in chat history

rawagner:


libs/chatbot/lib/components/ChatBot/ChatBotHistory.tsx
```diff
@@ -64,44 +74,92 @@ const ChatBotHistory = ({
...
+        isLoading={isLoading}
+        conversations={conversations.map<Conversation>((c) => ({
+          ...c,
+          additionalProps: {
+            actions: (
+              <MenuItemAction
+                icon={<TrashAltIcon />}
+                actionId="delete"
+                onClick={() => setDeleteConversation(c.id)}
+                aria-label={`Delete conversation ${c.id}`}
```
celdrake, line 98: Nit: Since it's meant for the user, would be better to show the date.

libs/chatbot/lib/components/ChatBot/DeleteConversationModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    >
+      <ModalHeader
+        titleIconVariant="danger"
+        title="Delete conversation?"
+        labelId="delete-conversation-modal"
+      />
+      <ModalBody id="modal-box-body-delete-conversation">
+        <Stack hasGutter>
+          <StackItem>
+            Are you sure you want to delete conversation <b>{conversationId}</b>?
```
celdrake, line 58: Again, I think for the user the ID does not have any meaning. 
They only get to see the date so that's what I would show here again instead.

PR #3155, opened 2025-09-03
AGENT-1261: Update dependency sourcing to remote

pawanpinjarkar:
Update dockerfile for ART to manage dependencies from REMOTE_SOURCES and CACHED_YARN for the ART builds

### DIFF
--- a/apps/assisted-disconnected-ui/Containerfile.ocp
+++ b/apps/assisted-disconnected-ui/Containerfile.ocp
@@ -1,21 +1,48 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-base-nodejs-openshift-4.19 AS ui-build
-USER root
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-base-nodejs-openshift-4.20 AS ui-build
+USER 0
 
 WORKDIR /app
-COPY --chown=1001:0 / /app
-RUN ls /app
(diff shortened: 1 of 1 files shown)


apps/assisted-disconnected-ui/Containerfile.ocp
```diff
@@ -1,23 +1,43 @@
...
+# The REMOTE_SOURCES value is set by the build system to indicate the location of the cachito-backed artifacts cache.
+COPY $REMOTE_SOURCES $REMOTE_SOURCES_DIR
+
+ARG YARN_VERSION=3.4.1
+
+# bootstrap yarn so we can install and run the other tools.
+RUN CACHED_YARN=./artifacts/${YARN_VERSION}.tar.gz; \
+    if [ -f ${CACHED_YARN} ]; then \
+      # use dependencies provided by Cachito
+      npm install -g ${CACHED_YARN}; \
```
pawanpinjarkar, line 18: @rawagner Is this correct and sufficient to install yarn 3.4.1 or it still needs corepack?

rawagner: installing only yarn 3.4.1 should be enough

pawanpinjarkar:
@rawagner can you please cherrypick this to 4.20. Seems like I can't have enough access.

PR #3188, opened 2025-10-02
OCPBUGS-62680: Include assisted disconnected UI image in release payload

pawanpinjarkar:
To onboard assisted disconnected UI onto OCP, add the `LABEL io.openshift.release.operator=true` so that the image is included in the release payload.


rwsu (commented):
/lgtm

rawagner (approved):

PR #3374, opened 2026-02-01
OCPBUGS-72598 | [OVE] Custom manifests are broken in local Assisted UI

asmasarw:
## Summary by CodeRabbit

* **Bug Fixes**
  * Refined cluster wizard step navigation to prevent forced advancement away from the custom manifests step while it is actively being edited. This improves the user experience when configuring cluster settings.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterWizard/ClusterWizardContextProvider.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterWizard/ClusterWizardContextProvider.tsx
@@ -160,10 +160,14 @@ const ClusterWizardContextProvider = ({
         isSingleClusterFeatureEnabled,
       );
 
-      // Only move step if there is still none, or the user is at a forbidden step
+      // Only move step if there is still none, or the user is at a forbidden step.
+      // Never force the user off the custom-manifests step when they are actively filling it
+      // (e.g. after pasting content triggers auto-save and uiSettings can update).
       if (
         !currentStepId ||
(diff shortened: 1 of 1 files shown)


libs/ui-lib/lib/ocm/components/clusterWizard/ClusterWizardContextProvider.tsx
```diff
@@ -160,10 +160,14 @@ const ClusterWizardContextProvider = ({
...
-      // Only move step if there is still none, or the user is at a forbidden step
+      // Only move step if there is still none, or the user is at a forbidden step.
+      // Never force the user off the custom-manifests step when they are actively filling it
+      // (e.g. after pasting content triggers auto-save and uiSettings can update).
       if (
         !currentStepId ||
-        (customManifestsStepNeedsToBeFilled && isStepAfter(currentStepId, requiredStepId))
+        (currentStepId !== 'custom-manifests' &&
+          customManifestsStepNeedsToBeFilled &&
+          isStepAfter(currentStepId, requiredStepId))
```
rawagner: I dont understand the fix. This useEffect is triggered when user opens the wizard and settings are loaded/or settings are changed. How can this have some effect on copying content into custom manifests field ? the settings should be loaded already, right ?

That leaves out that uiSettings changed and maybe we dont calculate `customManifestsStepNeedsToBeFilled` properly ?

Can you explain what is the root cause that triggers the original condition which changes the current step?

PR #3523, opened 2026-03-31
MGMT-23739: Cisco Intersight URL cannot be added after the AgentServiceConfig has been created

jgyselov:
https://redhat.atlassian.net/browse/MGMT-23739

## Summary by CodeRabbit

* **New Features**
  * More reliable CIM enable/edit flow with clearer error reporting during save operations.

* **Bug Fixes**
  * Improved handling of configuration metadata and storage sizing so settings apply correctly for new and existing setups.
  * More robust ingress/load balancer configuration and domain derivation with explicit failure reporting.

* **Chores**
  * Removed an unused localization entry for a route-query error message.

image

### DIFF
--- a/libs/ui-lib/lib/cim/components/modals/CimConfiguration/CimConfigurationModal.tsx
+++ b/libs/ui-lib/lib/cim/components/modals/CimConfiguration/CimConfigurationModal.tsx
@@ -20,12 +20,18 @@ import parseUrl from 'parse-url';
 import { useTranslation } from '../../../../common/hooks/use-translation-wrapper';
 import { getStorageSizeGiB } from '../../helpers';
 import { AlertPayload } from '../../../../common';
+import { getErrorMessage } from '../../../../common/utils';
 
 import { CimConfigurationModalProps, CimConfigurationValues } from './types';
-import { onEnableCIM } from './persist';
+import {
+  patchProvisioning,
+  createAgentServiceConfig,
+  editAgentServiceConfig,
+  configureIngressLoadBalancer,
+} from './persist';
 import { CimConfigDisconnectedAlert } from './CimConfigDisconnectedAlert';
 import { MIN_DB_VOL_SIZE, MIN_FS_VOL_SIZE, MIN_IMG_VOL_SIZE } from './constants';
-import { isCIMConfigProgressing } from './utils';
+import { isCIMConfigProgressing, LOCAL_STORAGE_ID_LAST_UPDATE_TIMESTAMP } from './utils';
 import { resetCimConfigProgressAlertSuccessStatus } from './CimConfigProgressAlert';
 import { CimConfigurationFormFields } from './CimConfigurationFormFields';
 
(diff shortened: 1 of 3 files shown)


libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
```diff
@@ -264,133 +194,98 @@ const createAgentServiceConfig = async ({
...
+export const configureIngressLoadBalancer = async ({ t }: { t: TFunction }) => {
+  if (await isIngressController()) {
+    return;
   }
 
-  return true;
+  const assistedImageServiceRoute = await getAssistedImageServiceRoute(t);
+  const domain = getClusterDomain(t, assistedImageServiceRoute);
+  await createIngressController(t, domain);
+  await patchAssistedImageServiceRoute(t, assistedImageServiceRoute, domain);
```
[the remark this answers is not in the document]
jgyselov: This is consistent with previous behavior.

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
```diff
@@ -4,106 +4,72 @@ import {
...
-  // the appsDomain can be prefixed either by "apps." or "nlb-apps." so search for first dot
   const domain = appsDomain.substring(appsDomain.indexOf('.') + 1);
 
   if (!domain) {
-    // It must be present
-    // console.error('Can not find domain in assistedImageServiceRoute: ', assistedImageServiceRoute);
-    setError({
-      title: t('ai:Can not find cluster domain.'),
-    });
+    throw new Error(t('ai:Can not find cluster domain.'));
```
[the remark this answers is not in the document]
jgyselov: This is consistent with previous behavior.

libs/ui-lib/lib/cim/components/modals/CimConfiguration/CimConfigurationModal.tsx
```diff
@@ -52,24 +58,44 @@ export const CimConfigurationModal: React.FC<CimConfigurationModalProps> = ({
...
+        });
+      } else {
+        await createAgentServiceConfig({ t, values });
+      }
+
+      window.localStorage.setItem(LOCAL_STORAGE_ID_LAST_UPDATE_TIMESTAMP, new Date().toISOString());
+
+      if (values.configureLoadBalancer) {
+        await configureIngressLoadBalancer({ t });
+      }
```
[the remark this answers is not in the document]
jgyselov: This is consistent with previous behavior.

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
```diff
@@ -186,71 +142,45 @@ const createIngressController = async (
...
-      message: getErrorMessage(e),
-      variant: AlertVariant.warning,
-    });
+  if (provisioning.spec?.watchAllNamespaces) {
+    return;
   }
+
+  const patches = [
+    {
+      op: provisioning.spec?.watchAllNamespaces ? 'replace' : 'add',
```
LiorSoffer: it will always be 'add' because of the early return on lines 158-159

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
```diff
@@ -186,71 +142,45 @@ const createIngressController = async (
...
+      path: '/spec/watchAllNamespaces',
+      value: true,
+    },
+  ];
+
+  await k8sPatch({
```
LiorSoffer, line 170: we can wrap with try and catch

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
```diff
@@ -4,106 +4,72 @@ import {
...
   return assistedImageServiceRoute;
 };
 
-// There are many ways how to get the domain name, let's follow the documentation.
-const getClusterDomain = (
-  t: TFunction,
-  setError: SetErrorFuncType,
-  assistedImageServiceRoute: RouteK8sResource,
-): string | undefined => {
+const getClusterDomain = (t: TFunction, assistedImageServiceRoute: RouteK8sResource) => {
```
LiorSoffer, line 48:
```suggestion
const getClusterDomain = (t: TFunction, assistedImageServiceRoute: RouteK8sResource): string => {
```

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
jgyselov: We don't know if `provisioning.spec?.watchAllNamespaces` is `false` or missing entirely.

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
jgyselov: It's wrapped by try...catch in CimConfigurationModal.

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
jgyselov: We don't need an explicit return type.

libs/ui-lib/lib/cim/components/modals/CimConfiguration/persist.ts
LiorSoffer: by the time you reach line 164, `provisioning.spec?.watchAllNamespaces` is guaranteed to be  `false`/`undefined`/`null`  (all falsy). 
So the ternary always resolves to 'add'

jgyselov:
```
      op: provisioning.spec?.watchAllNamespaces === false ? 'replace' : 'add',
```
Should do the trick.

LiorSoffer (approved):

PR #3644, opened 2026-04-26
OCPBUGS-84331: Remove 2 node arbiter option from the control plane dropdown option

ElayAharoni:
https://redhat.atlassian.net/browse/OCPBUGS-84331

## Summary by CodeRabbit

* **Bug Fixes**
  * Control plane nodes dropdown now respects a single-cluster feature toggle: when enabled, the two-node option is removed and tooltips explain disconnected-installation specifics.
  * Tooltip messages and option enabling now update immediately when the feature toggle changes, preventing stale or misleading guidance.

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterConfiguration/ControlPlaneNodesDropdown.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterConfiguration/ControlPlaneNodesDropdown.tsx
@@ -11,7 +11,7 @@ import {
 import { useField } from 'formik';
 import toNumber from 'lodash-es/toNumber';
 import { getFieldId, PopoverIcon, PreviewBadgePosition, TechnologyPreview } from '../../../common';
-
+import { useFeature } from '../../hooks/use-feature';
 import {
   NewFeatureSupportLevelMap,
   useNewFeatureSupportLevel,
@@ -45,6 +45,8 @@ interface ControlPlaneNodesOption {
   value: number;
   label: string;
   badge?: React.ReactNode;
+  /** When true, option is omitted from the list (e.g. product mode does not offer this topology). */
+  hidden?: boolean;
 }
 
 interface ControlPlaneNodesDropdownProps {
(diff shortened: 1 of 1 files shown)


libs/ui-lib/lib/ocm/components/clusterConfiguration/ControlPlaneNodesDropdown.tsx
```diff
@@ -104,7 +111,7 @@ const ControlPlaneNodesDropdown: React.FC<ControlPlaneNodesDropdownProps> = ({
     { value: 3, label: '3 (highly available cluster)' },
     { value: 4, label: '4 (highly available cluster+)' },
     { value: 5, label: '5 (highly available cluster++)' },
-  ];
+  ].filter(({ value }) => !(isSingleClusterFeatureEnabled && value === 2));
```
rawagner: we could instead do a bit more generic approach by adding optional `hidden` prop to `ControlPlaneNodesOption` and then

```ts
const options: ControlPlaneNodesOption[] = [
    { value: 1, label: '1 (Single Node OpenShift)' },
    {
      value: 2,
      label: '2 (Two-Nodes Arbiter)',
      badge:
        tnaSupport === 'tech-preview' ? (
          <TechnologyPreview position={PreviewBadgePosition.default} />
        ) : undefined,
      hidden: isSingleClusterFeatureEnabled,
    },
    { value: 3, label: '3 (highly available cluster)' },
    { value: 4, label: '4 (highly available cluster+)' },
    { value: 5, label: '5 (highly available cluster++)' },
  ].filter(({ hidden }) => !hidden);

```

rawagner: beside this additional filtering, i dont think we need to change any other code (like adding additional logic for disabledReason etc, since the TNA option will never show)

rawagner (approved):

PR #3646, opened 2026-04-26
OCPBUGS-84040: Hide internally generated custom manifests on the Cluster Review page

ElayAharoni:
https://redhat.atlassian.net/browse/OCPBUGS-84040

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom manifests section in the cluster review summary now displays only user-provided manifests, filtering out system-generated entries.
  * Section visibility is correctly gated by whether the user has added manifests.
  * Manifest entries now preserve their full metadata and content for accurate display and review.
  * UI settings now correctly detect and reflect presence of user-added manifests.

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterConfiguration/review/ReviewSummary.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterConfiguration/review/ReviewSummary.tsx
@@ -19,9 +19,16 @@ import { ReviewCustomManifestsTable } from './ReviewCustomManifestsTable';
 import PlatformIntegrationNote from '../platformIntegration/PlatformIntegrationNote';
 import useClusterCustomManifests from '../../../hooks/useClusterCustomManifests';
 import { Cluster } from '@openshift-assisted/types/assisted-installer-service';
+import { ListManifestsExtended } from '../manifestsConfiguration/data/dataTypes';
+
+const userProvidedManifests = (manifests: ListManifestsExtended | undefined) =>
+  (manifests ?? []).filter((m) => m.manifestSource !== 'system');
 
 export const ReviewSummaryContent = ({ cluster }: { cluster: Cluster }) => {
   const { customManifests } = useClusterCustomManifests(cluster.id, false);
+  const manifestsForReview = userProvidedManifests(customManifests);
+  const showCustomManifests = manifestsForReview.length > 0;
+
   return (
     <>
       <TableSummaryExpandable title={'Cluster details'} id={'cluster-details-expandable'}>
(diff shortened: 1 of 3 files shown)


libs/ui-lib/lib/ocm/hooks/useUISettings.ts
```diff
@@ -25,20 +23,8 @@ const useUISettings = (clusterId?: Cluster['id']) => {
   );
 
   React.useEffect(() => {
-    const handleUISettingsError = async () => {
```
rawagner, line 28: why was this error handling removed ?

ElayAharoni: customManifestsAdded: !!customManifests.length
That is exactly what causes the bug (system manifests counted as user custom manifests).
so  i removed it and on failure we now fall back to safe defaults

jgyselov: Empty UI settings are not a safe default. They are an inaccurate one. 🙂 

The `handleUISettingsError` function was constructed to fetch the custom manifest information in case fetching UI settings fails. Your version essentially tells the UI that there are **no** custom manifests if we cannot fetch the UI settings.

But that might not be the case—we cannot say whether there are custom manifests or not if we do not ask. The effect of this is frustrating on the Review page and even worse on the Custom manifests page.

Maybe we could just simply filter out the system manifests when we set `mockUISettings.customManifestsAdded`?

libs/ui-lib/lib/ocm/components/clusterConfiguration/review/ReviewSummary.tsx
```diff
@@ -19,9 +19,29 @@ import { ReviewCustomManifestsTable } from './ReviewCustomManifestsTable';
...
+
+export const ReviewSummaryContent = ({
+  cluster,
+  requireCustomManifestOptIn = false,
+  customManifestsAdded,
+}: ReviewSummaryContentProps) => {
   const { customManifests } = useClusterCustomManifests(cluster.id, false);
+  const manifestsForReview = userProvidedManifests(customManifests);
+  const showCustomManifests =
+    manifestsForReview.length > 0 && (!requireCustomManifestOptIn || customManifestsAdded === true);
```
jgyselov: The filtering of `manifestsForReview` is alright if we just want to show the ones the user added.

The rest of this condition is very confusing. If there are manifests the user added, they should be shown on the review page. At that point, looking at `customManifestsAdded` (from the UI settings) or another boolean prop (which I'll say I find incredibly confusing altogether) is imo redundant.

jgyselov (changes requested):

rawagner:
/approve

@jgyselov can you take a look too ?

jgyselov (approved):

PR #3699, opened 2026-05-07
fix iso name and instructions on above the sea review page

ElayAharoni:
https://redhat.atlassian.net/browse/OCPBUGS-83603


libs/ui-lib/lib/ocm/components/clusterWizard/disconnected/ReviewStep.tsx
```diff
@@ -35,6 +34,11 @@ const ReviewStep = () => {
   const opSpecs = getOperatorSpecs(() => undefined);
   const navigate = useNavigate();
 
+  const openshiftDocsVersion = disconnectedInfraEnv?.openshiftVersion
+    ? disconnectedInfraEnv.openshiftVersion.split('.').slice(0, 2).join('.')
+    : '';
```
rawagner: lets add some meaningful default instead of `''` which just creates an invalid doc link

maybe `4.22` ?

rawagner (approved):

PR #3704, opened 2026-05-07
OCPBUGS-85266: [release-4.21]Cherrypick 3687 release 4.21

ElayAharoni:
this is a manual cherry pick for https://github.com/openshift-assisted/assisted-installer-ui/pull/3687


rawagner:
I dont think the original PR (targeting master) fixes the issue and we actually cannot fix this on the UI side.

The backend does not provide any visibility into validation progress. We dont have an info if the validation is running or it is already finished AFAIK.

PR #3723, opened 2026-05-19
MGMT-24304: On disconnected environment no Openshift version is set by default on standalone cluster details page and no version is shown in the dropdown till Show all available are shown

jgyselov:
https://redhat.atlassian.net/browse/MGMT-24304

With OCP Console 4.22 the `customOpenshiftSelect` fallback wasn't getting set (likely to React 18 concurrency changes).

Needs:  https://github.com/stolostron/console/pull/6242

## Summary by CodeRabbit

* **Style**
  * Updated dropdown components to use full-width styling for improved consistency and better UI presentation.

* **Improvements**
  * Enhanced focus behavior in dropdown menus when selecting items for improved keyboard navigation experience.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openshift-assisted/assisted-installer-ui/pull/3723?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3731, opened 2026-05-19
OCPBUGS-86070: Fix disconnected flow version selection — replace hardcoded 4.20 with dynamic dropdown

asadawar:
## Summary

https://redhat.atlassian.net/browse/OCPBUGS-86070

The disconnected/air-gapped BasicStep had the OpenShift version hardcoded to `"4.20"` via a static `OcmOpenShiftVersion` component. This prevented users from selecting 4.21, 4.22 (including RC builds), or any future versions when the disconnected toggle is enabled.

**Root cause:** `OcmOpenShiftVersion openshiftVersion="4.20"` in `BasicStep.tsx` — a static display, not a selector.

**Fix:** Replace with `OcmOpenShiftVersionSelect` — the same dynamic dropdown component used in the connected flow. It consumes `OpenShiftVersionsContext` (already available in the wizard component tree) to fetch and display all available versions from the `/v2/openshift-versions` API.

### Changes
- Replace static `OcmOpenShiftVersion` with dynamic `OcmOpenShiftVersionSelect` in disconnected `BasicStep.tsx`
- Add typed Formik initial values (`openshiftVersion`, `customOpenshiftSelect`) required by the dropdown
- Remove unused imports (`ExternalLink`, `OCP_RELEASES_PAGE`, `OcmOpenShiftVersion`)

### Context
- This is the **standalone OCM path** (`libs/ui-lib/lib/ocm/`). The similar issue in the **CIM/ACM path** (`libs/ui-lib/lib/cim/`) is being fixed separately in #3723.
- Commit [0b7af138](https://github.com/openshift-assisted/assisted-installer-ui/commit/0b7af138) ([MGMT-22941](https://redhat.atlassian.net/browse/MGMT-22941)) previously fixed this as part of the "ABI above the sea" flow, but was reverted in #3517. This PR re-applies only the minimal version selection fix without the larger reverted flow.

### Known follow-ups (out of scope)
- CPU architecture is still hardcoded to `x86_64` in the same component
- `ReviewStep.tsx` also has a hardcoded `"4.20"` reference

## Test plan
- [ ] Navigate to Assisted Installer → Create Cluster → toggle "disconnected/air-gapped"
- [ ] Verify the OpenShift version dropdown shows all available versions (4.20, 4.21, 4.22, etc.)
- [ ] Verify "Show all available versions" modal works
- [ ] Verify connected flow (non-disconnected) is unaffected

## Summary by CodeRabbit

* **Refactor**
  * Updated the OpenShift version selection interface in the disconnected cluster wizard
  * Removed the external releases page link from the version selector

* **New Features**
  * Disconnected OpenShift version now persists through the wizard and is shown in the review/summary pages

[Review Change Stack](https://app.coderabbit.ai/change-stack/openshift-assisted/assisted-installer-ui/pull/3731?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterWizard/disconnected/BasicStep.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterWizard/disconnected/BasicStep.tsx
@@ -2,27 +2,47 @@ import * as React from 'react';
 import {
   ClusterWizardStep,
   TechnologyPreview,
-  ExternalLink,
-  OCP_RELEASES_PAGE,
   StaticTextField,
   useTranslation,
 } from '../../../../common';
+import { getDefaultOpenShiftVersion } from '../../../../common/components/ui/formik/utils';
 import { Flex, Grid, GridItem, Form, Content } from '@patternfly/react-core';
-import OcmOpenShiftVersion from '../../clusterConfiguration/OcmOpenShiftVersion';
+import OcmOpenShiftVersionSelect from '../../clusterConfiguration/OcmOpenShiftVersionSelect';
 import { useClusterWizardContext } from '../ClusterWizardContext';
 import ClusterWizardFooter from '../ClusterWizardFooter';
 import ClusterWizardNavigation from '../ClusterWizardNavigation';
 import { WithErrorBoundary } from '../../../../common/components/ErrorHandling/WithErrorBoundary';
 import InstallDisconnectedSwitch from './InstallDisconnectedSwitch';
-import { Formik } from 'formik';
+import { Formik, useFormikContext } from 'formik';
+import { useOpenShiftVersionsContext } from '../OpenShiftVersionsContext';
+
+type BasicStepFormikValues = {
+  openshiftVersion: string;
+  customOpenshiftSelect: string | null;
+};
+
+const VersionSyncEffect = () => {
+  const { values } = useFormikContext<BasicStepFormikValues>();
+  const { setDisconnectedOpenshiftVersion } = useClusterWizardContext();
+  React.useEffect(() => {
+    if (values.openshiftVersion) {
+      setDisconnectedOpenshiftVersion(values.openshiftVersion);
+    }
+  }, [values.openshiftVersion, setDisconnectedOpenshiftVersion]);
+  return null;
+};
 
 const BasicStep = () => {
   const { t } = useTranslation();
-  const { moveNext } = useClusterWizardContext();
+  const { moveNext, disconnectedOpenshiftVersion } = useClusterWizardContext();
+  const { latestVersions } = useOpenShiftVersionsContext();
+  const initialVersion =
+    disconnectedOpenshiftVersion || getDefaultOpenShiftVersion(latestVersions);
 
   return (
-    <Formik
-      initialValues={{}}
+    <Formik<BasicStepFormikValues>
+      initialValues={{ openshiftVersion: initialVersion, customOpenshiftSelect: null }}
+      enableReinitialize
       onSubmit={() => {
         // nothing to do
       }}
@@ -31,6 +51,7 @@ const BasicStep = () => {
         navigation={<ClusterWizardNavigation />}
         footer={<ClusterWizardFooter onNext={moveNext} />}
       >
+        <VersionSyncEffect />
         <WithErrorBoundary title="Failed to load Basic step">
           <Grid hasGutter>
             <GridItem>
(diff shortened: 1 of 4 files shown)


asadawar:
@jgyselov @ElayAharoni whenever you have a moment, can you please take a look and provide your feedback/approvals?

libs/ui-lib/lib/ocm/components/clusterWizard/disconnected/BasicStep.tsx
```diff
@@ -2,27 +2,30 @@ import * as React from 'react';
...
+
 const BasicStep = () => {
   const { t } = useTranslation();
   const { moveNext } = useClusterWizardContext();
 
   return (
-    <Formik
-      initialValues={{}}
+    <Formik<BasicStepFormikValues>
+      initialValues={{ openshiftVersion: '', customOpenshiftSelect: null }}
```
jgyselov: Let's set the initial value. We can fetch it from the OpenShiftVersionContext like this:

```
  const { latestVersions } = useOpenShiftVersionsContext();
  const defaultVersion = getDefaultOpenShiftVersion(latestVersions);
```

Then it's clear what the form is starting with.

jgyselov (commented):
The field is there but it doesn't _do_ anything. 😕 

Once you get to the "Review and download ISO" step, the OpenShift version is still hardcoded to `4.20` and the selected value is not saved/read/referenced anywhere.

I am assuming that it should be used as a parameter for the ISO download URL in the `ReviewStep` component? If so, you will need to implement some kind of a mechanism so that the selected OpenShift version can be read from there.

Once you have that, you can read from that on the Review page and use it to parametrise the ISO URL.

And you should also make sure that if the users navigate back to the Basic information step, the value they've selected is shown on the form  (and not the default one).

(cc @ElayAharoni in case I made some wrong assumptions about the disconnected flow)

jgyselov (changes requested):

libs/ui-lib/lib/ocm/components/clusterWizard/disconnected/BasicStep.tsx
asadawar: Done — addressed this and also wired the selected version through the full wizard flow:

- **BasicStep**: Fetches `latestVersions` from `useOpenShiftVersionsContext()` and sets `initialValues.openshiftVersion` to `getDefaultOpenShiftVersion(latestVersions)`. On back-navigation, restores the previously selected version from context.
- **ClusterWizardContext**: Added `disconnectedOpenshiftVersion` state (same pattern as `installDisconnected`) so the selected version persists across steps.
- **ReviewStep**: Reads `disconnectedOpenshiftVersion` from context instead of hardcoded `4.20`.
- Used a `VersionSyncEffect` component inside Formik to bridge Formik's local state to the shared context.

Download URL left as `latest/` for now since there's no confirmed version-specific URL pattern on mirror.openshift.com — happy to change if you know the pattern.

jgyselov: Exploring https://mirror.openshift.com/pub/cgw/assisted-installer-disconnected/, it looks like the pattern is `https://mirror.openshift.com/pub/cgw/assisted-installer-disconnected/${version}/agent-ove.x86_64.iso`.

But I see that the versions listed here do not correspond to the versions presented in the dropdown (fetched from assisted-service `/openshift-versions`), nor is there a `4.21` or `4.22` version.

In any case, it doesn't make sense to have a dropdown if the selected value has no real effect on anything. It just confuses the users.

asadawar: @jgyselov Good catch on the mirror versions not matching the API versions — I looked into it and you're right, the `latest/` path is what actually gets downloaded regardless of selection.

I think the cleanest fix is to remove the version dropdown entirely and show "Latest available" in the ReviewStep instead. A dropdown that doesn't control the download just creates false expectations for users.

This would mean:
- Remove `OcmOpenShiftVersionSelect` from BasicStep
- Show "Latest available" for the version in ReviewStep
- Revert the context wiring (`disconnectedOpenshiftVersion`) since it's no longer needed

The original bug (OCPBUGS-86070) was about the hardcoded `4.20` — this approach fixes that honestly without implying a choice that doesn't exist.

Does that work, or would you prefer a different approach?

jgyselov: If I look up "OpenShift latest", I am told that it would be 4.21, released in February. But looking at the mirror again, my guess is that `latest` there would be `4.20.8` (both versions added at the same time), which makes "Latest available" confusing and misleading information.

If I'm correct in my assumption about the versions, I think showing `4.20` as the OpenShift version is the most accurate option.

\---

Btw. before https://github.com/openshift-assisted/assisted-installer-ui/pull/3517 there was a way to select and set the OpenShift version. I don't know why this flow was reverted, though.

Question for @ElayAharoni or @rawagner ?

asadawar: Hello @jgyselov circling back on this. The discussion stalled on the mirror versioning question back in May. You raised a valid point that the mirror serves from `latest/` regardless of version selection, which makes the dropdown misleading.

What approach would you prefer here? I see two paths:

1. Remove the version dropdown, show the actual mirror version that will be downloaded (e.g. by querying the mirror metadata)
2. Keep the current hardcoded behavior but fix the display to not show a version that doesn't match what gets downloaded

Happy to rework the PR in whichever direction makes sense. Also, should I rebase now or wait until we agree on the approach?

PR #3732, opened 2026-05-20
AGENT-1519 | [Below-the-sea UI] Allow the user to select IPv6 in the Networking Stack Type field

asmasarw:
[AGENT-1519](https://redhat.atlassian.net/browse/AGENT-1519) | [Below-the-sea UI] Allow the user to select IPv6 in the Networking Stack Type field

## New Features
* Added explicit IPv6 single-stack networking stack type for cluster configuration
* Expanded cluster networking options to include dedicated IPv6 single-stack mode alongside existing IPv4 and dual-stack options
* Enhanced static IP configuration to support IPv6-only deployments

## Tests
* Added comprehensive test coverage for IPv6 single-stack and subnet selection scenarios

### DIFF
--- a/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
+++ b/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
@@ -252,9 +252,9 @@ export const AvailableSubnetsControl = ({
                   name="machineNetworks.0.cidr"
                   machineSubnets={primaryMachineSubnets}
                   isDisabled={isDisabled}
-                  onAfterSelect={(newSelection) =>
-                    reorderNetworksForPrimary(newSelection, values, setFieldValue)
-                  }
+                  onAfterSelect={(newSelection) => {
+                    reorderNetworksForPrimary(newSelection, values, setFieldValue);
+                  }}
                   data-testid="subnets-dropdown-toggle-primary"
                 />
               </StackItem>
--- a/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/StackTypeControlGroup.tsx
+++ b/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/StackTypeControlGroup.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ButtonVariant, FormGroup, Split, SplitItem, Tooltip } from '@patternfly/react-core';
+import { ButtonVariant, FormGroup, Radio, Split, SplitItem, Tooltip } from '@patternfly/react-core';
 import { useFormikContext } from 'formik';
 import { Address4, Address6 } from 'ip-address';
 
@@ -9,10 +9,9 @@ import {
   NETWORK_TYPE_OVN,
   NETWORK_TYPE_SDN,
 } from '../../../types';
-import { getFieldId } from '../../ui';
+import { ConfirmationModal, getFieldId, PopoverIcon } from '../../ui';
 import { DUAL_STACK, NO_SUBNET_SET, SINGLE_STACK } from '../../../config';
 import { RadioField } from '../../ui/formik';
-import { ConfirmationModal, PopoverIcon } from '../../ui';
 import { useTranslation } from '../../../hooks/use-translation-wrapper';
 import { reorderNetworksByCurrentPrimary } from './reorderNetworks';
 import {
@@ -72,36 +71,32 @@ export const StackTypeControlGroup = ({
 
   const IPv6Subnets = hostSubnets.filter((subnet) => Address6.isValid(subnet.subnet));
   const cidrIPv6 = IPv6Subnets.length >= 1 ? IPv6Subnets[0].subnet : NO_SUBNET_SET;
+  const isSingleStackSelected = values.stackType !== DUAL_STACK;
+
   const shouldSetSingleStack =
     !isViewerMode && !isDualStackSelectable && values.stackType === DUAL_STACK;
 
   const setSingleStack = React.useCallback(() => {
     setFieldValue('stackType', SINGLE_STACK);
 
-    // Determine whether the first machine network is IPv4
     const firstNetwork = values.machineNetworks?.[0];
     const isFirstIPv4 = Boolean(firstNetwork?.cidr && Address4.isValid(firstNetwork.cidr));
 
     if (values.machineNetworks && values.machineNetworks?.length >= 2) {
-      // For single-stack IPv4, prefer IPv4 machine network
       if (isFirstIPv4) {
-        // Keep the first network if it's IPv4
         setFieldValue('machineNetworks', [firstNetwork]);
       } else {
-        // If first is IPv6, look for IPv4 network or set empty for auto-selection
         const ipv4Network = values.machineNetworks.find(
           (network) => network.cidr && Address4.isValid(network.cidr),
         );
         if (ipv4Network) {
           setFieldValue('machineNetworks', [ipv4Network]);
         } else {
-          // No IPv4 network found, set empty for auto-selection
           setFieldValue('machineNetworks', []);
         }
       }
     }
 
-    // Set cluster/service to IPv4 defaults when switching to single-stack
     if (values.clusterNetworks && values.clusterNetworks.length >= 1) {
       const defaultCluster =
         defaultNetworkValues.clusterNetworksIpv4 && defaultNetworkValues.clusterNetworksIpv4[0];
@@ -110,12 +105,12 @@ export const StackTypeControlGroup = ({
         ? {
             cidr: defaultCluster.cidr,
             hostPrefix: defaultCluster.hostPrefix,
-            clusterId: firstCluster.clusterId, // ← Preserve clusterId from original network
+            clusterId: firstCluster.clusterId,
           }
         : {
             cidr: '',
             hostPrefix: '',
-            clusterId: clusterId, // ← Use current clusterId
+            clusterId: clusterId,
           };
       setFieldValue('clusterNetworks', [ipv4Cluster]);
     }
@@ -126,22 +121,20 @@ export const StackTypeControlGroup = ({
       const ipv4Service = defaultService
         ? {
             cidr: defaultService.cidr,
-            clusterId: firstService.clusterId, // ← Preserve clusterId from original network
+            clusterId: firstService.clusterId,
           }
         : {
             cidr: '',
-            clusterId: clusterId, // ← Use current clusterId
+            clusterId: clusterId,
           };
       setFieldValue('serviceNetworks', [ipv4Service]);
     }
 
-    // Keep only IPv4 VIPs when switching to single-stack (unless DHCP allocation is used)
     if (!values.vipDhcpAllocation) {
       const pruneVips = (
         vips?: { ip?: string; clusterId?: Cluster['id'] }[],
       ): { ip: string; clusterId: Cluster['id'] }[] => {
         if (!vips || vips.length === 0) return [];
-        // Choose which index to keep based on whether the first network is IPv4
         const keepIndex = isFirstIPv4 ? 0 : 1;
         const chosen = vips[keepIndex];
         return chosen && chosen.ip
@@ -177,12 +170,10 @@ export const StackTypeControlGroup = ({
     }
 
     if (values.machineNetworks && values.machineNetworks?.length < 2) {
-      // Ensure IPv4 is primary when switching to dual-stack
       const firstNetwork = values.machineNetworks[0];
       const isFirstIPv6 = firstNetwork?.cidr && Address6.isValid(firstNetwork.cidr);
 
       if (isFirstIPv6) {
-        // If first network is IPv6, make IPv4 primary and IPv6 secondary
         const IPv4Subnets = hostSubnets.filter((subnet) => Address4.isValid(subnet.subnet));
         const cidrIPv4 = IPv4Subnets.length >= 1 ? IPv4Subnets[0].subnet : NO_SUBNET_SET;
         setFieldValue(
@@ -194,7 +185,6 @@ export const StackTypeControlGroup = ({
           false,
         );
       } else {
-        // If first network is IPv4, add IPv6 as secondary
         setFieldValue(
           'machineNetworks',
           [...values.machineNetworks, { cidr: cidrIPv6, clusterId: clusterId }],
@@ -225,7 +215,6 @@ export const StackTypeControlGroup = ({
       };
       setFieldValue('serviceNetworks', [...values.serviceNetworks, secondServiceNetwork], false);
     }
-    // Initialize VIPs for dual-stack
     if (values.apiVips && values.apiVips.length < 2) {
       setFieldValue(
         'apiVips',
@@ -244,8 +233,9 @@ export const StackTypeControlGroup = ({
       );
     }
   };
-  // Ensure cluster/service networks ordering matches the new primary family
+
   const primaryCidr = values.machineNetworks?.[0]?.cidr;
+
   React.useEffect(() => {
     if (isViewerMode) {
       return;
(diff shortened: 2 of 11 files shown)


ElayAharoni (approved):

jgyselov:
These changes would affect the SaaS UI. Let me review before merging :slightly_smiling_face:

jgyselov:
I talked to BE and, unfortunately, IPv6-only is currently not an option for SaaS, meaning this feature will need to be controlled by the `ASSISTED_INSTALLER_SINGLE_CLUSTER_FEATURE` flag.

We already have something similar on the ACM side (see https://github.com/openshift-assisted/assisted-installer-ui/pull/3414 on how it's done there, please). We could make use of the shared `StackTypeControlGroup` and `AvailableSubnetsControl` components and set the `allowSingleStackIPv6` prop according to the single cluster feature flag—this will set the copy and options correctly.

We don't need a specific `"IPv6"` stack type option. We can just re-label "IPv4" to "Single-stack", extend the machine networks options, and tweak the validation schema.

(Potentially, we could get rid of the IPv4 label altogether and align everything under "Single-stack". But that's just some food for thought, not a requirement.)

Screenshot From 2026-06-11 15-16-55

When you make your changes, please make sure:

- IPv6 single stack is only allowed in the single cluster mode (SaaS is not affected)
- cluster and service networks are switched according to the selected machine network stack in the background (unless the user has selected the "Use advanced networking" option)
- switching between doesn't cause any issues

---

For the **static IP**, I'm not sure IPv6 only is supported (and I wasn't able to get an answer from the BE). Until we know for sure, could we _wait_ with the UI implementation of this?

If the users really want to, they can use the Form-view option to provide their yamls without any UI limitations so it's not like it's blocking anyone. (And it doesn't look like it was requested in the ticket either.)

cc @asmasarw @ElayAharoni 

---

And a complete side-note: when testing the changes, did neither of you come across these issues?

image

(IPv6 machine network was not selectable—all options were IPv4—and the advanced networking fields can completely disappear.)

image

(It's possible to submit an IPv6 network-wide configuration, but the UI crashes when it tries to read it.)

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/NetworkingStackTypeSelect.tsx
```diff
@@ -0,0 +1,87 @@
...
+                {option.tooltip ? (
+                  <Tooltip content={option.tooltip} hidden={option.isTooltipHidden ?? true}>
+                    {radio}
+                  </Tooltip>
+                ) : (
+                  radio
+                )}
+              </SplitItem>
+              <SplitItem>
+                <PopoverIcon bodyContent={option.popoverContent} buttonStyle={{ top: '-4px' }} />
```
jgyselov: You probably won't need this file at all. But if you're making changes here, can you remove the negative top-align? (I think that would look better and be more consistent with our other label popovers.)

With:

Image

Without:

Image

```suggestion
                <PopoverIcon bodyContent={option.popoverContent}  />
```

jgyselov (changes requested):

asmasarw:
Thanks @jgyselov For your review and Valuable Information.

Changes were applied:

IPv6 Selected:

image

IPv4 Selected:

image



SaaS Mode: (IPv6 NOT Preset - Controlled by the flag: ASSISTED_INSTALLER_SINGLE_CLUSTER_FEATURE):

image

libs/ui-lib/lib/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
```diff
@@ -177,25 +181,40 @@ const NetworkConfiguration = ({
...
+        if (isDualStack) {
+          serviceNetworkDefaults = defaultNetworkSettings.serviceNetworksDualstack;
+        } else if (isIPv6Stack) {
+          const ipv6Entry = getIPv6FromDualstack(defaultNetworkSettings.serviceNetworksDualstack);
+          serviceNetworkDefaults = ipv6Entry
+            ? [ipv6Entry]
+            : defaultNetworkSettings.serviceNetworksIpv4;
+        } else {
+          serviceNetworkDefaults = defaultNetworkSettings.serviceNetworksIpv4;
+        }
```
jgyselov: No reason to do these conditions twice 🙂 
```suggestion
        let clusterNetworkDefaults;
        let serviceNetworkDefaults;
        if (isDualStack) {
          clusterNetworkDefaults = defaultNetworkSettings.clusterNetworksDualstack;
          serviceNetworkDefaults = defaultNetworkSettings.serviceNetworksDualstack;
        } else if (isIPv6Stack) {
          const clusterNetworksIpv6 = getIPv6FromDualstack(
            defaultNetworkSettings.clusterNetworksDualstack,
          );
          clusterNetworkDefaults = clusterNetworksIpv6
            ? [clusterNetworksIpv6]
            : defaultNetworkSettings.clusterNetworksIpv4;

          const serviceNetworksIPv6 = getIPv6FromDualstack(
            defaultNetworkSettings.serviceNetworksDualstack,
          );
          serviceNetworkDefaults = serviceNetworksIPv6
            ? [serviceNetworksIPv6]
            : defaultNetworkSettings.serviceNetworksIpv4;
        } else {
          clusterNetworkDefaults = defaultNetworkSettings.clusterNetworksIpv4;
          serviceNetworkDefaults = defaultNetworkSettings.serviceNetworksIpv4;
        }

```

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/StackTypeControlGroup.tsx
```diff
@@ -38,6 +38,7 @@ export type StackTypeControlGroupProps = {
   defaultNetworkValues: StackTypeDefaultNetworkValues;
   isViewerMode?: boolean;
   allowSingleStackIPv6?: boolean;
+  showIPv6Option?: boolean;
```
jgyselov: There's probably no need to have both these props if you're going to always evaluate their disjunction. 🙂 Let's keep just `allowSingleStackIPv6` and use it to control the labels etc.

libs/ui-lib/lib/ocm/components/clusterConfiguration/review/utils.tsx
```diff
@@ -1,12 +1,20 @@
...
+import { Address6 } from 'ip-address';
 import { isDualStack } from '../../../../common';
 import { Cluster, DiskEncryption } from '@openshift-assisted/types/assisted-installer-service';
 
 export const getManagementType = ({ userManagedNetworking }: Cluster): string =>
   userManagedNetworking ? 'User-managed networking' : 'Cluster-managed networking';
 
-export const getStackTypeLabel = (cluster: Cluster): string =>
-  isDualStack(cluster) ? 'Dual-stack' : 'IPv4';
+export const getStackTypeLabel = (cluster: Cluster): string => {
```
jgyselov: I'd suggest to make it:

- `Dual-stack` for dual-stack clusters (duh)
- `IPv4` for non-dual-stack clusters in SaaS
- `Single-stack` for non-dual-stack clusters in ABI

to keep it in line with the Networking step options.

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -252,9 +292,18 @@ export const AvailableSubnetsControl = ({
...
+                  onAfterSelect={(newSelection) => {
+                    const previousCidr = values.machineNetworks?.[0]?.cidr;
+                    if (newSelection !== previousCidr && !values.vipDhcpAllocation) {
+                      setFieldValue('apiVips.0.ip', '', false);
+                      setFieldValue('ingressVips.0.ip', '', false);
+                      if (isDualStack) {
+                        setFieldValue('apiVips.1.ip', '', false);
+                        setFieldValue('ingressVips.1.ip', '', false);
+                      }
+                    }
```
jgyselov: Instead of deleting without a warning, let's be consistent with the rest of the app and leave changing the IP values to the user.

```suggestion
```

libs/ui-lib/lib/common/types/clusters.ts
```diff
@@ -46,7 +46,7 @@ export type NetworkConfigurationValues = Pick<
   | 'serviceNetworks'
 > & {
   managedNetworkingType: 'userManaged' | 'clusterManaged';
-  stackType?: 'singleStack' | 'dualStack';
+  stackType?: 'singleStack' | 'singleStackIpv6' | 'dualStack';
```
jgyselov: I think the code becomes unnecessarily complicated with the addition of `singleStackIpv6`.

It's not an option tied to a control the user can click but instead gets set based on the primary machine subnet stack. We could easily just have `singleStack` and `dualStack` only, get rid of a lot of condition branches for the `singleStackIpv6` option, and just base the _validation_ on the primary machine stack.

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -12,27 +12,62 @@ import { FieldArray, useFormikContext, FormikHelpers } from 'formik';
...
+  allSubnets: HostSubnet[];
+  IPv4Subnets: HostSubnet[];
+  IPv6Subnets: HostSubnet[];
+}): HostSubnet[] {
+  if ((isDualStack && supportsIPv6Primary) || (!isDualStack && allowSingleStackIPv6)) {
+    return allSubnets;
+  }
+  if (stackType === SINGLE_STACK_IPV6) {
+    return IPv6Subnets;
+  }
```
jgyselov: Your code never gets into this part because you are passing the `allowSingleStackIPv6` prop.

And any way, we probably don't want to get rid of the IPv4 options in the primary machine subnet dropdown (that would prevent the users to select single stack IPv4 altogether)!

The original implementation for selecting the primary machine subnet options was probably sufficient.

jgyselov (changes requested):

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
asmasarw: Removed.

libs/ui-lib/lib/common/types/clusters.ts
asmasarw: Done

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
asmasarw: OK, Didn't know about this behavior - I undo the changes.

libs/ui-lib/lib/ocm/components/clusterConfiguration/review/utils.tsx
asmasarw: Done

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/StackTypeControlGroup.tsx
asmasarw: Also DONE

libs/ui-lib/lib/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
asmasarw: Good Catch :)
I didn't notice this when I reviewed the changes :)
Done.

libs/ui-lib/lib/ocm/components/clusterConfiguration/review/utils.tsx
```diff
@@ -1,12 +1,21 @@
...
-export const getStackTypeLabel = (cluster: Cluster): string =>
-  isDualStack(cluster) ? 'Dual-stack' : 'IPv4';
+export const getStackTypeLabel = (cluster: Cluster, isSingleClusterFeature = false): string => {
+  if (isDualStack(cluster)) return 'Dual-stack';
+  if (isSingleClusterFeature) return 'Single stack';
+  const machineNetworks = cluster.machineNetworks ?? [];
+  if (machineNetworks.length === 1) {
+    const primaryCidr = machineNetworks[0]?.cidr;
+    if (primaryCidr && Address6.isValid(primaryCidr)) return 'IPv6';
+  }
```
jgyselov:
```suggestion
```

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -98,13 +133,16 @@ export const AvailableSubnetsControl = ({
...
+    !!values.machineNetworks?.[0]?.cidr && Address6.isValid(values.machineNetworks[0].cidr);
   useAutoSelectDefaultMachineNetworks(
     autoSelectNetwork,
     setFieldValue,
     cidrV4,
     cidrV6,
     clusterId,
     isDualStack,
+    isIPv6Stack,
   );
```
jgyselov: Autoselect runs only when `machineNetworks` is an empty array.

If `machineNetworks` is empty, `isIPv6Stack` will always be `false`.

My suggestion is to keep `useAutoSelectDefaultMachineNetworks` as-is, including the `cidrV4` and `cidrV6` types. It doesn't need any changes. If `machineNetworks` is empty (e.g. after switching from user-managed-networking, there is nothing wrong with an IPv4 default).

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
```diff
@@ -17,22 +17,57 @@ import { isMajorMinorVersionEqualOrGreater } from '../../../utils';
...
+  allSubnets: HostSubnet[];
+  IPv4Subnets: HostSubnet[];
+  IPv6Subnets: HostSubnet[];
+}): HostSubnet[] {
+  if ((isDualStack && supportsIPv6Primary) || (!isDualStack && allowSingleStackIPv6)) {
+    return allSubnets;
+  }
+  if (primaryCidr && Address6.isValid(primaryCidr)) {
+    return IPv6Subnets;
+  }
```
jgyselov: This function will always return either `allSubnets` or `IPv4Subnets`. This will never happen.

I'd suggest keeping the original implementation.

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/AvailableSubnetsControl.tsx
jgyselov, line 1: Looking at this file, I'm not sure you need any changes at all for your feature. Just passing the `allowSingleStackIPv6` prop correctly to this component should take care of everything.

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/StackTypeControlGroup.tsx
```diff
@@ -261,10 +252,60 @@ export const StackTypeControlGroup = ({
     values.serviceNetworks,
   ]);
 
-  const setStackType = (e: React.ChangeEvent<HTMLInputElement>) => {
-    if (e.target.value === DUAL_STACK) {
-      setDualStack();
-    } else if (
+  // When IPv6 is enabled, update cluster/service network defaults to match the selected
+  // machine network IP family so advanced fields stay consistent.
+  React.useEffect(() => {
```
jgyselov: This looks like similar logic to that in `NetworkConfiguration` (setting default values for the advanced networking fields)—any chance you could remove it from this file and do it solely in `NetworkConfiguration`? I'm pretty sure there's some logic that could be reused for both use-cases.

Additionally, there, you can also check if `Use advanced configuration` selected and only reset the values in that case.

If the users have `Use advanced configuration` selected, they might have made changes to those fields. It wouldn't be nice of us to override their custom values.

jgyselov (changes requested):

libs/ui-lib/lib/ocm/components/clusterConfiguration/networkConfiguration/NetworkConfiguration.tsx
jgyselov, line 1: Any chance some of these big code-blocks that handle setting default cluster/service networks could be extracted into functions and reused somehow?

jgyselov (changes requested):

jgyselov (approved):

PR #3733, opened 2026-05-20
OCPBUGS-85310: Static ip prefill for ove below the sea

ElayAharoni:
https://redhat.atlassian.net/browse/OCPBUGS-85310

## Summary by CodeRabbit

* **New Features**
  * Static IP form and page can prefill network data from the infrastructure environment when the single-cluster feature is enabled
  * Cluster creation flow now considers infrastructure environment data when determining the current step

* **Bug Fixes**
  * Empty fields now show only when prefill is unavailable
  * InfraEnv loading now respects the single-cluster feature and selects the available infra env when enabled

### DIFF
--- a/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/StaticIpForm.tsx
+++ b/libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/StaticIpForm.tsx
@@ -8,6 +8,8 @@ import { useErrorMonitor } from '../../../../../common/components/ErrorHandling/
 import { getApiErrorMessage } from '../../../../../common/api';
 import { StaticIpFormProps } from './propTypes';
 import { selectCurrentClusterPermissionsState } from '../../../../store/slices/current-cluster/selectors';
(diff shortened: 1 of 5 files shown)


libs/ui-lib/lib/ocm/components/clusterConfiguration/staticIp/components/StaticIpForm.tsx
```diff
@@ -104,7 +113,13 @@ export const StaticIpForm = <StaticIpFormValues extends object>({
 
   const onSubmit = isViewerMode ? () => Promise.resolve() : handleSubmit;
   return (
-    <Formik initialValues={initialValues} onSubmit={onSubmit} validate={validate} validateOnMount>
+    <Formik
+      enableReinitialize={prefillFromInfraEnv}
```
rawagner: I dont think there is ever a case where we want to enable reinitialization, which can reset the form at any time.

rawagner (approved):

rawagner (approved):

PR #3777, opened 2026-06-10
MGMT-24493: Allow HTTPS proxy for Discovery ISO

LiorSoffer:
https://redhat.atlassian.net/browse/MGMT-24493
https://redhat.atlassian.net/browse/MGMT-25058

BE: https://github.com/openshift/assisted-service/pull/10589

### ACM

**Create InfraEnv**
- Path: `libs/ui-lib/lib/cim/components/InfraEnv/InfraEnvFormPage.tsx`
- Resource: `InfraEnv.spec.proxy`

**Edit InfraEnv proxy**
- Path: `libs/ui-lib/lib/cim/components/modals/EditProxyModal.tsx`
- Resource: `InfraEnv.spec.proxy`

**Generate Discovery ISO / Add hosts**
- Path: `libs/ui-lib/lib/cim/components/modals/DiscoveryImageConfigForm.tsx`
- Resource: `InfraEnv.spec.proxy`

**Create cluster → Networking**
- Path: `libs/ui-lib/lib/cim/components/ClusterDeployment/networking/ClusterDeploymentNetworkingForm.tsx`
- Resource: `AgentClusterInstall.spec.proxy`

### Hypershift

**Network step**
- Path: `libs/ui-lib/lib/cim/components/Hypershift/HostedClusterWizard/NetworkStep/NetworkForm.tsx`
- Resource: `HostedCluster.spec.configuration.proxy`

### OCM

**Generate Discovery ISO**
- Path: `libs/ui-lib/lib/ocm/components/clusterConfiguration/OcmDiscoveryImageConfigForm.tsx`


.cursor/rules/translations.mdc
```diff
@@ -42,6 +42,7 @@ yarn workspace @openshift-assisted/locales run validate_translation_files
 - **`lib/es/`, `lib/fr/`, `lib/ja/`, `lib/ko/`, `lib/zh/`** — Handled by a separate localization
   process. **Do not edit** those `translation.json` files.
 - **`libs/ui-lib/lib/ocm/**`** — No **`t('ai:…')`**; use plain English strings or text supplied via props.
+- **Bare string literals in `i18nKey` or other translation props** — Always wrap `'ai:…'` strings with `t()` even when assigning to props like `i18nKey`.
```
jgyselov: ???

If it was "prefix keys with `ai:`" then it would be correct. But I'm not sure about wrapping with `t()` in `Trans` components.

libs/ui-lib/lib/common/validationSchemas/proxyValidation.tsx
```diff
@@ -48,6 +48,49 @@ export const httpProxyValidationSchema = ({
   );
 };
 
+export const httpsProxyValidationSchema = ({
```
jgyselov: This is a lot of code duplication. Could you reuse the parts that this has common with `httpProxyValidationSchema` somehow?

Maybe have a generic `proxyValidationSchema` that is used and chained with the http/https validation schemas?

jgyselov (changes requested):

.cursor/rules/translations.mdc
LiorSoffer: I wrote this a long time ago and I can't reproduce/remember the problem I was having :( 
Deleting this for now, and next time I encounter it I'll add an example.

jgyselov (approved):

PR #3908, opened 2026-07-22
MGMT-21975: The link to OpenShift networking documentation opens 4.15 documentation

jgyselov:
https://redhat.atlassian.net/browse/MGMT-21975

Based on https://github.com/openshift-assisted/assisted-installer-ui/pull/3271

### DIFF
--- a/libs/ui-lib/lib/common/config/docs_links.ts
+++ b/libs/ui-lib/lib/common/config/docs_links.ts
@@ -1,4 +1,22 @@
+import { getMajorMinorVersion, isMajorMinorVersionEqualOrGreater } from '../utils';
+
 const DEFAULT_OPENSHIFT_DOCS_VERSION = 4.15;
+
+const LATEST_OPENSHIFT_DOCS_VERSION = '4.21';
+const MIN_OPENSHIFT_DOCS_VERSION = '4.14';
+const getDocsOpenshiftVersion = (ocpVersion?: string): string => {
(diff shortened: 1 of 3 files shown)


libs/ui-lib/lib/common/config/docs_links.ts
```diff
@@ -1,5 +1,27 @@
...
+  }
+  if (isMajorMinorVersionEqualOrGreater(ocpVersion, LATEST_OPENSHIFT_DOCS_VERSION)) {
+    return LATEST_OPENSHIFT_DOCS_VERSION;
+  }
+
+  const versionParts = ocpVersion.split('.').slice(0, 2).map(Number);
+  if (versionParts.length < 2) {
+    return LATEST_OPENSHIFT_DOCS_VERSION;
+  }
+  return `${versionParts[0]}.${versionParts[1]}`;
```
LiorSoffer: we can reuse `getMajorMinorVersion` here
also,  checking `versionParts.length >=2`  doesn't guarantee a valid version

maybe something like this could work:

```suggestion
  const majorMinor = getMajorMinorVersion(ocpVersion);
  return majorMinor || LATEST_OPENSHIFT_DOCS_VERSION;
```

jgyselov: Done 🙂

LiorSoffer (approved):

PR #3934, opened 2026-08-02
MGMT-24494: Add exclusive NTP sources to the Add hosts discovery ISO …

LiorSoffer:
https://redhat.atlassian.net/browse/MGMT-24494

This adds a way to configure replace-default NTP ( `ntpSources` ) .
In the Add hosts modal you can check Replace default NTP sources and enter a comma-separated list. 
On generate, that goes to both the cluster and the InfraEnv so discovery hosts use only those servers (not the default public pools). Handy for air-gapped or locked-down networks.

image

image

image

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring custom NTP sources when generating discovery images.
  * Users can replace default NTP sources with a comma-separated list of custom sources.
  * Custom NTP settings are saved for cluster and infrastructure environment configurations.
  * Added guidance for updating custom NTP sources and regenerating discovery images.

* **Bug Fixes**
  * Improved NTP synchronization failure messages and translated dialog labels.
  * Prevented the option to add NTP sources from appearing when custom sources are already configured.


libs/types/assisted-installer-service.d.ts
```diff
@@ -1896,6 +1896,10 @@ export interface InfraEnv {
    * A comma-separated list of NTP sources (name or IP) going to be added to all the hosts.
    */
   additionalNtpSources?: string;
+  /**
+   * A comma-separated list of NTP sources (name or IP) to be used as the only NTP configuration for hosts in this infra-env.
+   */
+  ntpSources?: string;
```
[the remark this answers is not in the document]
LiorSoffer: I kept only the relevant changes for this PR

build-tools
jgyselov, line 1: Remove, please.

libs/ui-lib/diff.txt
jgyselov, line 1: Remove, please.

libs/ui-lib/lib/ocm/services/DiscoveryImageFormService.ts
```diff
@@ -7,8 +7,20 @@ import {
...
+const getNtpUpdateParams = (formValues: OcmDiscoveryImageFormValues) => {
+  const ntpList =
+    formValues.enableNtpSources && formValues.ntpSourcesList
+      ? trimCommaSeparatedList(formValues.ntpSourcesList)
+      : undefined;
+
+  return ntpList?.length
+    ? { clusterNtpParams: { ntpSources: ntpList }, infraEnvNtpParams: { ntpSources: ntpList } }
+    : { clusterNtpParams: {}, infraEnvNtpParams: {} };
+};
```
jgyselov: What if you did something like

```suggestion
const getNtpUpdateParams = (formValues: OcmDiscoveryImageFormValues) => {
  const ntpSources =
    formValues.enableNtpSources && formValues.ntpSourcesList
      ? trimCommaSeparatedList(formValues.ntpSourcesList)
      : '';

  return ntpSources;
};
```
and then you can add `ntpSources` directly to the cluster/infra env update parameters?

(Plus, setting `nptSources` to an empty string clears the field which is a functionality we're kinda lacking right now.)

libs/ui-lib/lib/ocm/components/clusterConfiguration/OcmDiscoveryImageConfigForm.tsx
jgyselov: Let's up the `maxHeight` to maybe `750px` so that the modal doesn't by default have a scrollbar.

libs/ui-lib/lib/common/components/clusterConfiguration/NtpSourcesFields.tsx
```diff
@@ -0,0 +1,52 @@
...
+      name="enableNtpSources"
+      className="ai-ntp-sources-fields"
+      helperText={
+        <p>
+          {t(
+            'ai:Replaces default public NTP pools. Only the specified sources are used on discovery hosts and during installation. Use in air-gapped or restricted networks where default pools are unreachable or not permitted.',
+          )}
+        </p>
+      }
+      onChange={(value: boolean) => resetNtpSources(value)}
```
jgyselov: We can simplify to 

```suggestion
      onChange={resetNtpSources}
```

libs/ui-lib/lib/common/components/clusterConfiguration/NtpSourcesFields.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { setFieldValue, values, initialValues } = useFormikContext<NtpSourcesFieldsType>();
+  const { t } = useTranslation();
+
+  const resetNtpSources = (isNewlyChecked: boolean) => {
+    if (isNewlyChecked) {
+      setFieldValue('ntpSourcesList', initialValues.ntpSourcesList);
+    } else {
+      setFieldValue('ntpSourcesList', '');
+    }
+  };
```
jgyselov: Can we rename `isNewlyChecked` to just `isChecked` or something? The _newly_ is messing with me 😅

jgyselov (commented):
Also, please try the following flow:

1. set NTP sources (or maybe have them set and then edit them)
2. generate ISO
3. click "Edit ISO configuration"

you'll see the old NTP values. You should see the new ones.

(Try with the proxy fields for comparison.)

jgyselov (changes requested):

libs/ui-lib/lib/common/components/clusterConfiguration/NtpSourcesFields.tsx
LiorSoffer: Sorry it is weird

libs/ui-lib/lib/common/config/constants.ts
```diff
@@ -158,7 +158,7 @@ export const hostValidationLabels = (t: TFunction): { [key in HostValidationId]:
 });
 
 export const hostValidationFailureHints = (
-  t: TFunction,
+  _t: TFunction,
```
jgyselov: You could remove the `t` argument altogether if there's nothing to translate. (Or, I'd maybe consider removing the function altogether if the result is always an empty string 😅)

libs/ui-lib/lib/ocm/components/hosts/AdditionaNTPSourceDialogToggle.tsx
```diff
@@ -1,11 +1,33 @@
...
+    );
+  }
 
-  return <AlertActionLink onClick={() => open()}>Add NTP sources</AlertActionLink>;
+  return (
+    <>
+      <div>
+        {t("ai:Manually fix the host's NTP configuration or provide additional NTP sources.")}
+      </div>
+      <br />
```
jgyselov: What if we used the PatternFly components instead of divs and brs? 😄 (Same suggestion applies for the other message.)

```suggestion
       <Content>
        {t("ai:Manually fix the host's NTP configuration or provide additional NTP sources.")}
      </Content>
```

libs/ui-lib/lib/ocm/components/hostsTable/components/AdditionaNTPSourceDialogToggle.tsx
jgyselov, line 1: I'm not sure this is the best place for these messages. The component ends up in the popover footer (which is maybe not ideal for formatting?)

Image

[NtpValidationStatus.tsx](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/common/components/hosts/NtpValidationStatus.tsx) might be a better place for it. (I'd just put he text in the popover children. And you can use the logic there to optionally attach the link component for the dialog.

libs/ui-lib/lib/ocm/components/hosts/AdditionaNTPSourceDialogToggle.tsx
```diff
@@ -1,11 +1,33 @@
...
 import { useModalDialogsContext } from './ModalDialogsContext';
+import { useTranslation } from '../../../common/hooks/use-translation-wrapper';
+import { selectCurrentCluster } from '../../store/slices/current-cluster/selectors';
 
 export const AdditionalNTPSourcesDialogToggle = () => {
   const {
     additionalNTPSourcesDialog: { open },
   } = useModalDialogsContext();
+  const { t } = useTranslation();
+  const cluster = useSelector(selectCurrentCluster);
```
jgyselov: This might not work for Day2. I think it would be better to just drill down the cluster (or the select relevant data) from the tables.

.cursor/rules/form-state-patterns.mdc
jgyselov, line 1: Let's remove for now.

jgyselov (changes requested):

libs/ui-lib/lib/common/config/constants.ts
LiorSoffer: removed the t but keeping it , maybe we'll use it in the future

jgyselov: If we need something like this in the future, we can add it again. But I don't see the point of keeping a function which always returns an empty string.

libs/ui-lib/lib/common/components/clusterConfiguration/NtpSourcesFields.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <CheckboxField
+      label={t('ai:Replace default NTP sources')}
+      name="enableNtpSources"
+      className="ai-ntp-sources-fields"
+      helperText={
+        <p>
+          {t(
+            'ai:Replaces default public NTP pools. Only the specified sources are used on discovery hosts and during installation. Use in air-gapped or restricted networks where default pools are unreachable or not permitted.',
```
jgyselov: I'm not sure about some of the wording here. What if we did

```
Only the specified sources are used on hosts and during installation.
```
or 
```
Only the specified sources are used during host discovery and installation.
```
We don't use the term `discovery host` anywhere else in the project. I don't know where it came from but it doesn't sound right to me.

libs/ui-lib/lib/common/components/clusterConfiguration/NtpSourcesFields.tsx
```diff
@@ -0,0 +1,52 @@
...
+          )}
+        </p>
+      }
```
jgyselov: You probably don't need to wrap this in `p`
```suggestion
      helperText={t(
        'ai:Replaces default public NTP pools. Only the specified sources are used on discovery hosts and during installation. Use in air-gapped or restricted networks where default pools are unreachable or not permitted.',
      )}
```

libs/ui-lib/lib/ocm/components/hosts/NtpSyncFailureMessage.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const NtpSyncFailureMessage = () => {
+  const reduxCluster = useSelector(selectCurrentCluster);
+  const { cluster: day2Cluster } = React.useContext(AddHostsContext);
+  const cluster = day2Cluster || reduxCluster;
+
+  if (cluster?.ntpSources?.trim()) {
+    return (
+      <>
+        This cluster replaces default NTP sources with a custom list. To update the list, change NTP
+        sources in the Add hosts dialog, generate a new discovery ISO, and reboot discovery hosts.
```
jgyselov: Again, just `hosts` is probably sufficient.

```suggestion
        This cluster replaces default NTP sources with a custom list. To update the list, change NTP
        sources in the Add hosts dialog, generate a new discovery ISO, and reboot your hosts.
```

jgyselov (changes requested):

jgyselov (approved):

PR #3959, opened 2026-08-13
[release-4.22] OCPBUGS-104483: Pin minimatch to ^10.2.6 and brace-expansion to ^5.0.9 (CVE-2026-14257)

sunku5494:


germanparente:
@sunku5494 has checked and the fix is alreadty in main. That's the reason why I have skipped the check

jgyselov (approved):

PR #3996, opened 2026-08-27
MGMT-25128: Assisted Installer UI (above-the-sea) adjustment to 4.22 GA

asmasarw:
https://redhat.atlassian.net/browse/MGMT-25128

## Summary by CodeRabbit

* **New Features**
  * Disconnected installations now support rendezvous IP and host-network configuration in the optional configuration step.
  * Static IP setup is integrated consistently into the disconnected installation wizard.
  * Wizard navigation now displays sequential step numbers, including grouped static IP sub-steps.
  * Review screens display the rendezvous IP when available.

* **Bug Fixes**
  * Improved handling of disconnected wizard navigation and configuration updates.
  * Empty IP address fields are now accepted when optional.
  * ISO regeneration alerts no longer appear for disconnected installations.

### DIFF
--- a/libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx
+++ b/libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx
@@ -17,7 +17,7 @@ export const ipValidationSchema = (t: TFunction) =>
   Yup.string().test(
     'ip-validation',
     t('ai:Not a valid IP address'),
-    (value?: string) => Address4.isValid(value || '') || Address6.isValid(value || ''),
+    (value?: string) => !value || Address4.isValid(value) || Address6.isValid(value),
   );
 
 export const ipNoSuffixValidationSchema = (t: TFunction) =>
--- a/libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx
+++ b/libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx
@@ -23,7 +23,7 @@ export const ipValidationSchema = (t: TFunction) =>
 export const ipNoSuffixValidationSchema = (t: TFunction) =>
   Yup.string().test('ip-validation-no-suffix', t('ai:Not a valid IP address'), (value?: string) => {
     if (!value) return true;
-    const address = getAddress(value || '');
+    const address = getAddress(value);
     return !!address && address.address === address.addressMinusSuffix;
   });
 
--- a/libs/ui-lib/lib/ocm/components/wizard/NewClusterWizard.tsx
+++ b/libs/ui-lib/lib/ocm/components/wizard/NewClusterWizard.tsx
@@ -13,11 +13,7 @@ import {
 import { useClusterWizardContext } from './clusterWizardContext';
 import { ClusterWizardStepsType } from './utils';
 
-const getCurrentStep = (
-  currentStepId: ClusterWizardStepsType,
-  installDisconnected: boolean,
-  infraEnv?: InfraEnv,
-) => {
+const getCurrentStep = (currentStepId: ClusterWizardStepsType, infraEnv?: InfraEnv) => {
   switch (currentStepId) {
     case 'disconnected-review':
       return <DisconnectedReviewStep />;
@@ -28,10 +24,7 @@ const getCurrentStep = (
     case 'static-ip-yaml-view':
     case 'static-ip-network-wide-configurations':
     case 'static-ip-host-configurations':
-      if (installDisconnected) {
-        return <DisconnectedStaticIp />;
-      }
-      return <ClusterDetails infraEnv={infraEnv} />;
+      return <DisconnectedStaticIp />;
     default:
       return <ClusterDetails infraEnv={infraEnv} />;
   }
@@ -39,11 +32,11 @@ const getCurrentStep = (
 
 export const NewClusterWizard: React.FC = () => {
   const { infraEnv } = useInfraEnv('', CpuArchitecture.USE_DAY1_ARCHITECTURE);
-  const { currentStepId, installDisconnected } = useClusterWizardContext();
+  const { currentStepId } = useClusterWizardContext();
 
   return (
     <div className={classNames('pf-v6-c-wizard', 'cluster-wizard')}>
-      {getCurrentStep(currentStepId, installDisconnected, infraEnv)}
+      {getCurrentStep(currentStepId, infraEnv)}
     </div>
   );
 };
--- a/libs/ui-lib/lib/ocm/components/wizard/NewClusterWizard.tsx
+++ b/libs/ui-lib/lib/ocm/components/wizard/NewClusterWizard.tsx
@@ -8,11 +8,16 @@ import {
   BasicStep,
   OptionalConfigurationsStep,
   DisconnectedReviewStep,
+  DisconnectedStaticIp,
 } from './steps';
 import { useClusterWizardContext } from './clusterWizardContext';
 import { ClusterWizardStepsType } from './utils';
 
-const getCurrentStep = (currentStepId: ClusterWizardStepsType, infraEnv?: InfraEnv) => {
+const getCurrentStep = (
+  currentStepId: ClusterWizardStepsType,
+  installDisconnected: boolean,
+  infraEnv?: InfraEnv,
+) => {
   switch (currentStepId) {
     case 'disconnected-review':
       return <DisconnectedReviewStep />;
@@ -20,6 +25,13 @@ const getCurrentStep = (currentStepId: ClusterWizardStepsType, infraEnv?: InfraE
       return <OptionalConfigurationsStep />;
     case 'disconnected-basic':
       return <BasicStep />;
+    case 'static-ip-yaml-view':
+    case 'static-ip-network-wide-configurations':
+    case 'static-ip-host-configurations':
+      if (installDisconnected) {
+        return <DisconnectedStaticIp />;
+      }
+      return <ClusterDetails infraEnv={infraEnv} />;
     default:
       return <ClusterDetails infraEnv={infraEnv} />;
   }
@@ -27,11 +39,11 @@ const getCurrentStep = (currentStepId: ClusterWizardStepsType, infraEnv?: InfraE
 
 export const NewClusterWizard: React.FC = () => {
   const { infraEnv } = useInfraEnv('', CpuArchitecture.USE_DAY1_ARCHITECTURE);
-  const { currentStepId } = useClusterWizardContext();
+  const { currentStepId, installDisconnected } = useClusterWizardContext();
 
   return (
     <div className={classNames('pf-v6-c-wizard', 'cluster-wizard')}>
-      {getCurrentStep(currentStepId, infraEnv)}
+      {getCurrentStep(currentStepId, installDisconnected, infraEnv)}
     </div>
   );
 };
--- a/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContext.tsx
+++ b/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContext.tsx
@@ -23,6 +23,9 @@ export type ClusterWizardContextType = {
   setDisconnectedCluster: (cluster: Cluster | undefined) => void;
   disconnectedInfraEnv?: InfraEnv;
   setDisconnectedInfraEnv: (infraEnv: InfraEnv | undefined) => void;
+  /** UI-only: network config type selected in disconnected optional configurations step. */
+  disconnectedHostsNetworkConfigurationType?: 'dhcp' | 'static';
+  setDisconnectedHostsNetworkConfigurationType: (value: 'dhcp' | 'static') => void;
 };
 
 export const ClusterWizardContext = React.createContext<ClusterWizardContextType | null>(null);
--- a/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContext.tsx
+++ b/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContext.tsx
@@ -23,9 +23,6 @@ export type ClusterWizardContextType = {
   setDisconnectedCluster: (cluster: Cluster | undefined) => void;
   disconnectedInfraEnv?: InfraEnv;
   setDisconnectedInfraEnv: (infraEnv: InfraEnv | undefined) => void;
-  /** UI-only: network config type selected in disconnected optional configurations step. */
-  disconnectedHostsNetworkConfigurationType?: 'dhcp' | 'static';
-  setDisconnectedHostsNetworkConfigurationType: (value: 'dhcp' | 'static') => void;
 };
 
 export const ClusterWizardContext = React.createContext<ClusterWizardContextType | null>(null);
--- a/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
+++ b/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
@@ -173,15 +173,19 @@ export const ClusterWizardContextProvider = ({
       onUpdateHostNetworkConfigType(type: HostsNetworkConfigurationType): void {
         if (installDisconnected) {
           if (type === HostsNetworkConfigurationType.STATIC) {
-            setDisconnectedWizardStepIds(getDisconnectedWizardStepIds(StaticIpView.FORM));
+            const staticIpInfo = disconnectedInfraEnv
+              ? getStaticIpInfo(disconnectedInfraEnv)
+              : undefined;
+            const view = staticIpInfo?.view ?? StaticIpView.FORM;
+            setDisconnectedWizardStepIds(getDisconnectedWizardStepIds(view));
           } else {
             setDisconnectedWizardStepIds(getDisconnectedWizardStepIds('dhcp-selected'));
           }
         } else {
           if (type === HostsNetworkConfigurationType.STATIC) {
-            setWizardStepIds(
-              getWizardStepIds(wizardStepIds, StaticIpView.FORM, isSingleClusterFeatureEnabled),
-            );
+            const staticIpInfo = infraEnv ? getStaticIpInfo(infraEnv) : undefined;
+            const view = staticIpInfo?.view ?? StaticIpView.FORM;
+            setWizardStepIds(getWizardStepIds(wizardStepIds, view, isSingleClusterFeatureEnabled));
           } else {
             setWizardStepIds(
               getWizardStepIds(wizardStepIds, 'dhcp-selected', isSingleClusterFeatureEnabled),
--- a/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
+++ b/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
@@ -12,7 +12,6 @@ import {
 import useSetClusterPermissions from '../../../hooks/useSetClusterPermissions';
 import { HostsNetworkConfigurationType } from '../../../services';
 import { isOciPlatformType } from '../../utils';
-import { defaultWizardSteps, staticIpFormViewSubSteps } from '../constants';
 import { StaticIpView } from '../steps/staticIp/data/dataTypes';
 import { getStaticIpInfo } from '../steps/staticIp/data/fromInfraEnv';
 import {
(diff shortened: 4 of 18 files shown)


jgyselov (commented):
This alert is not relevant in the above-the-sea flow. It should not be shown in any of the steps.

Image

Image

image

libs/ui-lib/lib/ocm/components/wizard/NewClusterWizard.tsx
```diff
@@ -8,30 +8,42 @@ import {
...
       return <OptionalConfigurationsStep />;
     case 'disconnected-basic':
       return <BasicStep />;
+    case 'static-ip-yaml-view':
+    case 'static-ip-network-wide-configurations':
+    case 'static-ip-host-configurations':
+      if (installDisconnected) {
+        return <DisconnectedStaticIp />;
+      }
+      return <ClusterDetails infraEnv={infraEnv} />;
```
jgyselov: The `currentStepId` should be controlled by ClusterWizardContext. We shouldn't need the distinction for disconnected. In the classic SaaS flow, in `NewClusterWizard`, the `currentStepId` will never be anything else but `cluster-details`.

```suggestion
    case 'static-ip-yaml-view':
    case 'static-ip-network-wide-configurations':
    case 'static-ip-host-configurations':
        return <DisconnectedStaticIp />;
```

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -73,6 +95,26 @@ const OptionalConfigurationsForm: React.FC<OptionalConfigurationsFormProps> = ({
...
+const buildInfraEnvUpdateParams = (
+  values: OptionalConfigurationsValues,
+  disconnectedInfraEnv: InfraEnv | undefined,
+): InfraEnvUpdateParams => ({
+  sshAuthorizedKey: values.sshPublicKey || undefined,
+  ...(values.rendezvousIp && { rendezvousIp: values.rendezvousIp }),
+  ...(shouldSendDummyStaticConfig(values, disconnectedInfraEnv) && {
+    staticNetworkConfig: getDummyInfraEnvField(),
+  }),
+});
```
jgyselov, line 119: This fails to clear the values in the BE when the user clears the field or selects DHCP. It's a big regression.

```suggestion
const buildInfraEnvUpdateParams = (
  values: OptionalConfigurationsValues,
  disconnectedInfraEnv: InfraEnv | undefined,
): InfraEnvUpdateParams => ({
  sshAuthorizedKey: values.sshPublicKey,
  rendezvousIp: values.rendezvousIp,
  staticNetworkConfig: shouldSendDummyStaticConfig(values, disconnectedInfraEnv) ?  getDummyInfraEnvField() : "", 
});
```

(just a suggestion, untested; but we need something like this)

libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
```diff
@@ -162,17 +167,28 @@ export const ClusterWizardContextProvider = ({
...
-        throw `Wizard step is currently ${currentStepId}, but no static ip info is defined`;
+      if (installDisconnected) {
+        const staticIpInfo = disconnectedInfraEnv
+          ? getStaticIpInfo(disconnectedInfraEnv)
+          : undefined;
+        if (!staticIpInfo) {
+          // No static config persisted yet (e.g. first pass before API round-trip completes).
+          // Keep the current disconnectedWizardStepIds as-is so the user can still navigate back.
+          return;
+        }
```
jgyselov, line 122: I don't understand this part. Can you explain why it's here?

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/DisconnectedReviewStep.tsx
```diff
@@ -91,6 +81,14 @@ export const DisconnectedReviewStep = () => {
             </List>
           </Alert>
           <DescriptionList isHorizontal>
+            {disconnectedInfraEnv?.rendezvousIp && (
```
jgyselov, line 90: I would put this below the OpenShift version, not right at the top.

libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContext.tsx
```diff
@@ -23,6 +23,9 @@ export type ClusterWizardContextType = {
   setDisconnectedCluster: (cluster: Cluster | undefined) => void;
   disconnectedInfraEnv?: InfraEnv;
   setDisconnectedInfraEnv: (infraEnv: InfraEnv | undefined) => void;
+  /** UI-only: network config type selected in disconnected optional configurations step. */
+  disconnectedHostsNetworkConfigurationType?: 'dhcp' | 'static';
+  setDisconnectedHostsNetworkConfigurationType: (value: 'dhcp' | 'static') => void;
```
jgyselov: Any chance we could reuse the existing `onUpdateStaticIpView`, `onUpdateHostNetworkConfigType` props here?

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/DisconnectedStaticIp.tsx
jgyselov, line 1: Any chance you could (maybe slightly tweak and) reuse https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/ocm/components/wizard/steps/staticIp/StaticIp.tsx, so we have less code duplication?

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/DisconnectedStaticIp.tsx
```diff
@@ -0,0 +1,74 @@
...
+        throw new Error('No disconnected infraEnv available');
+      }
+      const { data: updatedInfraEnv } = await InfraEnvsAPI.update(disconnectedInfraEnv.id, params);
+      setDisconnectedInfraEnv(updatedInfraEnv);
+      return updatedInfraEnv;
+    },
+    [disconnectedInfraEnv, setDisconnectedInfraEnv],
+  );
+
+  const isNextDisabled = !formState.isValid || !!alerts.length || formState.isSubmitting;
```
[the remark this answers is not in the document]
jgyselov: cc @asmasarw

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -63,8 +77,16 @@ const OptionalConfigurationsForm: React.FC<OptionalConfigurationsFormProps> = ({
           </GridItem>
           <GridItem>
             <Form id="wizard-cluster-optional-config__form">
+              <InputField
+                label="Rendezvous IP"
+                name="rendezvousIp"
+                helperText="The IP address that hosts will use to communicate with the bootstrap node during installation."
+                placeholder="e.g., 192.168.1.10"
```
jgyselov: This is a very specific IP (usually reserved for routers etc.), I'm not sure it's a good idea to use this as a placeholder. Let's leave the placeholder out.

```suggestion
```

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -91,13 +135,27 @@ export const OptionalConfigurationsStep = () => {
       Yup.object({
         sshPublicKey: sshPublicKeyValidationSchema(t),
         pullSecret: isInOcm ? Yup.string() : pullSecretValidationSchema(t),
+        rendezvousIp: Yup.string()
+          .max(45, 'IP address must be at most 45 characters')
```
jgyselov, line 140: This test is probably obsolete since we have `ipValidationSchema`.
```suggestion
```

jgyselov: Also, everywhere in the codebase `ipValidationSchema` seems to be always preceeded by a 
```
if (!value) {
    return true;
}
```
block. Maybe we could make it part of the validation schema and remove all those checks?

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -91,13 +135,27 @@ export const OptionalConfigurationsStep = () => {
...
+          .test('ip-validation', 'Not a valid IP address', (value) => {
+            if (!value) return true;
+            return ipValidationSchema(t).isValidSync(value);
+          }),
+        hostsNetworkConfigurationType: Yup.string()
+          .oneOf(Object.values(HostsNetworkConfigurationType))
+          .required(),
```
jgyselov: The users are selecting from two radio buttons. There is no way they could select a wrong value, therefore this validation is completely obsolete.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -126,10 +182,34 @@ export const OptionalConfigurationsStep = () => {
             imageType: DISCONNECTED_IMAGE_TYPE,
             openshiftVersion: DISCONNECTED_OPENSHIFT_VERSION,
             sshAuthorizedKey: values.sshPublicKey || undefined,
+            ...(values.rendezvousIp && { rendezvousIp: values.rendezvousIp }),
+            ...(values.hostsNetworkConfigurationType === HostsNetworkConfigurationType.STATIC && {
+              staticNetworkConfig: getDummyInfraEnvField(),
+            }),
```
jgyselov: Let's continue the existing pattern instead of destructuring.


```suggestion
            rendezvousIp: values.rendezvousIp || undefined,
            staticNetworkConfig: values.hostsNetworkConfigurationType === HostsNetworkConfigurationType.STATIC ? getDummyInfraEnvField() : undefined,
```

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -126,10 +182,34 @@ export const OptionalConfigurationsStep = () => {
...
+          ...(values.hostsNetworkConfigurationType === HostsNetworkConfigurationType.STATIC &&
+            !infraEnvToUse.staticNetworkConfig && {
+              staticNetworkConfig: JSON.stringify(getDummyInfraEnvField()),
+            }),
+        });
+        setDisconnectedHostsNetworkConfigurationType(
+          values.hostsNetworkConfigurationType === HostsNetworkConfigurationType.STATIC
+            ? 'static'
+            : 'dhcp',
+        );
```
jgyselov: Why not just 
```suggestion
        setDisconnectedHostsNetworkConfigurationType(
          values.hostsNetworkConfigurationType
        );
```
?

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -126,10 +182,34 @@ export const OptionalConfigurationsStep = () => {
...
 
+        setDisconnectedInfraEnv({
+          ...infraEnvToUse,
+          sshAuthorizedKey: values.sshPublicKey || undefined,
+          rendezvousIp: values.rendezvousIp || undefined,
+          ...(values.hostsNetworkConfigurationType === HostsNetworkConfigurationType.STATIC &&
+            !infraEnvToUse.staticNetworkConfig && {
+              staticNetworkConfig: JSON.stringify(getDummyInfraEnvField()),
+            }),
+        });
```
jgyselov: Why not just 

```suggestion
        setDisconnectedInfraEnv(infraEnvToUse);
```
?

libs/ui-lib/lib/ocm/components/wizard/utils/wizardTransition.ts
jgyselov, line 1: This is not a good place for these functions. If you need them, put them inside https://github.com/openshift-assisted/assisted-installer-ui/tree/master/libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext (either in the provider file or consider creating `utils.ts` and moving the helper functions in there).

But my biggest question is—there are functions for adding and removing static IP steps already. Can you tweak and reuse them instead of adding a duplicate implementation?

jgyselov (changes requested):

libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
asmasarw: when the user navigates away from a static IP step, we call `getStaticIpInfo(disconnectedInfraEnv) ` to rebuild the wizard step list. But on the first pass (before InfraEnvsAPI.register completes), disconnectedInfraEnv has no staticNetworkConfig yet, so getStaticIpInfo returns undefined. 

Without this guard, we'd call `getDisconnectedWizardStepIds(undefined)` which would reset the step list to the base 3 disconnected steps, stripping out the static IP sub-steps and breaking back-navigation. 

The early return preserves the current disconnectedWizardStepIds so the user can still navigate backward. 

**(The SaaS flow throws here instead because an existing infraEnv is always available.)**

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: DONE

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/DisconnectedReviewStep.tsx
asmasarw: DONE

libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx
```diff
@@ -17,7 +17,7 @@ export const ipValidationSchema = (t: TFunction) =>
   Yup.string().test(
     'ip-validation',
     t('ai:Not a valid IP address'),
-    (value?: string) => Address4.isValid(value || '') || Address6.isValid(value || ''),
+    (value?: string) => !value || Address4.isValid(value) || Address6.isValid(value),
```
jgyselov, line 20: Nice! But please, go through all places where `ipValidationSchema` is used and remove the 
```
if (!value) {
      return true;
}
```
checks which preceede the call. They're not necessary anymore.

libs/ui-lib/lib/ocm/components/wizard/clusterWizardContext/ClusterWizardContextProvider.tsx
jgyselov: How would the user end up at a Static IP step without an infra env? Shouldn't the UI wait for the infra env to be created and set before moving anywhere?

jgyselov (changes requested):

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: Applied

libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx
asmasarw: This Done- but some places has multiple checks.
e.g. 

const validator = allowSuffix ? ipValidationSchema : ipNoSuffixValidationSchema;
in this file: 

libs/ui-lib/lib/common/validationSchemas/addressValidation.tsx 
Line: 48

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: DONE

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: DONE

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -91,13 +133,20 @@ export const OptionalConfigurationsStep = () => {
       Yup.object({
         sshPublicKey: sshPublicKeyValidationSchema(t),
         pullSecret: isInOcm ? Yup.string() : pullSecretValidationSchema(t),
+        rendezvousIp: Yup.string()
+          .max(45, 'IP address must be at most 45 characters')
+          .concat(ipValidationSchema(t)),
```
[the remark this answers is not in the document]
jgyselov: In OCM translation strings are allowed but not required.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -73,10 +94,31 @@ const OptionalConfigurationsForm: React.FC<OptionalConfigurationsFormProps> = ({
...
+
+const buildInfraEnvUpdateParams = (
+  values: OptionalConfigurationsValues,
+  disconnectedInfraEnv: InfraEnv | undefined,
+): InfraEnvUpdateParams => ({
+  sshAuthorizedKey: values.sshPublicKey,
+  rendezvousIp: values.rendezvousIp,
+  staticNetworkConfig: shouldSendDummyStaticConfig(values, disconnectedInfraEnv)
+    ? getDummyInfraEnvField()
+    : [],
```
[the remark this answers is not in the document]
jgyselov: 👆 @asmasarw 

what the bot said. Send `[]` if the user selected DHCP, send dummy data if user selected Static IP and there is no `.staticNetworkConfig`, send `undefined` otherwise.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -91,13 +133,20 @@ export const OptionalConfigurationsStep = () => {
...
     [t],
   );
 
   const initialValues: OptionalConfigurationsValues = {
     sshPublicKey: disconnectedInfraEnv?.sshAuthorizedKey ?? '',
     pullSecret: defaultPullSecret ?? '',
+    rendezvousIp: disconnectedInfraEnv?.rendezvousIp ?? '',
+    hostsNetworkConfigurationType: wizardStepIds.some((id) => id.startsWith('static-ip'))
+      ? HostsNetworkConfigurationType.STATIC
+      : HostsNetworkConfigurationType.DHCP,
```
jgyselov:
```suggestion
    hostsNetworkConfigurationType: disconnectedInfraEnv.hostsNetworkConfigurationType === HostsNetworkConfigurationType.STATIC
      ? HostsNetworkConfigurationType.STATIC
      : HostsNetworkConfigurationType.DHCP,
```

libs/ui-lib/lib/ocm/components/wizard/utils/wizardTransition.ts
```diff
@@ -382,3 +382,27 @@ export const canNextNetwork = ({ cluster }: TransitionProps): boolean =>
...
+      entries.push({ stepId, visualNumber, isStaticIpFormGroup: true });
+      i += 2;
+    } else {
+      entries.push({ stepId, visualNumber, isStaticIpFormGroup: false });
+      i += 1;
+    }
+    visualNumber += 1;
+  }
+  return entries;
+};
```
jgyselov: Move to [ClusterWizardNavigation.tsx](https://github.com/openshift-assisted/assisted-installer-ui/blob/master/libs/ui-lib/lib/ocm/components/wizard/wizardComponents/ClusterWizardNavigation.tsx)

jgyselov (changes requested):

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: we don't have 'hostsNetworkConfigurationType' property -- we have staticNetworkConfig.

libs/ui-lib/lib/ocm/components/wizard/utils/wizardTransition.ts
asmasarw: DONE.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -73,6 +95,32 @@ const OptionalConfigurationsForm: React.FC<OptionalConfigurationsFormProps> = ({
...
+    return []; // clear any existing static config
+  }
+  // Static IP selected — only seed with dummy if no real config exists yet
+  if (
+    !infraEnv?.staticNetworkConfig ||
+    (typeof infraEnv.staticNetworkConfig === 'string' && isDummyYaml(infraEnv.staticNetworkConfig))
+  ) {
+    return getDummyInfraEnvField();
+  }
+  return undefined; // real config already present — omit from PATCH to preserve it
```
jgyselov: This could be made even simpler:

```suggestion
  if (!infraEnv?.staticNetworkConfig) {
    return getDummyInfraEnvField();
  }
  return undefined;
```

If there already is a config (real or dummy), you can omit `staticNetworkConfig` to persist anything that's already set.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -91,13 +132,20 @@ export const OptionalConfigurationsStep = () => {
...
+          .concat(ipValidationSchema(t)),
       }),
     [t],
   );
 
   const initialValues: OptionalConfigurationsValues = {
     sshPublicKey: disconnectedInfraEnv?.sshAuthorizedKey ?? '',
     pullSecret: defaultPullSecret ?? '',
+    rendezvousIp: disconnectedInfraEnv?.rendezvousIp ?? '',
+    hostsNetworkConfigurationType: disconnectedInfraEnv?.staticNetworkConfig
```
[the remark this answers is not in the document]
jgyselov: The UI sends  `[]` to clear the staticNetworkConfig. However, when doing so, the response has `staticNetworkConfig` set to undefined (not empty array). Which is falsy, so this condition works.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -107,29 +162,44 @@ export const OptionalConfigurationsStep = () => {
...
           const { data: cluster } = await ClustersAPI.registerDisconnected({
             name: DISCONNECTED_CLUSTER_NAME,
             openshiftVersion: DISCONNECTED_OPENSHIFT_VERSION,
           });
           setDisconnectedCluster(cluster);
+          clusterToUse = cluster;
+        }
 
+        // Create the infraEnv if one doesn't exist yet; otherwise update the existing one
+        if (!infraEnvToUse?.id) {
```
jgyselov: I'm not sure why you have two separate conditions like this now.

I think the proper way should be to 
1. if cluster and infraEnv don't exist, create cluster _and_ infra env
2. if cluster and infraEnv exist, update infra env

There should never be a scenario where one of these isn't true.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
```diff
@@ -73,6 +96,32 @@ const OptionalConfigurationsForm: React.FC<OptionalConfigurationsFormProps> = ({
...
+  values: OptionalConfigurationsValues,
+  infraEnv: InfraEnv | undefined,
+): HostStaticNetworkConfig[] => {
+  if (values.hostsNetworkConfigurationType === HostsNetworkConfigurationType.DHCP) {
+    return [];
+  }
+  // Static IP selected — resend existing config, or seed with dummy if none exists yet.
+  // Note: we intentionally skip isDummyYaml here. Partially-filled configs (e.g. network-wide
+  // values set but no host IPs yet) also contain DUMMY NIC names and would be misidentified
+  // as dummy, causing real data to be overwritten.
```
jgyselov: I fear this part of the comment would create more confusion than anything. It might have made sense in your dev process but with the code on its own, it's redundant.

```suggestion
  // Static IP selected — resend existing config, or seed with dummy if none exists yet.
```

jgyselov (approved):

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: Done.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/OptionalConfigurationsStep.tsx
asmasarw: Done

jgyselov (approved):

PR #4013, opened 2026-09-03
OCPBUGS-115122: OVE Below the sea level UI: Technology preview is shown right next to Machine Network dropdown when using IPV6

asmasarw:
Before: 

image

After:

image

### DIFF
--- a/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/SubnetsDropdown.tsx
+++ b/libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/SubnetsDropdown.tsx
@@ -10,16 +10,16 @@ import {
 } from '@patternfly/react-core';
 import { useField } from 'formik';
 import { Address4, Address6 } from 'ip-address';
-import { getFieldId } from '../../ui';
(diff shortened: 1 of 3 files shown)


jgyselov (dismissed):
This doesn't seem to address the linked issue.

libs/ui-lib/lib/common/components/clusterWizard/networkingSteps/SubnetsDropdown.tsx
```diff
@@ -167,11 +160,6 @@ export const SubnetsDropdown = ({
       {...props}
     >
       {currentDisplayValue}
-      {showBadge && (
```
LiorSoffer, line 170: This change will remove the badge completely
let’s remove it only for the OVE below-the-sea flow.

libs/ui-lib/lib/ocm/components/clusters/AssistedInstallerHeader.tsx
LiorSoffer, line 1: Can you create and add another ticket for removing the tech preview badge from the header?

LiorSoffer (changes requested):

LiorSoffer (approved):

PR #4030, opened 2026-09-08
Cherry pick 4012 to release 4.21

asmasarw:


LiorSoffer (changes requested):
Please remove the last 2 commits here, they are bringing back the `OCM_REFRESH_TOKEN` which we're trying to remove. then remove the `OCM_REFRESH_TOKEN` from `apps/assisted-ui/src/components/Chatbot.tsx‎`
use `yarn fix-code-style:all` before commit  :)

PR #4054, opened 2026-09-16
Add Single-Stack, Increase ImageSize Label and show LVMS Label

asmasarw:
## Summary by CodeRabbit

- **New Features**
  - Added disconnected-install support for network-wide static IP configuration.
  - Supports IPv4-only, IPv6-only, and dual-stack protocol selections.
  - Added protocol-aware IP, DNS, gateway, subnet, VLAN, and prefix validation.
  - Preserves static IP settings when loading and saving disconnected configurations.
  - Improved gateway detection for IPv4 and IPv6 network routes.
  - Updated disconnected cluster creation to use the supported OpenShift version.

- **Updates**
  - Increased the displayed disconnected-install ISO size estimate to 60+ GB.


libs/ui-lib/lib/ocm/components/featureSupportLevels/FeatureSupportLevelProvider.tsx
```diff
@@ -65,7 +65,8 @@ export const NewFeatureSupportLevelProvider: React.FC<NewSupportLevelProviderPro
     return featureSupportLevels;
   }, [featureSupportLevels]);
 
-  const isLoading = !supportLevelData || loadingOCPVersions || isInfraEnvLoading;
+  const isLoading =
+    (!!openshiftVersion && !featureSupportLevels) || loadingOCPVersions || isInfraEnvLoading;
```
ElayAharoni: what if featureSupportLevels fetch fails? the user will get stuck on the loading state

ElayAharoni (changes requested):

libs/ui-lib/lib/ocm/components/featureSupportLevels/FeatureSupportLevelProvider.tsx
asmasarw: hi @ElayAharoni - This was missed - 
I have added it locally, but it's not necessary at all :) 
Reverted the change.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/DisconnectedReviewStep.tsx
```diff
@@ -28,9 +28,10 @@ import { ClusterWizardFooter, ClusterWizardNavigation } from '../../wizardCompon
 import { useClusterWizardContext } from '../../clusterWizardContext';
 import { DISCONNECTED_OPENSHIFT_VERSION } from './BasicStep';
 
+const opSpecs = getOperatorSpecs(() => undefined, true);
+
 export const DisconnectedReviewStep = () => {
   const { moveBack, disconnectedInfraEnv } = useClusterWizardContext();
-  const opSpecs = getOperatorSpecs(() => undefined);
```
rawagner, line 33: is this change needed? If yes, can you explain why ?

ElayAharoni (approved):

libs/ui-lib/lib/ocm/components/wizard/steps/staticIp/components/FormViewNetworkWide/FormViewNetworkWideFields.tsx
```diff
@@ -239,15 +245,172 @@ export const ProtocolTypeSelect = () => {
...
+            setIpv6(getEmptyIpConfig());
             setProtocolType('ipv4');
           }}
         />
       )}
     </>
   );
 };
+
+const useDisconnectedSingleStackFamilyFromSubnet = () => {
```
rawagner: this flow is very confusing.

we should try to avoid using useEffect to change the formik's state.

IMO, when we select single-stack in disconnected mode, it should use a different set of fields/state/validation that handles this ipv4/ipv6 natively (input, validation..) without a need to use effects.

libs/ui-lib/lib/ocm/components/wizard/steps/disconnected/DisconnectedReviewStep.tsx
asmasarw: The "true" value is required here in order to have 'useLVMS' = true
In the Review & Download Step, we have List of available operators:

and there it renderes the LVMS as "Logical Volume Manager"
So in order to render the correct value, we used the true flag and now its rendered as "Logical Volume Manager Storage"
 
 
image

libs/ui-lib/lib/ocm/components/wizard/steps/staticIp/components/FormViewNetworkWide/FormViewNetworkWideFields.tsx
asmasarw: Changed the behavior to:

Refactor disconnected single-stack static IP without Formik useEffect

Use family-agnostic single-stack fields and normalize IPv4/IPv6 at load/save
via disconnected wrappers, keeping connected flow on existing schemas.

rawagner (approved):

ElayAharoni (approved):

